PR #35, opened 2020-07-18
Use more idiomatic Single.fromCallable() in RxRoom

ZacSweers:
This is a more idiomatic built-in API from RxJava for creating a `Single` from a given `Callable`. Functionally it's the same, but with more pedantic internal checks and adherence to RxJava's conventional plugin system. `Single.create()` is more for bridging from non-first-class-APIs like external listeners, other concurrency frameworks, etc.

(this is also helping test a PR flow! CC @yigit)


dlam (approved):

yigit (changes requested):
i think this is intentioanl.

PR #37, opened 2020-07-23
Add a link for github contributions.

tikurahul:
Test: N/A


dlam (approved):

dlam:
Do you need ^ and $ as well?

tikurahul:
I don't because GitHub does not format the PR body in a way you would expect. :/

PR #42, opened 2020-07-23
Add a note to use blob filter (updated PR)

liutikas:
Test: N/A


tikurahul (approved):
You will need to add a `Test` stanza.

tikurahul (approved):

khaclinh157 (approved):

PR #44, opened 2020-07-24
Better events

tikurahul:
Test: N/A

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -1,7 +1,7 @@
 name: AndroidX Presubmits
 on:
   pull_request:
-    types: [assigned, opened, synchronize, reopened, labeled]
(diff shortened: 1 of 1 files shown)


dlam (approved):
For external contributors retrying changes, I suppose we expect them to go through the actions UI then?

tikurahul:
No, the only way to re-trigger is by adding a comment. 💬

dlam:
Even on their fork? Don't they have access to workflow_dispatch from their actions tab? I guess it doesn't matter since it retriggers on push anyway?

ZacSweers:
One thing you can do is add comment hotwords and make your action hide or delete the comment when it's done. Example:

https://github.com/ZacSweers/rebase/blob/41a109c7fe042cc94419a72900e67150d40dc91e/entrypoint.sh#L97-L111

PR #45, opened 2020-07-24
Emphasize the "Test:" string requirement

ZacSweers:
Just makes it clear that it's not just a suggestion or polite ask

## Testing

Test: N/A

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,7 +6,7 @@
 
 ## Testing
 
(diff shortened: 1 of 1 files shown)


ZacSweers:
Worth mentioning that the template, unfilled, will _also_ pass that check. Might want to make it ephemeral as a forcing function to make sure people do it

Example of an ephemeral message

```
<!-- Add a line starting with "Test:" to explain the test. plan here. This comment won't be visible in the PR -->
```

tikurahul (approved):

PR #49, opened 2020-07-31
Support opening gzipped databases.

ahmedre:
This patch adds a method for creating a database from a gzipped file.
This is only implemented for external files, since assets are
transparently gzipped and extracted by the build tools.

Test: Added a unit test to SQLiteCopyOpenHelperTest.
Fixes: b/146911060


room/runtime/src/main/java/androidx/room/RoomDatabase.java
```diff
@@ -624,6 +625,31 @@ private static boolean isLowRamDevice(@NonNull ActivityManager activityManager)
...
+         * {@link Database#exportSchema()} is enabled.
+         * <p>
+         * This method is not supported for an in memory database {@link Builder}.
+         *
+         * @param databaseFile The database file.
+         *
+         * @return This {@link Builder} instance.
+         */
+        @NonNull
+        public Builder<T> createFromGzippedFile(@NonNull File databaseFile) {
```
ahmedre: debated a bit about whether to add this as a new method or just detect based on the filename ending in `.gz`, but opted for this at the end because thought it's cleaner. open to suggestions though.

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -63,9 +66,21 @@
             @Nullable File copyFromFile,
             int databaseVersion,
             @NonNull SupportSQLiteOpenHelper supportSQLiteOpenHelper) {
+        this(context, copyFromAssetPath, copyFromFile, false, databaseVersion,
```
ahmedre: kept this method for binary compatibility purposes - but now that I think about, I guess this class is not open for the public anyway so I can probably remove this constructor

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -177,7 +192,12 @@ private void copyDatabaseFile(File destinationFile) throws IOException {
         if (mCopyFromAssetPath != null) {
             input = Channels.newChannel(mContext.getAssets().open(mCopyFromAssetPath));
         } else if (mCopyFromFile != null) {
-            input = new FileInputStream(mCopyFromFile).getChannel();
+            final FileInputStream fileInputStream = new FileInputStream(mCopyFromFile);
+            if (mIsCopyFromFileGzipped) {
+                input = Channels.newChannel(new GZIPInputStream(fileInputStream));
```
ahmedre: only opted for gzip because it's simple - zip itself will be a lot of code to loop through, get and extract zip entries, and so on which is probably not necessary?

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelperFactory.java
```diff
@@ -32,24 +34,36 @@
...
     private final File mCopyFromFile;
+    private final boolean mIsCopyFromFileGzipped;
     @NonNull
     private final SupportSQLiteOpenHelper.Factory mDelegate;
 
     SQLiteCopyOpenHelperFactory(
             @Nullable String copyFromAssetPath,
             @Nullable File copyFromFile,
             @NonNull SupportSQLiteOpenHelper.Factory factory) {
+        this(copyFromAssetPath, copyFromFile, false, factory);
```
ahmedre: kept this for backward compatibility, but since this is an internal class, can just remove it i guess

room/runtime/src/test/java/androidx/room/SQLiteCopyOpenHelperTest.kt
```diff
@@ -180,6 +184,30 @@ class SQLiteCopyOpenHelperTest {
...
+            true,
+            DB_VERSION,
+            delegate
+        ).apply { setDatabaseConfiguration(configuration) }
+        openHelper.writableDatabase
+        openHelper.writableDatabase
+
+        verify(delegate).databaseName
+        verify(context).getDatabasePath(DB_NAME)
+        verifyNoMoreInteractions(configuration)
```
ahmedre: i am sad about this - i copied the single `singleCopy` test, but lost the part that makes sure the file is only read once - would love some suggestions on how to make this test better.

room/runtime/src/test/java/androidx/room/SQLiteCopyOpenHelperTest.kt
```diff
@@ -221,6 +249,14 @@ class SQLiteCopyOpenHelperTest {
         }
     }
 
+    private fun gzipFile(input: File, output: File) {
+        input.inputStream().channel.use { inputChannel ->
```
ahmedre: would have used `FileUtil` but it takes a `FileChannel` instead of a `WriteableByteChannel`

ahmedre:
The original request wanted support for assets as well - I did this initially but realized there's no point - the Android build tools automatically gzip files in assets (and will ungzip ones that have already been gzipped). There's no size difference of bundling file.db, file.gz, or file.bin (a gzip file in disguise).

Screen Shot 2020-08-01 at 12 35 16 AM

ahmedre:
The current failure is due to:

> This project uses AndroidX dependencies, but the 'android.useAndroidX' property is not enabled. Set this property to true in the gradle.properties file and retry.

This happens for me even without my commit, and I worked around it by adding `android.useAndroidX=true` to my `.gradle/gradle.properties` - appreciate any advice on getting past this

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
ahmedre: (removed).

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelperFactory.java
ahmedre: (removed).

tikurahul:
@ahmedre Let me take a look at the build failure  for the `android.useAndroidX=true` related failure.

room/runtime/src/main/java/androidx/room/RoomDatabase.java
ahmedre: lint doesn't seem happy about this - it wants a method name starting with `set` or `get` instead, but thought this is more in line with what's there.

room/runtime/src/test/java/androidx/room/SQLiteCopyOpenHelperTest.kt
danysantiago: These host-side unit tests were really about the file & thread locking mechanism, I say a better test would be one in the integration app, here: https://github.com/androidx/androidx/blob/androidx-master-dev/room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/PrepackageTest.java

You can simply copy one of the existing databases and just gzip it: https://github.com/androidx/androidx/tree/androidx-master-dev/room/integration-tests/testapp/src/main/assets/databases

room/runtime/src/test/java/androidx/room/SQLiteCopyOpenHelperTest.kt
danysantiago: If we move the tests to an android test then this won't be needed.

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
yigit: sometimes we need to them if generated code calls it but otherwise, we can just update. i think generated code is not calling here so should be fine, i'll defer to @danysantiago .

danysantiago (commented):
Thanks for making this PR!

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
danysantiago: Should be fine to update this constructor, not called by gen code.

room/runtime/src/main/java/androidx/room/RoomDatabase.java
danysantiago: What if instead we do a `createFromInputStream(@NonNull Callable<InputStream>)` such that the user can provide a callable that knows how to create the input stream to be used for the copy and specifically they can then provide us with the `GZIPInputStream` or a `ZipInputStream` (at the right zip entry), etc... and Room would simply copy those bytes and attempt to use it as a DB.

yigit:
> The current failure is due to:
> 
> > This project uses AndroidX dependencies, but the 'android.useAndroidX' property is not enabled. Set this property to true in the gradle.properties file and retry.
> 
> This happens for me even without my commit, and I worked around it by adding `android.useAndroidX=true` to my `.gradle/gradle.properties` - appreciate any advice on getting past this

i'm working on that, sorry some hacks i've put failed after AGP update. 
I sent a quick CL to fix for this case and working on a better solution. (or trying to find one :) )

Sorry for making you deal w/ these problems until we make this process more resilient.

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -172,12 +179,14 @@ private void verifyDatabaseFile() {
         }
     }
 
-    private void copyDatabaseFile(File destinationFile) throws IOException {
+    private void copyDatabaseFile(File destinationFile) throws Exception {
```
ahmedre: had to do this as a result of `Callable`'s `call` method throwing :(

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelperFactory.java
```diff
@@ -20,7 +20,11 @@
 import androidx.annotation.Nullable;
 import androidx.sqlite.db.SupportSQLiteOpenHelper;
 
+import org.jetbrains.annotations.NotNull;
```
ahmedre: oops, will fix

room/runtime/src/main/java/androidx/room/RoomDatabase.java
ahmedre: done.

room/runtime/src/test/java/androidx/room/SQLiteCopyOpenHelperTest.kt
ahmedre: done, thanks!

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
danysantiago: This is a bit scary, lets wrap the Callable's `call()` invocation in a try-catch with `Exception` and throw a `IOException` (using the catched one as cause) instead otherwise this messes up with the destructive migration flow.

room/runtime/src/main/java/androidx/room/RoomDatabase.java
```diff
@@ -919,18 +948,24 @@ public T build() {
...
                 if (mName == null) {
                     throw new IllegalArgumentException("Cannot create from asset or file for an "
                             + "in-memory database.");
                 }
-                if (mCopyFromAssetPath != null && mCopyFromFile != null) {
-                    throw new IllegalArgumentException("Both createFromAsset() and "
-                            + "createFromFile() was called on this Builder but the database can "
-                            + "only be created using one of the two configurations.");
+
+                final int copiesSet = (mCopyFromAssetPath == null ? 0 : 1) +
```
danysantiago: Maybe rename to 'copyConfigurations' ? Indicating there are more than one copy configuration was given.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/PrepackageTest.java
```diff
@@ -67,6 +70,30 @@ public void createFromAsset() {
         database.close();
     }
 
+    @Test
+    public void createFromZippedAsset() {
```
danysantiago, line 77: Would be amazing if you also add another one for your use case (GZIPInputStream) just to have good coverage there when transforming it to a Channel.

danysantiago (changes requested):

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/PrepackageTest.java
ahmedre: done.

danysantiago (approved):
Latest changes look good. Thank you so much.
We'll figure out the build issues and get this merged soon.

yigit (approved):

danysantiago:
Hey @ahmedre, sorry for the inconvenience, we got some changes merged for the build issues but you'll need to rebase this PR in other for the changes to get picked up. When you get a chance can you please rebase this PR? Thanks!

tikurahul:
Awesome :clap:

PR #50, opened 2020-07-31
set androidx property via gradlew

yigit:
This is a quick hack to fix the build.

The way we are currently copying gradle.properties is not fast
enough for configuration time. On the other hand, we cannot move
them into a real gradle.properties file because the main androidx
build ends up reading them.

I'm sending this PR to at least fix the workflow build until
finding a better solution than copying them in settings gradle
file

Test: ./gradlew bOS


tikurahul (approved):

yigit:
abandoning this in favor of https://android-review.googlesource.com/c/platform/frameworks/support/+/1384014

PR #52, opened 2020-08-04
Add support for `pull_request_target` events.

tikurahul:
* This allows us to share secrets with actions, because the workflow
  gets triggered in the `base` context.

Test: Will need to run additional tests after the PR lands.


dlam:
Do we still need the regular pull_request dispatch? Or are we adding both for migration purposes?

dlam (approved):

PR #54, opened 2020-08-04
Workflow improvements.

tikurahul:
* Makes adding more modules easier.
* Remove the `pull_request` hook in favor of `pull_request_target`.
* Add secrets back to the hooks.

Test: Workflow tests.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -3,8 +3,6 @@ on:
   push:
   pull_request_target:
     types: [opened, synchronize, reopened]
-  pull_request:
-    types: [opened, synchronize, reopened]
   pull_request_review:
     types: [submitted, edited]
   workflow_dispatch:
@@ -31,6 +29,7 @@ jobs:
         uses: tikurahul/github-workflow-webhook-action@master
         with:
           url: 'https://androidx.dev/github/androidX/presubmit/hook'
+          secret: ${{ secrets.ANDROIDX_PRESUBMIT_HOOK_SECRET }}
           payload: '{ "platform": "${{ matrix.os }}", "token": "${{ secrets.GITHUB_TOKEN }}", "state": "started"}'
       - name: Setup Java environment based on setup-java
         uses: actions/setup-java@v1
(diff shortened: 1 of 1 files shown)


tikurahul:
Also: https://github.com/tikurahul/androidx/runs/947050349

.github/workflows/presubmit.yml
```diff
@@ -72,26 +71,23 @@ jobs:
...
       - name: Upload Results
         if: always()
         uses: actions/upload-artifact@v2
         with:
           name: artifacts
           path: ~/dist
-      - name: Completed WebHook
-        if: always()
+      - name: Successful WebHook
+        if: ${{ success() }}
```
dlam: Is the ${{ }} necessary here?

dlam: i.e., why isn't it needed for `always()`

.github/workflows/presubmit.yml
```diff
@@ -72,26 +71,23 @@ jobs:
...
+        uses: tikurahul/github-workflow-webhook-action@master
+        with:
+          url: 'https://androidx.dev/github/androidX/presubmit/hook'
+          secret: ${{ secrets.ANDROIDX_PRESUBMIT_HOOK_SECRET }}
+          payload: '{ "platform": "${{ matrix.os }}", "token": "${{ secrets.GITHUB_TOKEN }}", "state": "completed", "success": true }'
```
dlam, line 86: it kinda sucks that we can't inline the if failure / success check here? That looks like the only difference to me from these two branches

.github/workflows/presubmit.yml
```diff
@@ -72,26 +71,23 @@ jobs:
...
+      - name: Failure WebHook
+        if: ${{ failure() || cancelled() }}
```
dlam, line 88: Are these states based on previous step or the entire job? e.g., what happens if:
```
- name: bOS fails
  ...
- name: always succeeds
  if: always()
  run: true
- name Successful WebHook
  if: success()
  run: echo "this shouldn't run!"
```

dlam (approved):

.github/workflows/presubmit.yml
tikurahul: Yes, it does. But unfortunately there is no easy way to inline this without having to write more actions. :disappointed:

.github/workflows/presubmit.yml
tikurahul: These are based on the entire job: https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions has more information.

.github/workflows/presubmit.yml
tikurahul: I just copied the example at https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#job-status-check-functions. 

I think this is something I can probably get rid of.

tikurahul: Done.

PR #55, opened 2020-08-05
Add onOpenPrepackagedDatabase callback

mzgreen:
This patch adds a new builder methods that accept new PrepackagedCallback for creating a database.
The new PrepackagedCallback contains a single method onOpenPrepackagedDatabase(SupportSQLiteDatabase) which is called right after creating a db from Asset/File/InputStream.

onOpenPrepackagedDatabase(SupportSQLiteDatabase) callback is transactional which means that it'll be called only once after creating a db from Asset/File/InputStream. If an exception is thrown inside onOpenPrepackagedDatabase callback then the next time this database is opened, it should be created from Asset/File/InputStream again and onOpenPrepackagedDatabase should also be called again.

Test: Added a unit tests to PrepackageTest.
Fixes: b/148934423


room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/DatabaseCallbackTest.java
```diff
@@ -205,6 +211,82 @@ public void onDestructiveMigration_calledOnUpgrade() {
         database.close();
     }
 
+    @Test
+    public void onCreateFromAsset_calledOnOpenPrepackagedDatabase() {
+        Context context = ApplicationProvider.getApplicationContext();
```
tikurahul: Might want to do this in an `@After` block so tests can be idempotent.

room/runtime/api/api_lint.ignore
```diff
@@ -7,8 +7,14 @@ BuilderSetStyle: androidx.room.RoomDatabase.Builder#allowMainThreadQueries():
     Builder methods names should use setFoo() / addFoo() / clearFoo() style: method androidx.room.RoomDatabase.Builder.allowMainThreadQueries()
 BuilderSetStyle: androidx.room.RoomDatabase.Builder#createFromAsset(String):
     Builder methods names should use setFoo() / addFoo() / clearFoo() style: method androidx.room.RoomDatabase.Builder.createFromAsset(String)
+BuilderSetStyle: androidx.room.RoomDatabase.Builder#createFromAsset(String, androidx.room.RoomDatabase.PrepackagedCallback):
```
tikurahul: You should not be adding these to the ignore list. Use `@Suppress` instead. We added some notes in `Contributing.md` for this.

tikurahul (commented):
These are just initial thoughts on the PR. @danysantiago will look at the API in more detail.

room/runtime/api/restricted_current.txt
```diff
@@ -5,7 +5,7 @@ package androidx.room {
...
     ctor @Deprecated @RestrictTo(androidx.annotation.RestrictTo.Scope.LIBRARY_GROUP_PREFIX) public DatabaseConfiguration(android.content.Context, String?, androidx.sqlite.db.SupportSQLiteOpenHelper.Factory, androidx.room.RoomDatabase.MigrationContainer, java.util.List<androidx.room.RoomDatabase.Callback!>?, boolean, androidx.room.RoomDatabase.JournalMode!, java.util.concurrent.Executor, java.util.concurrent.Executor, boolean, boolean, boolean, java.util.Set<java.lang.Integer!>?, String?, java.io.File?);
-    ctor @RestrictTo(androidx.annotation.RestrictTo.Scope.LIBRARY_GROUP_PREFIX) public DatabaseConfiguration(android.content.Context, String?, androidx.sqlite.db.SupportSQLiteOpenHelper.Factory, androidx.room.RoomDatabase.MigrationContainer, java.util.List<androidx.room.RoomDatabase.Callback!>?, boolean, androidx.room.RoomDatabase.JournalMode, java.util.concurrent.Executor, java.util.concurrent.Executor, boolean, boolean, boolean, java.util.Set<java.lang.Integer!>?, String?, java.io.File?, java.util.concurrent.Callable<java.io.InputStream!>?);
```
yigit, line 8: we cannot remove these constructors for backwards compatibility (because generated code w/ older version of room might be calling it).
so need to add the parameter to the end and chain.
@danysantiago to double check in case this does not get called by generated code. (though seems like we've been chaining these: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:room/runtime/src/main/java/androidx/room/DatabaseConfiguration.java?q=DatabaseConfiguration)

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -101,6 +120,21 @@ public synchronized SupportSQLiteDatabase getReadableDatabase() {
         if (!mVerified) {
             verifyDatabaseFile();
             mVerified = true;
+            if(mDatabaseConfiguration.prepackagedCallbacks != null) {
```
yigit: can we extract this into another method to avoid code duplication?

room/runtime/api/api_lint.ignore
mzgreen: I tried using annotations but `./gradlew updateApi` task was still failing on metalava lint. Adding those manually fixed the issue. But I'll double check this.

room/runtime/src/main/java/androidx/room/DatabaseConfiguration.java
```diff
@@ -291,6 +295,7 @@ public DatabaseConfiguration(@NonNull Context context, @Nullable String name,
             @NonNull SupportSQLiteOpenHelper.Factory sqliteOpenHelperFactory,
             @NonNull RoomDatabase.MigrationContainer migrationContainer,
             @Nullable List<RoomDatabase.Callback> callbacks,
+            @Nullable List<RoomDatabase.PrepackagedCallback> prepackagedCallbacks,
```
danysantiago: To be backwards compatible with old generated code, even though this API is marked with `@RestrictTo` we need to copy this constructor and add the new parameter and then update all other constructors to call the newest one with `null`. Here is an example: https://github.com/androidx/androidx/commit/1263185ef8c50d4059758e382610345a24f82c53#diff-a135d07f04ca20de890a27db682fe8e2

room/runtime/src/main/java/androidx/room/DatabaseConfiguration.java
```diff
@@ -61,6 +61,9 @@
     @Nullable
     public final List<RoomDatabase.Callback> callbacks;
 
+    @Nullable
+    public final List<RoomDatabase.PrepackagedCallback> prepackagedCallbacks;
```
danysantiago: I'm thinking it might be best / simpler if instead of a list we only take one callback. The reason is we only allow one of the configurations `createFromFile` / `createFromAsset` / `createFromStream`, and since the callback is passed on those builder methods it seems weird that calling it again to change the configuration (to like update the file name) you end up passing the same callback and it'll get invoked twice.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/DatabaseCallbackTest.java
```diff
@@ -205,6 +211,82 @@ public void onDestructiveMigration_calledOnUpgrade() {
         database.close();
     }
 
+    @Test
+    public void onCreateFromAsset_calledOnOpenPrepackagedDatabase() {
```
danysantiago: Can you plz move these test to https://github.com/androidx/androidx/blob/androidx-master-dev/room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/PrepackageTest.java? These tests are for the old callbacks and since we are scoping the new one I think it might be best to have it with the rest of prepackage DB tests.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/DatabaseCallbackTest.java
```diff
@@ -231,4 +313,28 @@ public void onDestructiveMigration(@NonNull SupportSQLiteDatabase db) {
...
+
+        boolean mOpenPrepackagedDatabase;
+
+        @Override
+        public void onOpenPrepackagedDatabase(@NonNull SupportSQLiteDatabase db) {
+            mOpenPrepackagedDatabase = true;
+        }
+    }
+
+    private static void copyAsset(InputStream input, File outputFile) throws IOException {
```
danysantiago: If we move the tests to `PrepackageTest` we won't need to copy this.

room/runtime/api/api_lint.ignore
danysantiago: Try `@SupressLint`, take a look at `createFromInputStream` which added the same suppression.

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -217,4 +251,35 @@ private void copyDatabaseFile(File destinationFile) throws IOException {
...
         }
     }
+
+    private SupportSQLiteOpenHelper createFrameworkOpenHelper() {
+        SupportSQLiteOpenHelper frameworkOpenHelper = null;
+        try {
+            FrameworkSQLiteOpenHelperFactory factory = new FrameworkSQLiteOpenHelperFactory();
+            Configuration configuration = Configuration.builder(mContext)
+                    .name(mDatabaseConfiguration.name)
+                    .callback(new Callback(mDatabaseVersion){
```
danysantiago: Lets not open the DB with the 'expected final version' we need to open the pre-package DB with the same version it already has, to avoid going through the migration path which will end up updating the version and when Room opens the DB again it will think its already updated. To find out the version of the DB without opening use https://github.com/androidx/androidx/blob/androidx-master-dev/room/runtime/src/main/java/androidx/room/util/DBUtil.java#L139

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
mzgreen: I wanted to extract this logic into a separate method and call it in `SQLiteCopyOpenHelper.getWritableDatabase()` and `SQLiteCopyOpenHelper.getReadableDatabase()` but when I temporarily open the db using `FrameworkSqliteOpenHelper` then I need to call `getReadableDatabase()` or `getWritableDatabase()` and I wasn't sure if I can just use one of those for both calls.
For example - can I open the db using framework helper, call `getWritableDatabase()` and pass it to the `onPrepackagedCallback` inside `SQLiteCopyOpenHelper.getReadableDatabase()`? Or I have to call `getReadableDatabase()` in this case?

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -217,4 +251,35 @@ private void copyDatabaseFile(File destinationFile) throws IOException {
...
+                        }
+
+                        @Override
+                        public void onUpgrade(@NonNull SupportSQLiteDatabase db, int oldVersion,
+                                int newVersion) {
+                        }
+                    })
+                    .build();
+                frameworkOpenHelper = factory.create(configuration);
+        } catch (SQLiteException e) {
```
danysantiago: I don't think we should catch all SQLiteException and ignore them.

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -101,6 +120,21 @@ public synchronized SupportSQLiteDatabase getReadableDatabase() {
         if (!mVerified) {
             verifyDatabaseFile();
             mVerified = true;
+            if(mDatabaseConfiguration.prepackagedCallbacks != null) {
+                // Temporarily open database using FrameworkSQLiteOpenHelper
+                SupportSQLiteOpenHelper helper = createFrameworkOpenHelper();
+                if (helper != null) {
+                    SupportSQLiteDatabase db = helper.getReadableDatabase();
+                    // Invoke callbacks passing db as a param
+                    callOnOpenPrepackagedDatabaseCallbacks(db);
```
danysantiago: `callOnOpenPrepackagedDatabaseCallbacks` should be inside a try-finally where in the finally block we close the DB.

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
danysantiago: You can still extract the logic into a function that receives a boolean parameter `writable` to know which of the two methods to call.

room/runtime/src/main/java/androidx/room/RoomDatabase.java
```diff
@@ -1119,4 +1218,19 @@ public void onOpen(@NonNull SupportSQLiteDatabase db) {
         public void onDestructiveMigration(@NonNull SupportSQLiteDatabase db){
         }
     }
+
+    /**
+     * Callback for {@link Builder#createFromAsset(String)}, {@link Builder#createFromFile(File)}
+     * and {@link Builder#createFromInputStream(Callable)}
```
danysantiago, line 1218: We should add a bit more docs here: "This callback will be invoked after the pre-package DB is copied but before Room had a chance to open it and therefore before the RoomDatabase.Callback methods are invoked. This callback can be useful for updating the pre-package DB schema to satisfy Room's schema validation."

danysantiago (changes requested):
This is great, thank you for making this pull request!

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
mzgreen: I appreciate those explanations. It helps a lot. Thanks!

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
mzgreen: What do you think we should do with the exceptions? Would it be enough to print some log here?

danysantiago: We should let it propagate up and crash the app (or let the user of Room catch it), its the same as exceptions in `onOpen`, they would crash the app.

room/runtime/api/api_lint.ignore
mzgreen: I think the problem is that for example `createFromFile` needs 2 suppressions: `BuilderSetStyle` and `StreamFiles`. If I add only one then it's picked up by the tooling but it still complains about second issue. When I do `@SuppressLint("BuilderSetStyle, StreamFiles")` then it doesn't work at all, same effect as not adding any suppression. I think this should work since the value of `SuppressLint` is an array of strings.

dlam: I'm not sure it's possible to pass an array of string via varargs to the annotation? Have you tried:
`@SuppressLint({"BuilderSetStyle", "StreamFiles"})`? This is how we do it elsewhere in the codebase.

dlam: Oh, I just noticed you are passing a single string that's comma separated :smiley: 

Definitely use `@SuppressLint({"BuilderSetStyle", "StreamFiles"})` instead.

mzgreen: Ah Kotlin habits.. You're right! It worked, thanks :)

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/DatabaseCallbackTest.java
mzgreen: I agree but I wanted to keep new tests consistent with the existing ones so I kept it this way

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
mzgreen: I made the changes but I'm still not sure if it's ok. There are a couple of things that may throw:
1. Reading the db version
2. Calling `getReadable/WritableDatabase()`

Take a look please and let me know if you'd like to change anything

dlam:
It looks like you may want to rebase your changes here or some kind of weird merge issue? The diff is really large now and includes a bunch of changes outside of room.

mzgreen:
I did a rebase and now Github shows more changes than it should 😕 
Seems like similar issue: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch
There is no other branch that I could use to try the suggested workaround.

Not sure what can I do about it. If I try to open a new PR then Github correctly shows my changes but I don't know how to refresh this one.

danysantiago:
You can try cherry-picking only the relevant commits on top a new checkout of androidx force-pushing to the branch of this PR. This also seems relevant: https://stackoverflow.com/questions/16306012/github-pull-request-showing-commits-that-are-already-in-target-branch/45286781#45286781, but beware I'm not a GitHub guru. xD

dlam:
If you want to give me force push access on your fork as well, I can try to fix it :smiley:

dlam:
Should be fixed now, let me know if I missed anything - the issue was mixing merge + rebase flows :slightly_smiling_face: 

i.e., tried to rebase a merge commit on top of origin/androidx-master-dev

mzgreen:
I see, thanks @dlam !

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/PrepackageTest.java
```diff
@@ -593,4 +663,14 @@ private static void copyStream(InputStream input, OutputStream output) throws IO
             output.close();
         }
     }
+
+    public static class TestPrepackagedCallback extends RoomDatabase.PrepackagedCallback {
+
+        boolean mOpenPrepackagedDatabase;
```
danysantiago: Lets make this an int and increment it every time the callback is called, that way we can assert it gets called exactly once.

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -211,10 +207,66 @@ private void copyDatabaseFile(File destinationFile) throws IOException {
...
+        try {
+            return DBUtil.readVersion(databaseFile);
+        } catch (IOException e) {
+            Log.w(Room.LOG_TAG, "Unable to read database version.", e);
+            return null;
+        }
+    }
+
+    private void callOnOpenPrepackagedDatabaseCallbacks(File databaseFile, boolean writable) {
+        if(mDatabaseConfiguration.prepackagedCallback != null) {
```
danysantiago: * missing space after `if`
* Also invert the condition so that we can return early and save indentation:
```
if (mDatabaseConfiguration.prepackagedCallback == null) {
  return
}
```

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -211,10 +207,66 @@ private void copyDatabaseFile(File destinationFile) throws IOException {
...
+    private Integer readDatabaseVersion(File databaseFile) {
+        try {
+            return DBUtil.readVersion(databaseFile);
+        } catch (IOException e) {
+            Log.w(Room.LOG_TAG, "Unable to read database version.", e);
+            return null;
+        }
+    }
+
+    private void callOnOpenPrepackagedDatabaseCallbacks(File databaseFile, boolean writable) {
```
danysantiago: Maybe call method `dispatchOnOpenPrepackagedDatabase`

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -211,10 +207,66 @@ private void copyDatabaseFile(File destinationFile) throws IOException {
...
+        if(mDatabaseConfiguration.prepackagedCallback != null) {
+
+            SupportSQLiteOpenHelper helper = createFrameworkOpenHelper(databaseFile);
+            try {
+                SupportSQLiteDatabase db = writable ? helper.getWritableDatabase() :
+                        helper.getReadableDatabase();
+
+                mDatabaseConfiguration.prepackagedCallback.onOpenPrepackagedDatabase(db);
+            } catch (SQLiteException e) {
```
danysantiago: `SQLiteException` is an unchecked exception, lets just not catch it and let it propagate up, since wrapping it in a RuntimeException makes it harder for the user to handle it.

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -211,10 +207,66 @@ private void copyDatabaseFile(File destinationFile) throws IOException {
...
+                // Close the db and let Room re-open it through a normal path
+                helper.close();
+            }
+        }
+    }
+
+    private SupportSQLiteOpenHelper createFrameworkOpenHelper(File databaseFile) {
+        String databaseName = databaseFile.getName();
+        Integer version = readDatabaseVersion(databaseFile);
+        if(version == null) {
```
danysantiago: nit: missing space in if: `if (version == null) {`

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -211,10 +207,66 @@ private void copyDatabaseFile(File destinationFile) throws IOException {
...
+            throw new RuntimeException("Unable to invoke pre-packaged callback.");
```
danysantiago: Let's change the message to 'Malformed database file, unable to read version'. It would also be better to use the `DBUtil` method directly here and catch the IOException so that you can propagate it with the new RuntimeException, it provides a bit more info if something goes wrong.

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
```diff
@@ -211,10 +207,66 @@ private void copyDatabaseFile(File destinationFile) throws IOException {
...
+        callOnOpenPrepackagedDatabaseCallbacks(intermediateFile, writable);
+
         if (!intermediateFile.renameTo(destinationFile)) {
             throw new IOException("Failed to move intermediate file ("
                     + intermediateFile.getAbsolutePath() + ") to destination ("
                     + destinationFile.getAbsolutePath() + ").");
         }
     }
+
+    private Integer readDatabaseVersion(File databaseFile) {
```
danysantiago: Not sure this method is necessary, in `callOnOpenPrepackagedDatabaseCallbacks` where it is now newly used we should use the `DBUtil` version so that we can add the `IOException` as the cause to the `RuntimeException`.

room/runtime/src/main/java/androidx/room/RoomDatabase.java
```diff
@@ -654,6 +714,38 @@ private static boolean isLowRamDevice(@NonNull ActivityManager activityManager)
...
+         *
+         * @param inputStreamCallable A callable that returns an InputStream from which to copy
+         *                            the database.
+         * @param callback The pre-packaged callback.
+         *
+         * @return This {@link Builder} instance.
+         */
+        @NonNull
+        @SuppressLint("BuilderSetStyle") // To keep naming consistency, methods
+        // pre-dates rule.
```
danysantiago: nit: I think `// pre-dates rule.` fits in the previous line.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/PrepackageTest.java
```diff
@@ -566,6 +571,156 @@ public void openDataDirDatabase_versionZero_badSchema() throws IOException {
...
+
+    @Test
+    public void onOpenedDbTwice_calledOnPrepackagedCallbackOnce() {
+        Context context = ApplicationProvider.getApplicationContext();
+        context.deleteDatabase("products.db");
+
+        ProductsDatabase db1 = null;
+        ProductsDatabase db2 = null;
+        try {
+            TestPrepackagedCallback callback1 = new TestPrepackagedCallback();
```
danysantiago: If we update `TestPrepackagedCallback` to use an int for counting everytime the callback is called then we don't need `callback2` making this slightly simpler.

danysantiago (changes requested):
This is looking great, just a few more comments.

room/runtime/src/main/java/androidx/room/SQLiteCopyOpenHelper.java
mzgreen: Is there an auto formatting task that I could run or maybe code formatting config file that I could import in AS? The default one seems to be using different rules.

danysantiago: I'm not sure, the Studio that is started with `studiow` should have the style configuration. I know we also have some tasks to run `ktlint` but I run them when uploading to Gerrit using the `repo` tool which is different from this Github flow. @dlam might know more about this. It's possible this might not be setup yet, sorry.

danysantiago (approved):
Looks good to me, thanks!

tikurahul (approved):

mzgreen:
Thanks!

Not sure if it matters but the description at the top contains an error. It says that tests were added to `DatabaseCallbackTest` but I later moved them to `PrepackageTest` file and forgot to update Github description. This description was copied to Gerrit, I tried to edit it there but I can't (probably missing permission).

dlam:
Does your gerrit account use the same e-mail as your git commit author? Copybara not updating the gerrit CL description sounds like a bug, thanks for letting us know.

dlam (approved):

dlam:
Looks like Copybara was just waiting for review approvals before syncing.

dlam:
The CL in gerrit seems to have been updated with the new description now.

mzgreen:
I think Copybara works fine. I was trying to update the description on Gerrit before and it didn't work. Now after your message I realized that Copybara will copy it from Github if I change it here so I updated it and it worked. I understand the process now.

PR #57, opened 2020-08-06
Update build.gradle

FunkyMuse:
Since the main java library is bumped to API 21 this can too because when using this version the manifest wants to override the library's SDK if using < 23

Whereas the https://github.com/androidx/androidx/blob/androidx-master-dev/security/crypto/build.gradle
uses minSdkVersion 21

Test: ./gradlew bOS

### DIFF
--- a/security/security-crypto-ktx/build.gradle
+++ b/security/security-crypto-ktx/build.gradle
@@ -39,7 +39,7 @@ dependencies {
 
 android {
     defaultConfig {
-        minSdkVersion 23
+        minSdkVersion 21
     }
 }
 


tikurahul:
We are not going to be able to accept contributions for the `security` library, as out GitHub workflows currently only support WorkManager and Room. 

You can make the change here: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:security/security-crypto-ktx/build.gradle;l=1

Click on the `Edit Code` button, and that will create a Gerrit change.

PR #58, opened 2020-08-07
Fix OUT_DIR default to path that exists within GH checkout

dlam:
Previously, `OUT_DIR` defaulted to a path that exists outside of the root folder for github checkout (it assumes this repo is nested in frameworks/support). This allows playground projects to ensure the OUT_DIR is writable when it is not specified.

Test: cd room && ./gradlew bOS

### DIFF
--- a/playground-common/out.gradle
+++ b/playground-common/out.gradle
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019 The Android Open Source Project
+ * Copyright (C) 2020 The Android Open Source Project
  *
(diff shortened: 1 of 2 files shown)


tikurahul (approved):

playground-common/out.gradle
```diff
@@ -0,0 +1,39 @@
+/*
+ * Copyright (C) 2019 The Android Open Source Project
```
yigit: 2020

PR #59, opened 2020-08-07
Add playground support for paging

dlam:
Test: ./gradlew bOS


playground-common/playground-include-settings.gradle
```diff
@@ -64,10 +64,16 @@ def setupPlayground(settings, String relativePathToRoot) {
...
+    settings.includeProject(":annotation:annotation-sampled",
+            new File(supportRoot, "annotation/annotation-sampled"))
     settings.includeProject(":lint-checks", new File(supportRoot, "lint-checks"))
     settings.includeProject(":fakeannotations", new File(supportRoot,"fakeannotations"))
     settings.includeProject(":internal-testutils-common",
             new File(supportRoot, "testutils/testutils-common"))
+    settings.includeProject(":internal-testutils-ktx",
+            new File(supportRoot, "testutils/testutils-ktx"))
+    settings.includeProject(":internal-testutils-paging",
+            new File(supportRoot, "testutils/testutils-paging"))
```
yigit: why is paging necessary for all, seems like these should go into the settings of paging.

.github/workflows/presubmit.yml
```diff
@@ -69,7 +69,9 @@ jobs:
         run: echo $DIST_DIR && ls $HOME/Library/Android/sdk
       - name: build on server
         run: |
-          cd room && ./gradlew :buildOnServer && cd ../work && ./gradlew :buildOnServer
+          cd paging && ./gradlew :buildOnServer
```
yigit: this wouldn't work since you won't end up in the right folder.

paging/.idea/scopes/buildSrc.xml
```diff
@@ -0,0 +1 @@
+../../../.idea/scopes/buildSrc.xml
```
yigit, line 1: these seems undesired or did you re-run the setup script and these are now tracked in .git while they werent before?

yigit (changes requested):

playground-common/playground-include-settings.gradle
dlam: I didn't realize that works. We should probably just declare all of these explicitly in each project to minimize modules loaded

paging/.idea/scopes/buildSrc.xml
dlam: ah thanks for catching this - i ran the script and I think I forgot to setup gitignore?

.github/workflows/presubmit.yml
dlam: Fixed now.

paging/.idea/scopes/buildSrc.xml
dlam: Actually these were added by script, but I don't think this particular file should be here

dlam: These files should be getting ignored by .gitignore but I'm getting:
```
~/p/a/paging (dlam/paging-playground|:) git ls-files -v ../.idea
H ../.idea/codeStyles/Project.xml
H ../.idea/codeStyles/codeStyleConfig.xml
H ../.idea/copyright/AndroidCopyright.xml
H ../.idea/copyright/profiles_settings.xml
H ../.idea/inspectionProfiles/Project_Default.xml
H ../.idea/scopes/Compose.xml
H ../.idea/scopes/Ignore_API_Files.xml
H ../.idea/scopes/buildSrc.xml
H ../.idea/vcs.xml
```

If I touch normal empty files, gitignore ignores them successfully, this is a bit puzzling :|

dlam:
```
~/p/a/paging (dlam/paging-playground|:) git ls-files -i ../.idea/* --exclude-standard
../.idea/inspectionProfiles/Project_Default.xml
../.idea/scopes/Compose.xml
../.idea/scopes/Ignore_API_Files.xml
../.idea/scopes/buildSrc.xml
```

dlam: I'm tempted to say we should keep them because changes to those files actually do show up and get checked in even there's no explicit exclusion rule to allow them to not be ignored in .gitignore. bit confused here, but changes to the original files are being tracked.

yigit (approved):

PR #63, opened 2020-08-13
Add paging to list of contributable projects

dlam:
Test: N/A doc change only


tikurahul (approved):

mzgreen:
Readme also contains a section with contributable projects. I think it should be updated as well.

PR #64, opened 2020-08-17
Switch to using labels to retrigger a workflow.

tikurahul:
* `pull_request_review_comment`s happen a lot more frequently than labelling a PR.
* This method is easier than having to write an action that invokes a `bot`.

Test: Workflows run on GitHub.

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -2,9 +2,7 @@ name: AndroidX Presubmits
 on:
   push:
   pull_request_target:
-    types: [opened, synchronize, reopened]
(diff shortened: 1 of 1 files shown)


dlam:
What happens if we're retrying a change that has already been retried? Do we need to remove the label then add it back?

tikurahul:
Yes, that is correct.

dlam:
:eyes: okay, I guess that's still better than re-running on every comment

dlam (approved):

dlam:
I like removing the review triggers at least

PR #73, opened 2020-08-24
Add docs about running Ktlint

simonschiller:
## Proposed Changes

  - Add docs about running Ktlint

## Testing

Test: N/A, docs change only

## Issues Fixed

Fixes: -

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -84,10 +84,7 @@ cd room
 ./gradlew test connectedCheck
 
 # Run additional checks
-./gradlew buildOnServer
-
-# Run coding standard & guideline checks
-./gradlew ktlint
+./gradlew buildOnServer ktlint
 
 # If you are testing on an emulator, you can disable benchmark tests as
 # follows since they require a real device to run
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -86,6 +86,9 @@ cd room
 # Run additional checks
 ./gradlew buildOnServer
 
+# Run coding standard & guideline checks
+./gradlew ktlint
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -86,6 +86,9 @@ cd room
 # Run additional checks
 ./gradlew buildOnServer
 
+# Run coding standard & guideline checks
```
yigit: @tikurahul @liutikas 
Do we know why `bOS` doesn't cover this?

tikurahul: `ktlint` checks are performed using git hooks in AOSP. These hooks are installed automatically. We should try and do something similar.

dlam: We can provide links to the commit hooks we use, but it's impossible to force someone to install them, so we should just add the check to our build imo

dlam: Also is ktlint sufficient? Don't we also have error-prone, etc?

dlam: Can just delete this header and combine it with the additional checks section.

dlam:
Let me spend a bit of time investigating why ktlint isn't running before merging this - https://issuetracker.google.com/166128980

oramirezd:
Hello!

oramirezd:
I want to download all this file but I can not download all the archive.Just download for example one part of the file.80mb or 90mb and then stop the app to download to I use.

oramirezd:
My phone number it's +5356404000 I am in Watsapp.

oramirezd:
My email it's oramirezd@nauta.cu

dlam:
This PR is superseded by actually integrating ktlint in presubmit here: https://github.com/androidx/androidx/pull/106

PR #75, opened 2020-09-02
Add Provided Type Converter feature

mzgreen:
This change introduces a new `@ProvidedTypeConverter` annotation that gives developers control over type converter instantiation so they can use external dependencies inside type converters. A type converter class annotated with `@ProvidedTypeConverter` won't be instantiated by Room. Room will expect such converters to be passed at runtime using new `builder.addTypeConverter()` method.

Design doc: https://docs.google.com/document/d/1nmXZUsSlH6WWsT3Duxnkd4LcAPZGcyVCwJe5scnh7HU/edit?usp=sharing
Test: Added `ProvidedTypeConverterTest`
Fixes: b/121067210
RelNote: Room can have provided type converter annotation which allow developers to take control over type converters instantiation

### DIFF
--- a/room/common/src/main/java/androidx/room/ProvidedTypeConverter.java
+++ b/room/common/src/main/java/androidx/room/ProvidedTypeConverter.java
@@ -16,21 +16,15 @@
 
 package androidx.room;
 
-import androidx.annotation.NonNull;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
 /**
- * Implementations of {@code TypeConverterFactory} interface are responsible to instantiate
- * TypeConverters.
+ * Marks a class as a type converter that will be provided to Room at runtime.
  */
-public interface TypeConverterFactory {
-    /**
-     * Creates a new instance of the given {@code Class}.
-     * <p>
-     *
-     * @param converterClass a {@code Class} whose instance is requested
-     * @param <T> The type parameter for the TypeConverter.
-     * @return a newly created TypeConverter
-     */
-    @NonNull
-    <T> T create(@NonNull Class<T> converterClass);
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.CLASS)
+public @interface ProvidedTypeConverter {
 }
--- a/room/common/src/main/java/androidx/room/ProvidedTypeConverter.java
+++ b/room/common/src/main/java/androidx/room/ProvidedTypeConverter.java
@@ -23,8 +23,10 @@ import java.lang.annotation.Target;
 
 /**
  * Marks a class as a type converter that will be provided to Room at runtime.
- * If Room uses the annotated type converter class, it will verify that it is provided in the builder and if not, will throw
- * an exception.
+ * If Room uses the annotated type converter class, it will verify that it is provided in the
+ * builder and if not, will throw an exception.
+ * An instance of a class annotated with this annotation has to be provided to Room using
+ * {@code Room.databaseBuilder.addTypeConverter(Object)}
  */
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.CLASS)
--- a/room/common/src/main/java/androidx/room/ProvidedTypeConverter.java
+++ b/room/common/src/main/java/androidx/room/ProvidedTypeConverter.java
@@ -23,6 +23,8 @@ import java.lang.annotation.Target;
 
 /**
  * Marks a class as a type converter that will be provided to Room at runtime.
+ * If Room uses the annotated type converter class, it will verify that it is provided in the builder and if not, will throw
+ * an exception.
  */
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.CLASS)
--- a/room/common/src/main/java/androidx/room/TypeConverterFactory.java
+++ b/room/common/src/main/java/androidx/room/TypeConverterFactory.java
@@ -20,7 +20,7 @@ import androidx.annotation.NonNull;
 
 /**
  * Implementations of {@code TypeConverterFactory} interface are responsible to instantiate
- * TypeConverters.
+ * TypeConverters and must be final.
  */
 public interface TypeConverterFactory {
     /**
--- a/room/common/src/main/java/androidx/room/TypeConverterFactory.java
+++ b/room/common/src/main/java/androidx/room/TypeConverterFactory.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright (C) 2017 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.room;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Specifies type converter factory that Room can use to instantiate a class containing
+ * {@link TypeConverter} annotated methods.
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.CLASS)
+public @interface TypeConverterFactory {
+    /**
+     * The type converter factory class.
+     *
+     * @return The class of a {@code TypeConverterFactory } that should be used to instantiate a
+     * class that contains the converter methods.
+     */
+    Class<?> value();
+}
--- a/room/common/src/main/java/androidx/room/TypeConverterFactory.java
+++ b/room/common/src/main/java/androidx/room/TypeConverterFactory.java
@@ -20,7 +20,7 @@ import androidx.annotation.NonNull;
 
 /**
  * Implementations of {@code TypeConverterFactory} interface are responsible to instantiate
- * TypeConverters and must be final.
+ * TypeConverters.
  */
 public interface TypeConverterFactory {
     /**
--- a/room/common/src/main/java/androidx/room/TypeConverterFactory.java
+++ b/room/common/src/main/java/androidx/room/TypeConverterFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017 The Android Open Source Project
+ * Copyright 2020 The Android Open Source Project
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,23 +16,21 @@
 
 package androidx.room;
 
-import java.lang.annotation.ElementType;
-import java.lang.annotation.Retention;
-import java.lang.annotation.RetentionPolicy;
-import java.lang.annotation.Target;
+import androidx.annotation.NonNull;
 
 /**
- * Specifies type converter factory that Room can use to instantiate a class containing
- * {@link TypeConverter} annotated methods.
+ * Implementations of {@code TypeConverterFactory} interface are responsible to instantiate
+ * TypeConverters.
  */
-@Target(ElementType.TYPE)
-@Retention(RetentionPolicy.CLASS)
-public @interface TypeConverterFactory {
+public interface TypeConverterFactory {
     /**
-     * The type converter factory class.
+     * Creates a new instance of the given {@code Class}.
+     * <p>
      *
-     * @return The class of a {@code TypeConverterFactory } that should be used to instantiate a
-     * class that contains the converter methods.
+     * @param converterClass a {@code Class} whose instance is requested
+     * @param <T> The type parameter for the TypeConverter.
+     * @return a newly created TypeConverter
      */
-    Class<?> value();
+    @NonNull
+    <T> T create(@NonNull Class<T> converterClass);
 }
--- a/room/compiler/src/main/kotlin/androidx/room/DatabaseProcessingStep.kt
+++ b/room/compiler/src/main/kotlin/androidx/room/DatabaseProcessingStep.kt
@@ -40,7 +40,6 @@ class DatabaseProcessingStep : XProcessingStep {
         val databases = elementsByAnnotation[Database::class]
             ?.mapNotNull {
                 try {
-                    context.databaseElement = it.asTypeElement()
                     DatabaseProcessor(
                         context,
                         it.asTypeElement()
@@ -68,8 +67,7 @@ class DatabaseProcessingStep : XProcessingStep {
         }
 
         databases?.forEach { db ->
-            DatabaseWriter(db, context.getTypeConverterFactories(db.element)).write(context
-                .processingEnv)
+            DatabaseWriter(db).write(context.processingEnv)
             if (db.exportSchema) {
                 val schemaOutFolder = context.schemaOutFolder
                 if (schemaOutFolder == null) {
--- a/room/compiler/src/main/kotlin/androidx/room/DatabaseProcessingStep.kt
+++ b/room/compiler/src/main/kotlin/androidx/room/DatabaseProcessingStep.kt
@@ -40,6 +40,7 @@ class DatabaseProcessingStep : XProcessingStep {
         val databases = elementsByAnnotation[Database::class]
             ?.mapNotNull {
                 try {
+                    context.databaseElement = it.asTypeElement()
                     DatabaseProcessor(
                         context,
                         it.asTypeElement()
@@ -67,7 +68,8 @@ class DatabaseProcessingStep : XProcessingStep {
         }
 
         databases?.forEach { db ->
-            DatabaseWriter(db).write(context.processingEnv)
+            DatabaseWriter(db, context.getTypeConverterFactories(db.element)).write(context
+                .processingEnv)
             if (db.exportSchema) {
                 val schemaOutFolder = context.schemaOutFolder
                 if (schemaOutFolder == null) {
--- a/room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
+++ b/room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
@@ -38,7 +38,7 @@ class Context private constructor(
     private val inheritedAdapterStore: TypeAdapterStore?,
     val cache: Cache,
     private val canRewriteQueriesToDropUnusedColumns: Boolean,
-    val typeConverterFactories: MutableSet<XType>
+    private val addTypeConverterFactory: (XType) -> Unit
 ) {
     val checker: Checks = Checks(logger)
     val COMMON_TYPES = CommonTypes(processingEnv)
@@ -93,7 +93,7 @@ class Context private constructor(
             inheritedAdapterStore = null,
             cache = Cache(null, LinkedHashSet(), emptySet()),
             canRewriteQueriesToDropUnusedColumns = false,
-            typeConverterFactories = HashSet())
+            addTypeConverterFactory = {})
 
     class CommonTypes(val processingEnv: XProcessingEnv) {
         val VOID: XType by lazy {
@@ -124,15 +124,20 @@ class Context private constructor(
                 inheritedAdapterStore = typeAdapterStore,
                 cache = cache,
                 canRewriteQueriesToDropUnusedColumns = canRewriteQueriesToDropUnusedColumns,
-                typeConverterFactories = typeConverterFactories)
+                addTypeConverterFactory = addTypeConverterFactory)
         subContext.databaseVerifier = databaseVerifier
         val result = handler(subContext)
         return Pair(result, collector)
     }
 
-    fun fork(element: XElement, forceSuppressedWarnings: Set<Warning> = emptySet()): Context {
+    fun fork(
+        element: XElement,
+        forceSuppressedWarnings: Set<Warning> = emptySet(),
+        addTypeConverterFactory: (XType) -> Unit = {}
+    ): Context {
         val suppressedWarnings = SuppressWarningProcessor.getSuppressedWarnings(element)
-        val processConvertersResult = CustomConverterProcessor.findConverters(this, element)
+        val processConvertersResult =
+            CustomConverterProcessor.findConverters(this, element, addTypeConverterFactory)
         val canReUseAdapterStore = processConvertersResult.classes.isEmpty()
         // order here is important since the sub context should give priority to new converters.
         val subTypeConverters = if (canReUseAdapterStore) {
@@ -152,7 +157,7 @@ class Context private constructor(
                 inheritedAdapterStore = if (canReUseAdapterStore) typeAdapterStore else null,
                 cache = subCache,
                 canRewriteQueriesToDropUnusedColumns = subCanRemoveUnusedColumns,
-                typeConverterFactories = typeConverterFactories)
+                addTypeConverterFactory = addTypeConverterFactory)
         subContext.databaseVerifier = databaseVerifier
         return subContext
     }
--- a/room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
+++ b/room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
@@ -24,7 +24,6 @@ import androidx.room.preconditions.Checks
 import androidx.room.compiler.processing.XElement
 import androidx.room.compiler.processing.XProcessingEnv
 import androidx.room.compiler.processing.XType
-import androidx.room.compiler.processing.XTypeElement
 import androidx.room.processor.cache.Cache
 import androidx.room.solver.TypeAdapterStore
 import androidx.room.verifier.DatabaseVerifier
@@ -39,8 +38,7 @@ class Context private constructor(
     private val inheritedAdapterStore: TypeAdapterStore?,
     val cache: Cache,
     private val canRewriteQueriesToDropUnusedColumns: Boolean,
-    // currently processed database element to type converter factory map
-    private val typeConverterFactoriesMap: MutableMap<XTypeElement, MutableSet<XType>>
+    val typeConverterFactories: MutableSet<XType>
 ) {
     val checker: Checks = Checks(logger)
     val COMMON_TYPES = CommonTypes(processingEnv)
@@ -56,9 +54,6 @@ class Context private constructor(
     // set when database and its entities are processed.
     var databaseVerifier: DatabaseVerifier? = null
         private set
-
-    // set when database processing is started.
-    var databaseElement: XTypeElement? = null
 
     val queryRewriter: QueryRewriter by lazy {
         val verifier = databaseVerifier
@@ -91,19 +86,6 @@ class Context private constructor(
         this.databaseVerifier = databaseVerifier
     }
 
-    fun addTypeConverterFactory(typeConverterFactory: XType) {
-        val dbElement = databaseElement
-        if (dbElement != null) {
-            val set = typeConverterFactoriesMap[dbElement] ?: HashSet()
-            set.add(typeConverterFactory)
-            typeConverterFactoriesMap[dbElement] = set
-        }
-    }
-
-    fun getTypeConverterFactories(databaseElement: XTypeElement): Set<XType> {
-        return typeConverterFactoriesMap[databaseElement] ?: emptySet()
-    }
-
     constructor(processingEnv: XProcessingEnv) : this(
             processingEnv = processingEnv,
             logger = RLog(processingEnv.messager, emptySet(), null),
@@ -111,7 +93,7 @@ class Context private constructor(
             inheritedAdapterStore = null,
             cache = Cache(null, LinkedHashSet(), emptySet()),
             canRewriteQueriesToDropUnusedColumns = false,
-            typeConverterFactoriesMap = HashMap())
+            typeConverterFactories = HashSet())
 
     class CommonTypes(val processingEnv: XProcessingEnv) {
         val VOID: XType by lazy {
@@ -142,7 +124,7 @@ class Context private constructor(
                 inheritedAdapterStore = typeAdapterStore,
                 cache = cache,
                 canRewriteQueriesToDropUnusedColumns = canRewriteQueriesToDropUnusedColumns,
-                typeConverterFactoriesMap = typeConverterFactoriesMap)
+                typeConverterFactories = typeConverterFactories)
         subContext.databaseVerifier = databaseVerifier
         val result = handler(subContext)
         return Pair(result, collector)
@@ -170,7 +152,7 @@ class Context private constructor(
                 inheritedAdapterStore = if (canReUseAdapterStore) typeAdapterStore else null,
                 cache = subCache,
                 canRewriteQueriesToDropUnusedColumns = subCanRemoveUnusedColumns,
-                typeConverterFactoriesMap = typeConverterFactoriesMap)
+                typeConverterFactories = typeConverterFactories)
         subContext.databaseVerifier = databaseVerifier
         return subContext
     }
--- a/room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
+++ b/room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
@@ -17,13 +17,13 @@
 package androidx.room.processor
 
 import androidx.room.RewriteQueriesToDropUnusedColumns
-import androidx.room.compiler.processing.XElement
-import androidx.room.compiler.processing.XProcessingEnv
-import androidx.room.compiler.processing.XType
 import androidx.room.log.RLog
 import androidx.room.parser.expansion.ProjectionExpander
 import androidx.room.parser.optimization.RemoveUnusedColumnQueryRewriter
 import androidx.room.preconditions.Checks
+import androidx.room.compiler.processing.XElement
+import androidx.room.compiler.processing.XProcessingEnv
+import androidx.room.compiler.processing.XType
 import androidx.room.processor.cache.Cache
 import androidx.room.solver.TypeAdapterStore
 import androidx.room.verifier.DatabaseVerifier
@@ -127,13 +127,9 @@ class Context private constructor(
         return Pair(result, collector)
     }
 
-    fun fork(
-        element: XElement,
-        forceSuppressedWarnings: Set<Warning> = emptySet()
-    ): Context {
+    fun fork(element: XElement, forceSuppressedWarnings: Set<Warning> = emptySet()): Context {
         val suppressedWarnings = SuppressWarningProcessor.getSuppressedWarnings(element)
-        val processConvertersResult =
-            CustomConverterProcessor.findConverters(this, element)
+        val processConvertersResult = CustomConverterProcessor.findConverters(this, element)
         val canReUseAdapterStore = processConvertersResult.classes.isEmpty()
         // order here is important since the sub context should give priority to new converters.
         val subTypeConverters = if (canReUseAdapterStore) {
--- a/room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
+++ b/room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
@@ -17,13 +17,13 @@
 package androidx.room.processor
 
 import androidx.room.RewriteQueriesToDropUnusedColumns
+import androidx.room.compiler.processing.XElement
+import androidx.room.compiler.processing.XProcessingEnv
+import androidx.room.compiler.processing.XType
 import androidx.room.log.RLog
 import androidx.room.parser.expansion.ProjectionExpander
 import androidx.room.parser.optimization.RemoveUnusedColumnQueryRewriter
 import androidx.room.preconditions.Checks
-import androidx.room.compiler.processing.XElement
-import androidx.room.compiler.processing.XProcessingEnv
-import androidx.room.compiler.processing.XType
 import androidx.room.processor.cache.Cache
 import androidx.room.solver.TypeAdapterStore
 import androidx.room.verifier.DatabaseVerifier
@@ -37,8 +37,7 @@ class Context private constructor(
     private val typeConverters: CustomConverterProcessor.ProcessResult,
     private val inheritedAdapterStore: TypeAdapterStore?,
     val cache: Cache,
-    private val canRewriteQueriesToDropUnusedColumns: Boolean,
-    private val addTypeConverterFactory: (XType) -> Unit
+    private val canRewriteQueriesToDropUnusedColumns: Boolean
 ) {
     val checker: Checks = Checks(logger)
     val COMMON_TYPES = CommonTypes(processingEnv)
@@ -92,8 +91,7 @@ class Context private constructor(
             typeConverters = CustomConverterProcessor.ProcessResult.EMPTY,
             inheritedAdapterStore = null,
             cache = Cache(null, LinkedHashSet(), emptySet()),
-            canRewriteQueriesToDropUnusedColumns = false,
-            addTypeConverterFactory = {})
+            canRewriteQueriesToDropUnusedColumns = false)
 
     class CommonTypes(val processingEnv: XProcessingEnv) {
         val VOID: XType by lazy {
@@ -123,8 +121,7 @@ class Context private constructor(
                 typeConverters = this.typeConverters,
                 inheritedAdapterStore = typeAdapterStore,
                 cache = cache,
-                canRewriteQueriesToDropUnusedColumns = canRewriteQueriesToDropUnusedColumns,
-                addTypeConverterFactory = addTypeConverterFactory)
+                canRewriteQueriesToDropUnusedColumns = canRewriteQueriesToDropUnusedColumns)
         subContext.databaseVerifier = databaseVerifier
         val result = handler(subContext)
         return Pair(result, collector)
@@ -132,12 +129,11 @@ class Context private constructor(
 
     fun fork(
         element: XElement,
-        forceSuppressedWarnings: Set<Warning> = emptySet(),
-        addTypeConverterFactory: (XType) -> Unit = {}
+        forceSuppressedWarnings: Set<Warning> = emptySet()
     ): Context {
         val suppressedWarnings = SuppressWarningProcessor.getSuppressedWarnings(element)
         val processConvertersResult =
-            CustomConverterProcessor.findConverters(this, element, addTypeConverterFactory)
+            CustomConverterProcessor.findConverters(this, element)
         val canReUseAdapterStore = processConvertersResult.classes.isEmpty()
         // order here is important since the sub context should give priority to new converters.
         val subTypeConverters = if (canReUseAdapterStore) {
@@ -156,8 +152,7 @@ class Context private constructor(
                 typeConverters = subTypeConverters,
                 inheritedAdapterStore = if (canReUseAdapterStore) typeAdapterStore else null,
                 cache = subCache,
-                canRewriteQueriesToDropUnusedColumns = subCanRemoveUnusedColumns,
-                addTypeConverterFactory = addTypeConverterFactory)
+                canRewriteQueriesToDropUnusedColumns = subCanRemoveUnusedColumns)
         subContext.databaseVerifier = databaseVerifier
         return subContext
     }
--- a/room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
+++ b/room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
@@ -24,6 +24,7 @@ import androidx.room.preconditions.Checks
 import androidx.room.compiler.processing.XElement
 import androidx.room.compiler.processing.XProcessingEnv
 import androidx.room.compiler.processing.XType
+import androidx.room.compiler.processing.XTypeElement
 import androidx.room.processor.cache.Cache
 import androidx.room.solver.TypeAdapterStore
 import androidx.room.verifier.DatabaseVerifier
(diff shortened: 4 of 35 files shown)


room/compiler/src/main/kotlin/androidx/room/DatabaseProcessingStep.kt
```diff
@@ -67,7 +68,8 @@ class DatabaseProcessingStep : XProcessingStep {
         }
 
         databases?.forEach { db ->
-            DatabaseWriter(db).write(context.processingEnv)
+            DatabaseWriter(db, context.getTypeConverterFactories(db.element)).write(context
```
yigit: you probably want DatabaseProcessor to extract the factory information to the Database value object it creates (instead of parsing here).
the validation (mostly) happens on the Processors.

room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
```diff
@@ -37,7 +38,9 @@ class Context private constructor(
     private val typeConverters: CustomConverterProcessor.ProcessResult,
     private val inheritedAdapterStore: TypeAdapterStore?,
     val cache: Cache,
-    private val canRewriteQueriesToDropUnusedColumns: Boolean
+    private val canRewriteQueriesToDropUnusedColumns: Boolean,
+    // currently processed database element to type converter factory map
+    private val typeConverterFactoriesMap: MutableMap<XTypeElement, MutableSet<XType>>
```
yigit: i'm not sure why we are keeping this in the context but i need to check more in detail (going into a meeting now :) )

yigit (commented):
partial review, thanks for sending this!

room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
yigit: ok i guess this is here because type converters were initially kept in here because they are scoped to whereever the annotation is used.
If possible, it feels like it would be cleaner if the database processor collected each `used` converter from its DAO's and entities then keep that list in the `Database` value object (to verify later in the generated code that all used ones are provided).

room/compiler/src/main/kotlin/androidx/room/processor/CustomConverterProcessor.kt
```diff
@@ -77,29 +78,38 @@ class CustomConverterProcessor(val context: Context, val element: XTypeElement)
...
-                    constructors.any {
-                        it.parameters.isEmpty()
-                    }, element, TYPE_CONVERTER_MISSING_NOARG_CONSTRUCTOR
-        )
+        if (typeConverterFactory == null) {
+            context.checker.check(
+                isKotlinObjectDeclaration || allStatic || constructors.isEmpty() ||
+                        constructors.any {
+                            it.parameters.isEmpty()
+                        }, element, TYPE_CONVERTER_MISSING_NOARG_CONSTRUCTOR
```
yigit, line 92: we probably want to update this message (in processor errors) to refer to the type converter factory if they can provide the instance.

room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
mzgreen: So since it's not possible to give a compile time validation for converter factories, I wanted to perform 2 checks at runtime (when `DatabaseBuilder.build()` is called). Those checks are:

1.  User added `@TypeConverterFactory(FooFactory.class)` on a type converter class but forgot to add the instance of `FooFactory` to database builder.

2. User provided an instance of `FooFactory` to database builder but there is no type converter class annotated with `@TypeConverterFactory(FooFactory.class)`.

In order to do that I needed an information about all `@TypeConverterFactory` annotations to be available in `RoomDatabase` file. I figured out that I can use `DatabaseWriter` to put this info there. `Context` already keeps track of type converters but it's a hierarchical data structure and at `DatabaseProcessor` level I have access only to the top level context. It's not possible to extract this information having access only to this context so I added a "flat" data structure that keeps track of all type converter factories.
Not sure what does it mean that a converter is being `used` or not but currently factories are added to a set inside a context. The logic of adding the factory to this set is in `CustomConverterProcessor` file. I'm not sure if this guarantees that a converter is being `used`. It was the only place I've found that allowed me to implement this feature. I'm open to suggestions though :)

room/compiler/src/main/kotlin/androidx/room/DatabaseProcessingStep.kt
mzgreen: Sure I'll move it there!

room/common/src/main/java/androidx/room/TypeConverterFactory.java
```diff
@@ -0,0 +1,38 @@
...
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.CLASS)
+public @interface TypeConverterFactory {
+    /**
+     * The type converter factory class.
+     *
+     * @return The class of a {@code TypeConverterFactory } that should be used to instantiate a
+     * class that contains the converter methods.
+     */
+    Class<?> value();
```
danysantiago: We can make this `Class<? extends TypeConverterFactory>` if we move the interface to the `common`, make it a top level class (not an inner class of RoomDatabase). Meanwhile to avoid conflict with this annotation, lets make this annotation an inner class of `@TypeConverter` and call it `Factory` so that the usage is then:
```
@TypeConverter.Factory(TimeStampConverterFactory.class)
public static class TimeStampConverter {
```

room/runtime/src/main/java/androidx/room/RoomDatabase.java
```diff
@@ -178,6 +190,27 @@ public void init(@NonNull DatabaseConfiguration configuration) {
             mInvalidationTracker.startMultiInstanceInvalidation(configuration.context,
                     configuration.name);
         }
+
+        if(configuration.typeConverterFactories != null) {
+            for (Map.Entry<String, TypeConverterFactory> entry :
+                    configuration.typeConverterFactories.entrySet()) {
+                if (!mTypeConverterFactories.containsKey(entry.getKey())) {
+                    throw new IllegalArgumentException("Unexpected factory " + entry.getKey() + ". "
+                            + "Annotate TypeConverter class with @TypeCoverterFactory annotation "
```
danysantiago: typo in `@TypeCoverterFactory`

room/runtime/src/main/java/androidx/room/RoomDatabase.java
```diff
@@ -178,6 +190,27 @@ public void init(@NonNull DatabaseConfiguration configuration) {
...
+                }
+                mTypeConverterFactories.put(entry.getKey(), entry.getValue());
+            }
+        }
+
+        for(Map.Entry<String, TypeConverterFactory> entry: getTypeConverterFactories().entrySet()) {
+            if (entry.getValue() == null) {
+                throw new IllegalArgumentException("Missing " + entry.getKey() + " factory "
+                        + "instance. Add it using addTypeConverterFactory method or "
+                        + "remove unnecessaty @TypeConverterFactory annotation from "
```
danysantiago: typo in `unnecessaty`

room/compiler/src/main/kotlin/androidx/room/writer/DatabaseWriter.kt
```diff
@@ -59,11 +65,43 @@ class DatabaseWriter(val database: Database) : ClassWriter(database.implTypeName
...
             addMethod(createCreateInvalidationTracker())
             addMethod(createClearAllTables())
+            addMethod(createCreateTypeConverterFactoriesMap())
         }
         addDaoImpls(builder)
         return builder
     }
 
+    private fun createCreateTypeConverterFactoriesMap():
+            MethodSpec {
```
danysantiago: return type can be in same line as method name

room/compiler/src/main/kotlin/androidx/room/writer/DatabaseWriter.kt
```diff
@@ -59,11 +65,43 @@ class DatabaseWriter(val database: Database) : ClassWriter(database.implTypeName
...
+        val typeConverterFactoryType = ClassName.get(TypeConverterFactory::class.java)
```
danysantiago: Instead of `ClassName.get(TypeConverterFactory::class.java)` use a constant defined here: https://github.com/androidx/androidx/blob/androidx-master-dev/room/compiler/src/main/kotlin/androidx/room/ext/javapoet_ext.kt#L53

room/runtime/src/main/java/androidx/room/RoomDatabase.java
```diff
@@ -1012,6 +1058,21 @@ private static boolean isLowRamDevice(@NonNull ActivityManager activityManager)
...
+         *
+         * @param factory The factory.
+         * @return This {@link Builder} instance.
+         */
+        @NonNull
+        public Builder<T> addTypeConverterFactory(@NonNull TypeConverterFactory factory) {
+            if(mTypeConverterFactories == null) {
+                mTypeConverterFactories = new HashMap<>();
+            }
+            mTypeConverterFactories.put(factory.getClass().getCanonicalName(), factory);
```
danysantiago: Since we use string as class identifier names we need to add a proguard rule to avoid obfuscation.
The rule will probably look like this:
`-keepclasseswithmembernames class * implements androidx.room.TypeConverterFactory`
to keep the name of classes that implement the factory interface.

room/compiler/src/main/kotlin/androidx/room/writer/DatabaseWriter.kt
```diff
@@ -59,11 +65,43 @@ class DatabaseWriter(val database: Database) : ClassWriter(database.implTypeName
...
+        return MethodSpec.methodBuilder("createTypeConverterFactoriesMap").apply {
+            addAnnotation(Override::class.java)
+            addModifiers(PROTECTED)
+            returns(ParameterizedTypeName.get(CommonTypeNames.MAP, CommonTypeNames.STRING,
+                typeConverterFactoryType))
+            val typeConverterFactoriesVar = "_typeConverterFactoriesMap"
```
danysantiago: Use `val typeConverterFactoriesVar = scope.getTmpVar("_typeConverterFactoriesMap")` to create a local scoped variable name without conflicts.

room/runtime/src/main/java/androidx/room/DatabaseConfiguration.java
```diff
@@ -363,6 +367,59 @@ public DatabaseConfiguration(@NonNull Context context, @Nullable String name,
             @Nullable File copyFromFile,
             @Nullable Callable<InputStream> copyFromInputStream,
             @Nullable RoomDatabase.PrepackagedDatabaseCallback prepackagedDatabaseCallback) {
+        this(context, name, sqliteOpenHelperFactory, migrationContainer, callbacks,
```
danysantiago, line 376: This constructor must now be deprecated, add `@Deprecated` and in the method doc add a `@deprecated` tag, see examples above. All other constructor bodies must be updated to call the newest constructor and their `@deprecated` tag updated.

room/compiler/src/main/kotlin/androidx/room/solver/types/CustomTypeConverterWrapper.kt
```diff
@@ -45,13 +50,75 @@ class CustomTypeConverterWrapper(val custom: CustomTypeConverter) :
...
+                        "$L = $N.$L($L)",
+                        outputVarName, typeConverter(scope),
+                        custom.methodName, inputVarName
+                    )
+                }
             }
         }
     }
 
+    fun typeConverterFactory(scope: CodeGenScope, enclosingClassFactory: XType): MethodSpec {
```
danysantiago: Can be private

room/compiler/src/main/kotlin/androidx/room/solver/types/CustomTypeConverterWrapper.kt
```diff
@@ -45,13 +50,75 @@ class CustomTypeConverterWrapper(val custom: CustomTypeConverter) :
...
+                return "converter_${custom.typeName}"
+            }
+
+            override fun prepare(writer: ClassWriter, builder: FieldSpec.Builder) {
+                builder.addModifiers(Modifier.PRIVATE)
+            }
+        })
+
+        return scope.writer.getOrCreateMethod(object : ClassWriter.SharedMethodSpec(
+            baseName) {
```
danysantiago: nit: `baseName) {` can be placed in the line before

room/compiler/src/main/kotlin/androidx/room/solver/types/CustomTypeConverterWrapper.kt
```diff
@@ -45,13 +50,75 @@ class CustomTypeConverterWrapper(val custom: CustomTypeConverter) :
...
+                    addCode(buildConvertMethodBody(writer))
+                }
+            }
+
+            private fun buildConvertMethodBody(
+                writer: ClassWriter
+            ): CodeBlock {
+                val methodScope = CodeGenScope(writer)
+                methodScope.builder().apply {
+                    beginControlFlow("if($N == null)", converterField)
```
danysantiago: nit: should produce a space between `if` and `(`

room/compiler/src/main/kotlin/androidx/room/solver/types/CustomTypeConverterWrapper.kt
```diff
@@ -45,13 +50,75 @@ class CustomTypeConverterWrapper(val custom: CustomTypeConverter) :
...
+                return "converterMethod_${custom.typeName}"
+            }
+
+            override fun prepare(
+                methodName: String,
+                writer: ClassWriter,
+                builder: MethodSpec.Builder
+            ) {
+                builder.apply {
+                    addModifiers(Modifier.PRIVATE)
```
danysantiago, line 96: add a `Modifier.SYNCHRONIZED` since converter instantiation can race between async queries trying to use it.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
```diff
@@ -0,0 +1,151 @@
...
+public class TypeConverterFactoryTest {
+
+    @Test
+    public void testTypeConverterFactory() {
+        Context context = ApplicationProvider.getApplicationContext();
+        context.deleteDatabase("test");
+        TestDatabaseWithConverter db = null;
+        Throwable throwable = null;
+        try {
+            db = Room.databaseBuilder(context, TestDatabaseWithConverter.class, "test")
```
danysantiago: Use the `inMemoryDatabaseBuilder` and then you can remove the codes regarding deleting the database file and it isolates the test more. Meanwhile, no need to surround `db` initialization with a try-catch, if something goes wrong the test will fail due to an exception opening or closing the DB, same as what you have now, but its just less noise in the test.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
```diff
@@ -0,0 +1,151 @@
...
+                    .build();
+            Pet pet = TestUtil.createPet(3);
+            pet.setName("pet");
+            db.getPetDao().insertOrReplace(pet);
```
danysantiago: add a `fail("Show have thrown an IllegalStateException")` at the end of the `try` to make sure we expect an exception

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
```diff
@@ -0,0 +1,151 @@
...
+        } catch (Throwable t) {
+            throwable = t;
+        } finally {
+            if (db != null) {
+                db.close();
+            }
+        }
+        Assert.assertThat(throwable, instanceOf(IllegalArgumentException.class));
```
danysantiago: This can be cleaned up a bit by not closing the DB (no need to since it should throw) and asserting the exception in the `catch` block

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
```diff
@@ -0,0 +1,151 @@
...
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import java.util.Date;
+
+@MediumTest
+@RunWith(AndroidJUnit4.class)
+public class TypeConverterFactoryTest {
```
danysantiago: Can you please add a test for two different converters relying on the same factory?

danysantiago (changes requested):

room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
danysantiago: I think we want to move the set of factories to the value object we have to represent a processed database: `androidx.room.vo.Database`. I do still think we need the `context` to gather them since they can be scoped to entity fields, daos, even POJOs so its not easy to gather those from the `DatabaseProcessor`. But what we can do is keep it more local by creating the map of factories in `DatabaseProcessor` and clearing it from the context once we are done and have created a `androidx.room.vo.Database`. It also means we can remove the mutable property `databaseElement` that was added. `DaoWritter` would also stay the same, since the info is in the value object.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
mzgreen: I modified the existing one to use 2 converters and single factory. Hope it's fine :)

room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
mzgreen: I implemented @danysantiago suggestions. Let me know if it's fine now :)

room/runtime/src/main/java/androidx/room/RoomDatabase.java
```diff
@@ -1012,6 +1058,21 @@ private static boolean isLowRamDevice(@NonNull ActivityManager activityManager)
...
 
+        /**
+         * Adds a {@link TypeConverterFactory} to this database.
+         *
+         * @param factory The factory.
+         * @return This {@link Builder} instance.
+         */
+        @NonNull
+        public Builder<T> addTypeConverterFactory(@NonNull TypeConverterFactory factory) {
+            if(mTypeConverterFactories == null) {
```
danysantiago: nit: missing space between `if` and `(`

room/runtime/src/main/java/androidx/room/RoomDatabase.java
```diff
@@ -178,6 +190,27 @@ public void init(@NonNull DatabaseConfiguration configuration) {
             mInvalidationTracker.startMultiInstanceInvalidation(configuration.context,
                     configuration.name);
         }
+
+        if(configuration.typeConverterFactories != null) {
```
danysantiago: nit: missing space between `if` and `(`

room/runtime/src/main/java/androidx/room/RoomDatabase.java
```diff
@@ -178,6 +190,27 @@ public void init(@NonNull DatabaseConfiguration configuration) {
...
+                if (!mTypeConverterFactories.containsKey(entry.getKey())) {
+                    throw new IllegalArgumentException("Unexpected factory " + entry.getKey() + ". "
+                            + "Annotate TypeConverter class with @TypeConverter.Factory annotation "
+                            + "or remove this factory from the builder.");
+                }
+                mTypeConverterFactories.put(entry.getKey(), entry.getValue());
+            }
+        }
+
+        for(Map.Entry<String, TypeConverterFactory> entry: getTypeConverterFactories().entrySet()) {
```
danysantiago: nit: missing space between `for` and `(`

room/runtime/src/main/java/androidx/room/RoomDatabase.java
danysantiago: One interesting issue here is that we build the map with the concrete factory class, so if a user provides a subclass of an expected factory Room won't be able to correctly use it / accept it. One solution to this is processing the factory type and making sure its `final`, i.e. making it a compiler error if its not. This is a bit aggressive IMHO, but would avoid any confusion since we can't quite support it.

room/compiler/src/main/kotlin/androidx/room/processor/DatabaseProcessor.kt
```diff
@@ -107,7 +107,11 @@ class DatabaseProcessor(baseContext: Context, val element: XTypeElement) {
                 views = views,
                 daoMethods = daoMethods,
                 exportSchema = dbAnnotation.value.exportSchema,
-                enableForeignKeys = hasForeignKeys)
+                enableForeignKeys = hasForeignKeys,
+                typeConverterFactories = HashSet(context.typeConverterFactories))
+
+        context.typeConverterFactories.clear()
```
danysantiago: This mutable map is still a bit scary, what we can do is move the set to a property in the `DatabaseProcessor` and then let the `Context` have a function that `CustomComnverterProcessor` can call into that is provided by `DatabaseProcessor` that adds to the set.

In other words:
```
class Context (
  ...
  val addTypeConverterFactory: (XType) -> Unit
) {
  ...
  fun fork(..., addTypeConverterFactory: (XType) -> Unit = {  }) {
    ...
  }
}
```
then:
```
class DatabaseProcessor(...) {
    val context = baseContext.fork(
        element = element,
        addTypeConverterFactory = { typeConverterFactories.add(it) }
    )
    private val typeConverterFactories = mutableSetOf<XType>()
}
```

danysantiago (changes requested):
This is looking great, just a few more comments, thanks!

room/runtime/src/main/java/androidx/room/RoomDatabase.java
mzgreen: Weird, I ran `./gradlew buildOnServer ktlint` before pushing. It doesn't detect this.

danysantiago: @dlam, does `./gradlew ktlint` perform formatting or is it just checking? I'm also not sure how strict is it with unconventional spaces...

mzgreen: I mean I ran `./gradlew buildOnServer ktlint` as suggested here: https://github.com/androidx/androidx/pull/73
and expected it to fail if there is something wrong. I fixed a lot of formatting issues using this approach before and I believe it detected some missing spaces too. Not sure why it doesn't work in this case. I've just ran it again and got no errors.

room/compiler/src/main/kotlin/androidx/room/processor/DatabaseProcessor.kt
mzgreen: Yeah it's much cleaner! Thanks 👍

room/runtime/src/main/java/androidx/room/RoomDatabase.java
mzgreen: Good point, I believe Yigit mentioned this as well. I can add a `final` check but I'm trying to understand why can't we support this. Wouldn't it be possible to do something like this:
1. Try to find the converter in the map and use it if it's there
2. Otherwise try to go over supertypes of a given factory and for every one of them check if it's in the map.
Not sure if it's worth it though.

room/runtime/src/main/java/androidx/room/RoomDatabase.java
dlam: ktlint just checks, ktformat applies the fix

dlam: we also normally run like three different kinds of lint as part of bOS, but I feel ktlint shouldve caught this.

room/compiler/src/main/kotlin/androidx/room/processor/DatabaseProcessor.kt
mzgreen: There is a slight problem though - it doesn't work 😅 
I think the problem is that `fork` calls `CustomConverterProcessor.findConverters(this, element)` and after that it creates a sub context. So `findConverters` is called before `addTypeConverterFactory` is assigned. I may be wrong though.
Trying to make it work now.

mzgreen: Maybe it could be done by passing `addTypeConverterFactory: (XType) -> Unit` as a third parameter to the `findConverters` function. Just like `element` param.

room/runtime/src/main/java/androidx/room/RoomDatabase.java
danysantiago: Right, we can't do this without going through superclasses at runtime via reflection. My take here is, it's not initially worth it, I would prefer if users compose their factories (and delegate) not via class hierarchy. We can always add support for this later (and remove the `final` restriction) if we get users with compelling reasons, but if we don't make it strict now and decided not to support this case in the long run we can't later add the restriction since it would cause ABI incompatibilities.

room/compiler/src/main/kotlin/androidx/room/processor/DatabaseProcessor.kt
danysantiago: Passing `addTypeConverterFactory` to `CustomConverterProcessor.findConverters` sounds totally reasonable, then we should make the function property `private` in `Context` to avoid misuse.

room/runtime/src/main/java/androidx/room/RoomDatabase.java
mzgreen: Makes sense, I'll add `final` check

room/compiler/src/main/kotlin/androidx/room/processor/DatabaseProcessor.kt
mzgreen: Yup! 👍

room/runtime/src/main/java/androidx/room/RoomDatabase.java
mzgreen: Do you think we should add an info about `final` requirement in the javadoc for `TypeConverterFactory`  interface?

danysantiago: Yes, definitely

room/compiler/src/main/kotlin/androidx/room/processor/CustomConverterProcessor.kt
```diff
@@ -41,14 +42,19 @@ class CustomConverterProcessor(val context: Context, val element: XTypeElement)
         private fun XType.isInvalidReturnType() =
             isError() || isVoid() || isNone()
 
-        fun findConverters(context: Context, element: XElement): ProcessResult {
+        fun findConverters(
+            context: Context,
+            element: XElement,
+            addTypeConverterFactory: (XType) -> Unit = {}
```
danysantiago: Lets not have a default value here for `addTypeConverterFactory` it makes sense in the `Context` class but not here since we want to make sure someone gets notified that a type converter factory has to be registered.

room/compiler/src/main/kotlin/androidx/room/processor/CustomConverterProcessor.kt
```diff
@@ -70,7 +76,7 @@ class CustomConverterProcessor(val context: Context, val element: XTypeElement)
         }
     }
 
-    fun process(): List<CustomTypeConverter> {
+    fun process(addTypeConverterFactory: (XType) -> Unit = {}): List<CustomTypeConverter> {
```
danysantiago: If we add a `private val` to `CustomConverterProcessor` in its constructor then we don't need this param since in `findConverters` can build a `CustomConverterProcessor` with the add lambda. This keeps `process()` clean as the processor has been built with everything it needs. It should also not have a default value.

danysantiago (changes requested):

danysantiago (approved):

room/runtime/src/main/java/androidx/room/DatabaseConfiguration.java
```diff
@@ -316,7 +320,64 @@ public DatabaseConfiguration(@NonNull Context context, @Nullable String name,
                 allowMainThreadQueries, journalMode, queryExecutor, transactionExecutor,
                 multiInstanceInvalidation, requireMigration, allowDestructiveMigrationOnDowngrade,
                 migrationNotRequiredFrom, copyFromAssetPath, copyFromFile, copyFromInputStream,
-                null);
+                null, null);
+    }
+
+     /**
```
yigit, line 326: unrealted to this CL but i'm not sure why we are still keeping older configurations.
DatabaseConfiguration is not created by generated code and even though constructor is public, it is `@RestrictTo(RestrictTo.Scope.LIBRARY_GROUP_PREFIX)` . is this because we are expected to keep api compatibilty even for restricted? 
cc: @danysantiago

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
```diff
@@ -0,0 +1,162 @@
...
+    @Test
+    public void testMissingTypeConverterFactoryInstance() {
+        Context context = ApplicationProvider.getApplicationContext();
+        try {
+            TestDatabaseWithConverter db =
+                    Room.inMemoryDatabaseBuilder(context, TestDatabaseWithConverter.class).build();
+            Pet pet = TestUtil.createPet(3);
+            pet.setName("pet");
+            db.getPetDao().insertOrReplace(pet);
+            fail("Show have thrown an IllegalArgumentException");
```
yigit: typo: "Should've"

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
```diff
@@ -0,0 +1,162 @@
...
+    public void testMissingTypeConverterFactoryAnnotation() {
+        Context context = ApplicationProvider.getApplicationContext();
+        try {
+            TestDatabase db = Room.inMemoryDatabaseBuilder(context, TestDatabase.class)
+                    .addTypeConverterFactory(new TimeStampConverterFactory())
+                    .build();
+            Pet pet = TestUtil.createPet(3);
+            pet.setName("pet");
+            db.getPetDao().insertOrReplace(pet);
+            fail("Show have thrown an IllegalArgumentException");
```
yigit: ditto

room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
```diff
@@ -37,7 +37,8 @@ class Context private constructor(
     private val typeConverters: CustomConverterProcessor.ProcessResult,
     private val inheritedAdapterStore: TypeAdapterStore?,
     val cache: Cache,
-    private val canRewriteQueriesToDropUnusedColumns: Boolean
+    private val canRewriteQueriesToDropUnusedColumns: Boolean,
+    private val addTypeConverterFactory: (XType) -> Unit
```
yigit: github doesn't let me comment on the previous discussion so i'll create a new one:
https://github.com/androidx/androidx/pull/75#discussion_r482967139

I'm still worried about this.
We keep type converters in Context because they are scoped to where they are declared and they are stateless.
The current implementation relies on a side effect which is a bit scary especially with caching where we try to avoid re-processing same classes (e.g. https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:room/compiler/src/main/kotlin/androidx/room/processor/TableEntityProcessor.kt;l=51)

I tried to create a repro for that but it didn't fail (not sure why, didn't try harder) but i'm afraid it might fail in similar cases.

In other words, there might be a failure for the following case:
There are two database, `Db1`,`Db2` that share same Entity `E1` where `E1` has a field that declares a type converter.

We start processing `Db1`, visit `E1` and add the factory.
We start processing `Db2`, visit `E1` again but it is cached hence return cached Entity information, missing on the factory.

I don't have a solution yet but the original implementation might be better than having a callback.
That is, have a mechanism for a `subContext` to report to its parent a type adapter.

Now, even that is not perfect though w/ the forks where sometimes we fork a context just to collect errors and sometime surpress them (https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:room/compiler/src/main/kotlin/androidx/room/processor/Context.kt;l=117?q=Context.kt)

Also, it would be great not to report errors on unused type converters but that is a good to have, not that important.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
```diff
@@ -0,0 +1,162 @@
+/*
+ * Copyright 2018 The Android Open Source Project
```
yigit: can we expand this test a little bit more to have cases where a type converter is declared in a lower layer, like a dao and entity.

for instance, while trying to understand why we need the add factory callback in Context, i removed it and all of these tests still pass since it only tests for factories declared in the database.

yigit: repro for that case:
https://android-review.googlesource.com/c/platform/frameworks/support/+/1419740

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
mzgreen: Oops copy paste mistake, thanks 😅

room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
mzgreen: I'm not super familiar with the codebase yet, not sure how exactly caching works but I think I know why it doesn't fail. One of the first things that `DatabaseProcessor` does is forking a baseContext. `fork()` method calls `CustomConverterProcessor.findConverters` before creating subcontext and returning it. So it happens before Daos and Entities are processed. That's why it works - Room executes `findConverters` logic without checking the cache.

I decided to report unused converters because I thought it may help developers to understand why the factory doesn't work even though they added it in the builder. But if you think it's better to not report it then I can remove this check.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
mzgreen: I'll look into that, thanks.

mzgreen: Yeah seems like keeping the add factory callback in Context may not be needed. It's only needed to be passed to `fork` method because this method calls `findConverters` logic.

I'll add more tests. But as I'm looking at this now I think you may be right and the tests you mentioned may fail. I need to think about collecting those factories a bit more.

room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
yigit: so the case above works because available converters is a cache key.

mzgreen: > so the case above works because available converters is a cache key.

I'm trying to understand this, can you elaborate? Does that mean that E1 will be visited again when processing Db2 because converters is a cache key?

mzgreen: One idea for gathering factories from lower levels (daos/entities etc) would be to modify fork method like this:
```
    fun fork(
        element: XElement,
        forceSuppressedWarnings: Set<Warning> = emptySet(),
        addTypeConverterFactory: ((XType) -> Unit)? = null // make it nullable
    ): Context {
        val suppressedWarnings = SuppressWarningProcessor.getSuppressedWarnings(element)
        val processConvertersResult =
            CustomConverterProcessor.findConverters(this, element, addTypeConverterFactory ?: this.addTypeConverterFactory) // pass argument if not null, otherwise use class property
        val canReUseAdapterStore = processConvertersResult.classes.isEmpty()
        // order here is important since the sub context should give priority to new converters.
        val subTypeConverters = if (canReUseAdapterStore) {
            this.typeConverters
        } else {
            processConvertersResult + this.typeConverters
        }
        val subSuppressedWarnings =
            forceSuppressedWarnings + suppressedWarnings + logger.suppressedWarnings
        val subCache = Cache(cache, subTypeConverters.classes, subSuppressedWarnings)
        val subCanRemoveUnusedColumns = canRewriteQueriesToDropUnusedColumns ||
                element.hasRemoveUnusedColumnsAnnotation()
        val subContext = Context(
                processingEnv = processingEnv,
                logger = RLog(logger.messager, subSuppressedWarnings, element),
                typeConverters = subTypeConverters,
                inheritedAdapterStore = if (canReUseAdapterStore) typeAdapterStore else null,
                cache = subCache,
                canRewriteQueriesToDropUnusedColumns = subCanRemoveUnusedColumns,
                addTypeConverterFactory = addTypeConverterFactory ?: this.addTypeConverterFactory) // same here, use arg if not null, otherwise pass class property
        subContext.databaseVerifier = databaseVerifier
        return subContext
    }
```
That way once a not null callback is passed to fork, it'll be used starting from that level and passed to lower levels. We're creating a proper callbacks that collects factories into a map inside `DatabaseProcessor` so all subcontexts would get this callback.


Still not sure if this covers the caching case with 2 DB sharing same entity.

yigit: OK here is a proposal.
tl;dr; Delegate verification to static methods in generated DAOs 🙈 

This provides the best behavior (we won't validate unless it is used) and does not require any change in the `Context`.

Right now, `CustomTypeConverterWrapper.typeConverterFactory` gets a field from the scope.
We can add a scope function specific to factories which can track all factories that are used.

`DaoWriter`, at the end, can generate a `static` method that returns the list of those factories.

`DatabaseWriter` can just query these static methods to build the list and verify against it. That even allows us to print a better runtime exception where we can tell the user "XXDAo needs this type converter factory".

wdyt?

yigit: > > so the case above works because available converters is a cache key.
> 
> I'm trying to understand this, can you elaborate? Does that mean that E1 will be visited again when processing Db2 because converters is a cache key?

So E1 will be visited again for DB2 and then skipped because we've visited E1.
Right now it doesn't because our caching is not clever enough to understand that we've visited E1 before with same type converters hence it works fine.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
yigit: i have a prototype for the DAO based suggestion, will upload it soon.

yigit: here is the *dirty* prototype for removing it from context fully.
https://android-review.googlesource.com/c/platform/frameworks/support/+/1420426

Basically what i do is:
* add a function to codegen scope that is special for converter factories (ugly)
* in dao, generate a method that will return all of the factory `Class`es (not strings)
* in db, generate a method that will return a `Map<String, List<Class<? : TypeConverterFactory>>` that is a map of DAO name to the list of factories it requires.
* in database constructor, get that list and verify against provided items.

I also made another change to not use the cannonical names as keys, instead, just use assignability.
I'm not sure why we preferred strings so this change might be wrong but i thought it would be handy for cases where a sub-type of the factory will be provided.

Lmk what you think. This is just a prototype, quite a dirty implementation but passes tests.

yigit: cc @danysantiago

mzgreen: Seems like there is not much left to be done here! 😛 Nice work

yigit: it needs some cleanup but yea i'm not unhappy about it. I'm not sure if it is better though, lets see what @danysantiago thinks and then if we'll go with that, i can send a PR to your fork.

You can probably already apply it to here via:
```
git fetch https://android.googlesource.com/platform/frameworks/support refs/changes/26/1420426/2 && git cherry-pick FETCH_HEAD
```
Not sure if that command will require authentication, if not, you just need to run it in your local checkout in theory.

Though I should first clean it up.

room/runtime/src/main/java/androidx/room/DatabaseConfiguration.java
danysantiago: Yes - It possible the API tools forced me to start doing this some time ago and know they are able to track restricted APIs and more importantly allow us to be incompatible knowing they are restricted but I have not tried. =/

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/TypeConverterFactoryTest.java
danysantiago: These are valid points, I like some of Yigit's approach, the DAO informing the factory it expects and then the Database gathering those for validation. However, I'm not convinced about using the `ClassWritter` and adding type converter factory specific code. I talked with @yigit offline (as in, online via a video call) and he is going to cleanup the usage of `ClassWitter` a bit more by having a more general solution for metadata, which in this case would be factories used. Ultimately, @mzgreen, you should able to cherry-pick his patch, adjust it if necessary and then get this merged.

yigit: here is that change:
https://android-review.googlesource.com/c/platform/frameworks/support/+/1420426/2..4

lmk what you think and i'll clean it up for @mzgreen to cherry-pick.

yigit: ok that implementation is ready for review as well:
https://android-review.googlesource.com/c/platform/frameworks/support/+/1420426

@danysantiago @mzgreen lmk what you prefer on whether we want to review that separately since this one is already reviewed.

The changes in that CL are:
* use class as key, allow subclasses
* collect factories in DAOs
* remove get converter factory map from RoomDatabase but instead provide a getter. This will allow not exposing the map so that it is easier to change in the future if we need to.

danysantiago:
Looks like we are still missing the changes to `MigrationTestHelper` to use the newest non-deprecated constructor from `Configuration`.

room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
```diff
@@ -127,9 +127,13 @@ class Context private constructor(
         return Pair(result, collector)
     }
 
-    fun fork(element: XElement, forceSuppressedWarnings: Set<Warning> = emptySet()): Context {
+    fun fork(
```
danysantiago: nit: After all the changes looks like this file basically stayed the same and only have some formatting changes, would be good to fully revert it to how it was originally to avoid unnecessary changes in the PR.

room/compiler/src/main/kotlin/androidx/room/processor/DatabaseProcessor.kt
```diff
@@ -37,7 +37,9 @@ import com.squareup.javapoet.TypeName
 import java.util.Locale
 
 class DatabaseProcessor(baseContext: Context, val element: XTypeElement) {
-    val context = baseContext.fork(element)
+    val context = baseContext.fork(
```
danysantiago: nit: With the new latest update this file only has a formatting change, would be good to revert it back.

room/compiler/src/test/data/databasewriter/output/ComplexDatabase.java
```diff
@@ -174,6 +177,14 @@ public void clearAllTables() {
         }
     }
 
+    @Override
+    protected Map<String, List<Class<?>>> getRequiredTypeConverters() {
+        final HashMap<String, List<Class<?>>> _typeConvertersMap =
+                new HashMap<String, List<Class<?>>>();
+        _typeConvertersMap.put("foo.bar.ComplexDao", ComplexDao_Impl.getRequiredConverters());
```
danysantiago: Instead of using DAO class name in Strings we should probably use the Class itself to respect obfuscation.

danysantiago (changes requested):
Thanks for the latest update, I think this is way better.

room/compiler/src/main/kotlin/androidx/room/processor/Context.kt
mzgreen: Done

room/compiler/src/main/kotlin/androidx/room/processor/DatabaseProcessor.kt
mzgreen: Done

room/compiler/src/test/data/databasewriter/output/ComplexDatabase.java
mzgreen: Right! Changed :)

danysantiago (approved):

room/common/src/main/java/androidx/room/ProvidedTypeConverter.java
```diff
@@ -16,21 +16,15 @@
...
-import androidx.annotation.NonNull;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
 
 /**
- * Implementations of {@code TypeConverterFactory} interface are responsible to instantiate
- * TypeConverters.
+ * Marks a class as a type converter that will be provided to Room at runtime.
```
yigit, line 25:
```suggestion
 * Marks a class as a type converter that will be provided to Room at runtime.
 * If Room uses the annotated type converter class, it will verify that it is provided in the builder and if not, will throw
 * an exception.
```

yigit (approved):

room/common/src/main/java/androidx/room/ProvidedTypeConverter.java
yigit: i think this got lost.

PR #77, opened 2020-09-12
Partial implementation for XProcessing with KSP

yigit:
XType implementation with KSP

This PR adds the first bits of support for KSP in Room. It is still far from complete but moving development back to aosp since both AndroidX and KSP moved to Kotlin 1.4.0

Bug: 160322705
Test: KspTypeTest,  KspTypeExtTest


yigit:
@danysantiago this is combination of CLs you've already reviewed in my fork.
Once it is mirrored into AOSP, i'll add prebuilts.

yigit:
try again github

danysantiago (approved):

buildSrc/src/main/kotlin/androidx/build/AndroidXPlaygroundRootPlugin.kt
```diff
@@ -157,6 +157,10 @@ class AndroidXPlaygroundRootPlugin : Plugin<Project> {
         google()
         mavenCentral()
         jcenter()
+        // TODO temporary until AndroidX merge
+        maven {
+            it.url = URI("https://jitpack.io")
```
danysantiago: Probably need to remove this?

yigit: good catch.
I also need to move away from its snapshot to an actual release but that is breaking the build right now, will take a look but may need to wait for another release from kotlin compile testing.

PR #87, opened 2020-10-14
Don't cache fragment classes across different class loaders

simonschiller:
## Proposed Changes

  - Don't cache fragment classes across different class loaders inside `FragmentFactory`.

## Testing

Test: Added tests to verify the change

## Issues Fixed

Fixes: 113886460

### DIFF
--- a/fragment/fragment/src/test/java/androidx/fragment/app/FragmentFactoryTest.kt
+++ b/fragment/fragment/src/test/java/androidx/fragment/app/FragmentFactoryTest.kt
@@ -16,7 +16,7 @@
 
 package androidx.fragment.app
 
-import org.junit.Assert.assertEquals
+import com.google.common.truth.Truth.assertWithMessage
 import org.junit.Test
 
 class FragmentFactoryTest {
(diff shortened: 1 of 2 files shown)


fragment/fragment/src/test/java/androidx/fragment/app/FragmentFactoryTest.kt
```diff
@@ -0,0 +1,53 @@
...
+
+class FragmentFactoryTest {
+    private val factory = FragmentFactory()
+
+    @Test
+    fun `Fragment classes are cached`() {
+        val classLoader = CountingClassLoader()
+        factory.instantiate(classLoader, TestFragment::class.java.name)
+        factory.instantiate(classLoader, TestFragment::class.java.name)
+        assertEquals(1, classLoader.loadCount)
```
ianhanniballake: Please use [Truth](https://truth.dev/) asserts (note, we already have a test dependency on Truth) over Junit asserts here and elsewhere in this file:
```
assertThat(classLoader.loadCount).isEqualTo(1)
```
Or preferably use the `assertWithMessage()` version that includes a more detailed message that is printed when the assert fails.

ianhanniballake (changes requested):

ianhanniballake (approved):

ianhanniballake:
Thanks for putting in this PR! It'll be available in the next Fragment release.

PR #89, opened 2020-10-14
Add extension function to retrieve a Flow of NavBackStackEntries

simonschiller:
## Proposed Changes

  - Added an extension function to retrieve the current `NavBackStackEntry` as a `Flow`.

## Testing

Test: Added a test for the new function

## Issues Fixed

Fixes: 163947280

### DIFF
--- a/navigation/navigation-runtime-ktx/src/androidTest/java/androidx/navigation/NavControllerTest.kt
+++ b/navigation/navigation-runtime-ktx/src/androidTest/java/androidx/navigation/NavControllerTest.kt
@@ -50,14 +50,14 @@ class NavControllerTest {
 
     @Test
     @Suppress("EXPERIMENTAL_API_USAGE")
-    fun backStackEntryFlow() = runBlocking {
+    fun currentBackStackEntryFlow() = runBlocking {
         navController.graph = navController.createGraph(startDestination = 1) {
             test(1)
             test(2)
             test(3)
         }
 
-        navController.backStackEntryFlow
+        navController.currentBackStackEntryFlow
             .take(navController.graph.count())
             .withIndex()
             .onEach { (index, backStackEntry) ->
--- a/navigation/navigation-runtime-ktx/src/androidTest/java/androidx/navigation/NavControllerTest.kt
+++ b/navigation/navigation-runtime-ktx/src/androidTest/java/androidx/navigation/NavControllerTest.kt
@@ -21,13 +21,12 @@ import androidx.test.core.app.ApplicationProvider
 import androidx.test.filters.SmallTest
 import androidx.testutils.TestNavigator
 import androidx.testutils.test
-import kotlinx.coroutines.ExperimentalCoroutinesApi
+import com.google.common.truth.Truth.assertWithMessage
 import kotlinx.coroutines.flow.collect
 import kotlinx.coroutines.flow.onEach
 import kotlinx.coroutines.flow.take
 import kotlinx.coroutines.flow.withIndex
 import kotlinx.coroutines.runBlocking
-import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
 import org.junit.Test
 
@@ -63,7 +62,9 @@ class NavControllerTest {
             .withIndex()
             .onEach { (index, backStackEntry) ->
                 val expectedDestination = index + 1
-                assertEquals(expectedDestination, backStackEntry.destination.id)
+                assertWithMessage("Flow emitted unexpected back stack entry (wrong destination)")
+                    .that(backStackEntry.destination.id)
+                    .isEqualTo(expectedDestination)
 
                 if (expectedDestination < navController.graph.count()) {
                     navController.navigate(expectedDestination + 1)
--- a/navigation/navigation-runtime-ktx/src/androidTest/java/androidx/navigation/NavControllerTest.kt
+++ b/navigation/navigation-runtime-ktx/src/androidTest/java/androidx/navigation/NavControllerTest.kt
@@ -21,6 +21,13 @@ import androidx.test.core.app.ApplicationProvider
 import androidx.test.filters.SmallTest
 import androidx.testutils.TestNavigator
 import androidx.testutils.test
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.flow.collect
+import kotlinx.coroutines.flow.onEach
+import kotlinx.coroutines.flow.take
+import kotlinx.coroutines.flow.withIndex
+import kotlinx.coroutines.runBlocking
+import org.junit.Assert.assertEquals
 import org.junit.Assert.assertTrue
 import org.junit.Test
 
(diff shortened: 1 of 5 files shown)


navigation/navigation-runtime-ktx/src/androidTest/java/androidx/navigation/NavControllerTest.kt
```diff
@@ -41,6 +48,29 @@ class NavControllerTest {
...
+            test(2)
+            test(3)
+        }
+
+        navController.getBackStackEntryFlow()
+            .take(navController.graph.count())
+            .withIndex()
+            .onEach { (index, backStackEntry) ->
+                val expectedDestination = index + 1
+                assertEquals(expectedDestination, backStackEntry.destination.id)
```
ianhanniballake: Please use [Truth](https://truth.dev/) asserts (note, we already have an androidTest dependency on Truth) over Junit asserts here:
```
assertThat(backStackEntry.destination.idt).isEqualTo(expectedDestination)
```
Or preferably use the `assertWithMessage()` version that includes a more detailed message that is printed when the assert fails.

navigation/navigation-runtime-ktx/src/main/java/androidx/navigation/NavController.kt
```diff
@@ -26,3 +31,19 @@ public inline fun NavController.createGraph(
...
 ): NavGraph = navigatorProvider.navigation(id, startDestination, builder)
+
+/**
+ * Creates and returns a [Flow] that will emit the currently active [NavBackStackEntry] whenever
+ * it changes. If there is no active [NavBackStackEntry], no item will be emitted.
+ */
+@Suppress("EXPERIMENTAL_API_USAGE")
+public fun NavController.getBackStackEntryFlow(): Flow<NavBackStackEntry> = callbackFlow {
+    val listener = NavController.OnDestinationChangedListener { controller, _, _ ->
+        controller.currentBackStackEntry?.let(::sendBlocking)
```
svasilinets, line 42: would it be better to use second argument instead?  given that it is non null it will make things as easy as `sendBlocking(destination)`

another corner case question: sendBlocking can theoretically block on the main thread, when there is slow consumer and a lot of actions here. Though it seems like fairly theoretical concern

navigation/navigation-runtime-ktx/src/main/java/androidx/navigation/NavController.kt
```diff
@@ -26,3 +31,19 @@ public inline fun NavController.createGraph(
     @IdRes startDestination: Int,
     builder: NavGraphBuilder.() -> Unit
 ): NavGraph = navigatorProvider.navigation(id, startDestination, builder)
+
+/**
+ * Creates and returns a [Flow] that will emit the currently active [NavBackStackEntry] whenever
+ * it changes. If there is no active [NavBackStackEntry], no item will be emitted.
+ */
+@Suppress("EXPERIMENTAL_API_USAGE")
+public fun NavController.getBackStackEntryFlow(): Flow<NavBackStackEntry> = callbackFlow {
```
svasilinets: I'd probably suggest to shape it as extension property:
`public val NavController.backStackEntryFlow : .... `

navigation/navigation-runtime-ktx/src/main/java/androidx/navigation/NavController.kt
```diff
@@ -26,3 +31,19 @@ public inline fun NavController.createGraph(
     @IdRes startDestination: Int,
     builder: NavGraphBuilder.() -> Unit
 ): NavGraph = navigatorProvider.navigation(id, startDestination, builder)
+
+/**
+ * Creates and returns a [Flow] that will emit the currently active [NavBackStackEntry] whenever
+ * it changes. If there is no active [NavBackStackEntry], no item will be emitted.
+ */
+@Suppress("EXPERIMENTAL_API_USAGE")
```
svasilinets: I think it should be marked as experimental api. We won't be able to release it as stable until callbackFlow will be stable (androidx gives binary compatibility guarantees that could be broken by callbackFlow)

navigation/navigation-runtime-ktx/src/main/java/androidx/navigation/NavController.kt
simonschiller: Would be possible, but then it would be a `Flow<NavDestination>` instead of a `Flow<NavBackStackEntry>`, which is not what we want I guess (as `NavBackStackEntry` is the class that gives access to things like `ViewModelStore` and the `SavedStateRegistry`). But maybe it would make sense to have both, a `backStackEntryFlow` and a `destinationFlow`?

Valid point, I also wasn't sure what's best to use here. The other option would be to use `offer`, which could lead to lost items it the consumer is slow.

navigation/navigation-runtime-ktx/src/main/java/androidx/navigation/NavController.kt
simonschiller: Just to make sure I understand correctly - since it is part of the public API `@ExperimentalCoroutinesApi` is required. If it were an internal function, `@Suppress` would be fine (as it does not cause binary incompatible changes), correct?

navigation/navigation-runtime-ktx/src/main/java/androidx/navigation/NavController.kt
```diff
@@ -36,8 +36,8 @@ public inline fun NavController.createGraph(
  * Creates and returns a [Flow] that will emit the currently active [NavBackStackEntry] whenever
  * it changes. If there is no active [NavBackStackEntry], no item will be emitted.
  */
-@Suppress("EXPERIMENTAL_API_USAGE")
-public fun NavController.getBackStackEntryFlow(): Flow<NavBackStackEntry> = callbackFlow {
+@ExperimentalCoroutinesApi
+public val NavController.backStackEntryFlow: Flow<NavBackStackEntry> get() = callbackFlow {
```
jbw0033: Could we change this to `NavController.currentBackStackEntryFlow` to be consistent with `NavController.currentBackStackEntry` API?

navigation/navigation-runtime-ktx/src/main/java/androidx/navigation/NavController.kt
jbw0033: We are specifically looking for the `NavBackStackEntry` here, not `NavDestination`.

jbw0033 (changes requested):

navigation/navigation-runtime-ktx/src/main/java/androidx/navigation/NavController.kt
simonschiller: Good point, done 👍

simonschiller:
CI is red, but looks unrelated - was this a build flake? Can I retrigger somehow?

jbw0033 (approved):

ianhanniballake (approved):

dlam:
Hey the build was just fixed! Sorry it took a few days to get it landed - could you try rebasing your change on top of androidx-master-dev and see if the tests pass now?

simonschiller:
Rebasing fixed the issue :+1:

ianhanniballake:
Thanks again for implementing this! It'll become available in Navigation 2.4.0-alpha01.

PR #90, opened 2020-10-16
Exclude compose navigation from navigation playground

yigit:
Compose projects are not supported in playground yet so this
PR excludes compose-navigation and its sub-projects from
the navigation playground setup.

Bug: 171017446
Test: cd navigation && ./gradlew buildOnServer


tikurahul (approved):

yigit:
i don't get why this is failing, do we have something that ignores workflow file changes?

dlam:
I *think* pull request target is against base, which is how it's able to do the security token, but something worth looking into.

dlam (approved):

yigit:
actually i just changed a settings file, not workflow file 🤷

PR #93, opened 2020-10-26
Parallelize builds across library groups

dlam:
This change parallelizes the primary bOS task across library groups by separating them into different jobs.

To re-use more of the code, https://github.com/androidx/build-on-server-action has been setup, but there currently isn't a good way to re-use actions, so the cache step needs to be duplicated. Of note is that the JDK install step has been replaced so that we can guarantee version + install location (which also now fixes testing on linux).

Follow-up for this will be to migrate ktlint + other preupload checks and to cache the jdk install. However, we're granted a 5gb actions/cache limit and I believe gradle cache will easily meet this as we scale up.

Note: This change also switches to the official androidx mirror of our custom actions.

Test: Github workflow .presubmit.yml
Fixes: 163145881

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -6,91 +6,406 @@ on:
   workflow_dispatch:
 
 jobs:
-  # This workflow contains a single job called "build"
-  build:
+  setup:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Start webhook"
+        uses: androidx/github-workflow-webhook-action@main
+        with:
+          url: 'https://androidx.dev/github/androidX/presubmit/hook'
+          secret: ${{ secrets.ANDROIDX_PRESUBMIT_HOOK_SECRET }}
+          payload: '{ "platform": "all", "token": "${{ secrets.GITHUB_TOKEN }}", "state": "started"}'
+
+  lint:
+    runs-on: ubuntu-latest
+    outputs:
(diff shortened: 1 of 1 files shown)


.github/workflows/presubmit.yml
```diff
@@ -6,91 +6,406 @@ on:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Start webhook"
+        uses: androidx/github-workflow-webhook-action@main
+        with:
+          url: 'https://androidx.dev/github/androidX/presubmit/hook'
+          secret: ${{ secrets.ANDROIDX_PRESUBMIT_HOOK_SECRET }}
+          payload: '{ "platform": "all", "token": "${{ secrets.GITHUB_TOKEN }}", "state": "started"}'
+
+  lint:
```
dlam, line 19: This will include `PREUPLOAD.cfg` in a follow-up

dlam:
Even though we don't cache JDK yet this brings presubmit from 50m -> 20m: https://github.com/dlam/androidx/actions

dlam:
There is currently no way to re-use the cache action, but we have a 5gb limit, so I'll need to play with removing the gradle cache since I think it's likely we are blowing out the cache every build.

tikurahul (approved):
This is awesome  :clap:

tikurahul:
As a follow up we could switch to the Gradle cache action itself, but does not have to be a part of this change.

PR #94, opened 2020-10-27
Delegate to beginTransactionNonExclusive when in WAL mode

R4md4c:
## Proposed Changes

As discussed in the ticket, I'm just delegating to `beginTransactionNonExclusive` when WAL mode is enabled.

## Testing

Test:
Executed
```
./gradlew \
    test connectedCheck \
    -x :room:room-benchmark:cC \
```
and nothing was broken, please inform me if you'd like to add a specific test case.

## Issues Fixed
Fixes: 126258791

### DIFF
--- a/room/runtime/src/main/java/androidx/room/InvalidationTracker.java
+++ b/room/runtime/src/main/java/androidx/room/InvalidationTracker.java
@@ -309,7 +309,7 @@ public class InvalidationTracker {
         return tables.toArray(new String[tables.size()]);
     }
 
-    private void beginTransactionInternal(SupportSQLiteDatabase database) {
+    private static void beginTransactionInternal(SupportSQLiteDatabase database) {
         if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.JELLY_BEAN
                 && database.isWriteAheadLoggingEnabled()) {
             database.beginTransactionNonExclusive();
--- a/room/runtime/src/main/java/androidx/room/InvalidationTracker.java
+++ b/room/runtime/src/main/java/androidx/room/InvalidationTracker.java
@@ -20,6 +20,7 @@ import android.annotation.SuppressLint;
 import android.content.Context;
 import android.database.Cursor;
 import android.database.sqlite.SQLiteException;
+import android.os.Build;
 import android.util.Log;
 
 import androidx.annotation.NonNull;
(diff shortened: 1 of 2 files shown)


R4md4c:
One open question though, do I apply the same change to [`InvalidationTracker`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-master-dev:room/runtime/src/main/java/androidx/room/InvalidationTracker.java;l=63?q=InvalidationTracker&sq=&ss=androidx)? as it still uses `beginTransaction` internally.

danysantiago:
This looks good, thanks for making this PR.

Yes, we should also updated the `InvalidationTracker` to conditionally use the right `beginTransaction` call/

danysantiago (approved):

room/runtime/src/main/java/androidx/room/InvalidationTracker.java
```diff
@@ -389,7 +399,7 @@ public void run() {
                     // This transaction has to be on the underlying DB rather than the RoomDatabase
                     // in order to avoid a recursive loop after endTransaction.
                     SupportSQLiteDatabase db = mDatabase.getOpenHelper().getWritableDatabase();
-                    db.beginTransaction();
+                    db.beginTransactionNonExclusive();
```
danysantiago, line 402: This line should call the new internal method, `beginTransactionInternal`

R4md4c: This line is already inside `if (mDatabase.mWriteAheadLoggingEnabled)`, so I thought it's better to use the native method?

danysantiago: Oh right, I missed that! Thanks!

MaxiPoison:
Зачем ты мне прислал это сообщение лучше просто денег скинь и всё

ср, 28 окт. 2020 г., 19:07 Daniel Santiago <notifications@github.com>:

dlam:
Hey - the build was just fixed! Could you try rebasing your change on top of androidx-master-dev and see if the tests pass now?

R4md4c:
@dlam Hey, thanks to your fix, the build is green now.

ZhaoxiZhang:
I know `beginTransactionNonExclusive` method will executes `BEGIN IMMEDIATE;` and `beginTransaction` executes `BEGIN EXCLUSIVE;`，they are different transcation. But in [SQLite Document WebSite](https://www.sqlite.org/lang_transaction.html) descrips "EXCLUSIVE and IMMEDIATE are the same in [WAL mode](https://www.sqlite.org/wal.html)". I don't think there is a need for a special judgment against WAL mode to execute `beginTransactionNonExclusive`. Maybe you have a nice explanation for this modify?
image

PR #98, opened 2020-11-17
A PagingSource Factory that can be invalidated

claraf3:
## Proposed Changes

A PagingSource Factory that stores references to PagingSources for PagingSource invalidation

## Testing

### DIFF
--- a/paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
+++ b/paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
@@ -51,7 +51,8 @@ abstract class InvalidatingPagingSourceFactory<Key : Any, Value : Any>(
         while (pagingSources.isNotEmpty()) {
             pagingSources.removeFirst().also {
                 if (!it.invalid) {
-                    it.invalidate() }
+                    it.invalidate()
+                }
             }
         }
     }
--- a/paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
+++ b/paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2020 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.paging
+
+import androidx.annotation.VisibleForTesting
+
+/**
+ * A PagingSource Factory that can be kept as a reference and stores a list of paging sources
+ * created with the factory. Includes invalidate() function to invalidate the list of [PagingSource]
+ * when data changes.
+ */
+abstract class InvalidatingPagingSourceFactory<Key: Any, Value: Any> : () ->  PagingSource<Key, Value>{
+
+    @VisibleForTesting
+    internal val listOfPagingSources = mutableListOf<PagingSource<Key, Value>>();
+
+    /**
+     * Returns a PagingSource
+     *
+     * Implement the logic to create a PagingSource
+     */
+    abstract fun create() : PagingSource<Key, Value>;
+
+    /**
+     * Returns a PagingSource which will also be stored into listOfPagingSources.
+     */
+    final override fun invoke(): PagingSource<Key, Value> {
+        return create(). also {listOfPagingSources.add(it)};
+    }
+
+    /**
+     * Calls [PagingSource] invalidate() on each PagingSource stored within listOfPagingSources
+     * and removes it from the list
+     *
+     * Invalidated [PagingSource] will have its invalid status equal 'true'
+     */
+    fun invalidate() {
+        while (listOfPagingSources.isNotEmpty()) {
+            listOfPagingSources.removeFirst().invalidate();
+        }
+    }
+}
--- a/paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
+++ b/paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
@@ -36,19 +36,22 @@ abstract class InvalidatingPagingSourceFactory<Key : Any, Value : Any>(
     internal val pagingSources = mutableListOf<PagingSource<Key, Value>>()
 
     /**
-     * @return [PagingSource] which will also be stored into [pagingSources].
+     * @return [PagingSource] which will be invalidated when this factory's [invalidate] method
+     * is called
      */
     final override fun invoke(): PagingSource<Key, Value> {
         return pagingSourceFactory().also { pagingSources.add(it) }
     }
 
     /**
-     * Calls [PagingSource.invalidate] on each [PagingSource] stored within [pagingSources]
-     * and removes it from [pagingSources]
+     * Calls [PagingSource.invalidate] on each [PagingSource] that was produced by this
+     * [InvalidatingPagingSourceFactory]
      */
     fun invalidate() {
         while (pagingSources.isNotEmpty()) {
-            pagingSources.removeFirst().invalidate()
+            pagingSources.removeFirst().also {
+                if (!it.invalid) it.invalidate()
+            }
         }
     }
 }
--- a/paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
+++ b/paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
@@ -29,7 +29,7 @@ import androidx.annotation.VisibleForTesting
  * @param pagingSourceFactory The [PagingSource] factory that returns a PagingSource when called
  */
 abstract class InvalidatingPagingSourceFactory<Key : Any, Value : Any>(
-    val pagingSourceFactory: () -> PagingSource<Key, Value>
+    private val pagingSourceFactory: () -> PagingSource<Key, Value>
 ) : () -> PagingSource<Key, Value> {
 
     @VisibleForTesting
--- a/paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
+++ b/paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
@@ -50,7 +50,8 @@ abstract class InvalidatingPagingSourceFactory<Key : Any, Value : Any>(
     fun invalidate() {
         while (pagingSources.isNotEmpty()) {
             pagingSources.removeFirst().also {
-                if (!it.invalid) it.invalidate()
+                if (!it.invalid) {
+                    it.invalidate() }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+     *
+     * Implement the logic to create a PagingSource
+     */
+    abstract fun create() : PagingSource<Key, Value>;
+
+    /**
+     * Returns a PagingSource which will also be stored into listOfPagingSources.
+     */
+    final override fun invoke(): PagingSource<Key, Value> {
+        return create(). also {listOfPagingSources.add(it)};
```
dlam:
```suggestion
        return create().also {
            listOfPagingSources.add(it)
        };
```

dlam: We have a lint task that you can run `./gradlew ktlint` to find errors and a way to automatically fix them using `./gradlew ktlintFormat`

paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+package androidx.paging
+
+import androidx.annotation.VisibleForTesting
+
+/**
+ * A PagingSource Factory that can be kept as a reference and stores a list of paging sources
+ * created with the factory. Includes invalidate() function to invalidate the list of [PagingSource]
+ * when data changes.
```
dlam:
```suggestion
 * Abstract [PagingSource] factory intended for usage in [Pager] construction, that tracks [PagingSource]s
 * it produces for invalidation when the backing dataset is updated.
 *
 * Calling [invalidate] on this [InvalidatingPagingSourceFactory] will automatically forward 
 * invalidate signals to all active [PagingSource]s.
```

dlam: Using [ClassName] or [function] will let KDocs link to the class, which is super helpful as it shows up as a clickable link in the generated docs.

paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+
+    @VisibleForTesting
+    internal val listOfPagingSources = mutableListOf<PagingSource<Key, Value>>();
+
+    /**
+     * Returns a PagingSource
+     *
+     * Implement the logic to create a PagingSource
+     */
+    abstract fun create() : PagingSource<Key, Value>;
```
dlam: This should at least be `internal` to prevent confusion for those using it incorrectly like in

```
Pager(...) {
    MyInvalidatingPagingSourceFactory.create()
}
```
but most likely even if it's internal people will still be able to do the incorrect thing since their usage would likely be in the same package.

A better approach might be to take in the original factory using a property in constructor like so:

abstract class InvalidatingPagingSourceFactory<Key: Any, Value: Any>(
    val pagingSourceFactory: () -> PagingSource<Key, Value>
) {
    ...
    final override fun invoke(): ... {
        return pagingSourceFactory().also { ... }
    }
}
```

File: paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt [hunk 4]
```diff
@@ -0,0 +1,56 @@
...
+
+    /**
+     * Returns a PagingSource which will also be stored into listOfPagingSources.
```
dlam: You can use `@returns` to specifically mark docs for the return value

paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+     */
+    final override fun invoke(): PagingSource<Key, Value> {
+        return create(). also {listOfPagingSources.add(it)};
+    }
+
+    /**
+     * Calls [PagingSource] invalidate() on each PagingSource stored within listOfPagingSources
```
dlam: You should be able to directly link [PagingSource.invalidate] here.

dlam:
```suggestion
     * Calls [PagingSource.invalidate] on each PagingSource stored within a list of [PagingSource]s
```

paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+     * and removes it from the list
+     *
+     * Invalidated [PagingSource] will have its invalid status equal 'true'
```
dlam:
```suggestion
```

This is more of a side-effect of invalidation and not the core purpose (updating backing dataset by signaling to Pager it needs a new PagingSource). I think it would be better to just remove this bit to focus the docs on what this function does

paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * A PagingSource Factory that can be kept as a reference and stores a list of paging sources
+ * created with the factory. Includes invalidate() function to invalidate the list of [PagingSource]
+ * when data changes.
+ */
+abstract class InvalidatingPagingSourceFactory<Key: Any, Value: Any> : () ->  PagingSource<Key, Value>{
+
+    @VisibleForTesting
+    internal val listOfPagingSources = mutableListOf<PagingSource<Key, Value>>();
```
dlam: nit: just use `pagingSources` as it's a bit shorter but is still obviously plural.

paging/common/src/test/kotlin/androidx/paging/InvalidatingPagingSourceFactoryTest.kt
```diff
@@ -0,0 +1,81 @@
...
+class InvalidatingPagingSourceFactoryTest {
+
+    @Test
+    fun getPagingSource() {
+
+        val testFactory = object : InvalidatingPagingSourceFactory<Int, Int>() {
+            override fun create(): PagingSource<Int, Int> = TestPagingSource();
+        }
+
+        testFactory.invoke()
```
dlam: suggest: Loop instead of copying this line?

paging/common/src/test/kotlin/androidx/paging/InvalidatingPagingSourceFactoryTest.kt
```diff
@@ -0,0 +1,81 @@
...
+
+
+class InvalidatingPagingSourceFactoryTest {
+
+    @Test
+    fun getPagingSource() {
+
+        val testFactory = object : InvalidatingPagingSourceFactory<Int, Int>() {
+            override fun create(): PagingSource<Int, Int> = TestPagingSource();
+        }
```
dlam: Instead of copying these lines into every test, pull it out into a class-level property or make a helper function / class

paging/common/src/test/kotlin/androidx/paging/InvalidatingPagingSourceFactoryTest.kt
```diff
@@ -0,0 +1,76 @@
...
+import kotlin.test.assertEquals
+
+class InvalidatingPagingSourceFactoryTest {
+
+    @Test
+    fun getPagingSource() {
+
+        val testFactory = getFactory()
+
+        for (i in 0..3) {
```
dlam: Bit more idiomatic is `repeat`: https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/repeat.html

Sorry should have mentioned in the first pass.

paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
```diff
@@ -0,0 +1,54 @@
...
+        return pagingSourceFactory().also { pagingSources.add(it) }
+    }
+
+    /**
+     * Calls [PagingSource.invalidate] on each [PagingSource] stored within [pagingSources]
+     * and removes it from [pagingSources]
+     */
+    fun invalidate() {
+        while (pagingSources.isNotEmpty()) {
+            pagingSources.removeFirst().invalidate()
```
dlam: Can we also add a check here if `pagingSource` is already invalid, then let's skip the `.invalidate()` call and ideally add a test for this as well.

paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
```diff
@@ -0,0 +1,54 @@
...
+ */
+abstract class InvalidatingPagingSourceFactory<Key : Any, Value : Any>(
+    val pagingSourceFactory: () -> PagingSource<Key, Value>
+) : () -> PagingSource<Key, Value> {
+
+    @VisibleForTesting
+    internal val pagingSources = mutableListOf<PagingSource<Key, Value>>()
+
+    /**
+     * @return [PagingSource] which will also be stored into [pagingSources].
```
dlam:
```suggestion
     * @return [PagingSource] which will be invalidated when this factory's [invalidate] method is called.
```

dlam: I think it's a bit better to explain directly what happens rather than what happens internally in public docs.

paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
```diff
@@ -0,0 +1,54 @@
...
+     */
+    final override fun invoke(): PagingSource<Key, Value> {
+        return pagingSourceFactory().also { pagingSources.add(it) }
+    }
+
+    /**
+     * Calls [PagingSource.invalidate] on each [PagingSource] stored within [pagingSources]
```
dlam: pagingSources is an internal impl detail, what about

```suggestion
     * Calls [PagingSource.invalidate] on each [PagingSource] that was produced by this [InvalidatingPagingSourceFactory]
```

dlam:
Build should be fixed now - can you rebase on top of `androidx-master-dev` and see if the build passes now?

dlam:
You'll also need to run `./gradlew updateApi` to generate the api file changes!

paging/common/src/main/kotlin/androidx/paging/InvalidatingPagingSourceFactory.kt
```diff
@@ -0,0 +1,58 @@
...
+ * Stores reference to the [PagingSource] factory and the [PagingSource]s it produces for
+ * invalidation when the backing dataset is updated.
+ *
+ * Calling [invalidate] on this [InvalidatingPagingSourceFactory] will automatically forward
+ * invalidate signals to all active [PagingSource]s.
+ *
+ * @param pagingSourceFactory The [PagingSource] factory that returns a PagingSource when called
+ */
+abstract class InvalidatingPagingSourceFactory<Key : Any, Value : Any>(
+    val pagingSourceFactory: () -> PagingSource<Key, Value>
```
dlam: Ah sorry, one more thing - this should be `private val` (will need to re-run updateApi again)

dlam (approved):

dlam:
Hitting this in internal presubmit:

```
> Task :paging:paging-runtime:lintDebug

* What went wrong:
Execution failed for task ':paging:paging-common:ktlint'.
> Process 'command '$CHECKOUT/prebuilts/jdk/jdk11/linux-x86/bin/java'' finished with non-zero exit value 1

* Try:
Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':paging:paging-common:ktlint'.
	at org.gradle...
Caused by: org.gradle.process.internal.ExecException: Process 'command '$CHECKOUT/prebuilts/jdk/jdk11/linux-x86/bin/java'' finished with non-zero exit value 1
	at org.gradle...
	at jdk.internal.reflect.GeneratedMethodAccessor1802.invoke(Unknown Source)
	at java.base...
```

dlam:
Looks like we need to `./gradlew ktlintFormat` since our last changes.

PR #99, opened 2020-11-18
Generalize artifactOrProject helper section parsing

dlam:
Test: GH workflows
Change-Id: I83e78da4ed6c458854e93f6812d9799e8af1e80f


tikurahul (approved):

settings.gradle
```diff
@@ -138,9 +138,9 @@ includeProject(":appcompat:appcompat-resources", "appcompat/appcompat-resources"
 includeProject(":appsearch:appsearch", "appsearch/appsearch", [BuildType.MAIN])
 includeProject(":appsearch:appsearch-compiler", "appsearch/compiler", [BuildType.MAIN])
 includeProject(":appsearch:appsearch-local-storage", "appsearch/local-storage", [BuildType.MAIN])
-includeProject(":arch:core-common", "arch/core-common", [BuildType.MAIN])
```
yigit, line 141: i don't think we should make this change. it is basically giving hte module a weird name w/ a dot.

dlam: How about arch:core:core-*?

PR #101, opened 2020-12-03
Add playground support for androidx.lifecycle

dlam:
Adds a workaround to exclude androidx.lifecycle transitive dependencies from legacy-support-core-utils
resolving an error for "Could not find lifecycle-runtime.aar" when building lifecycle-services.

Test: cd lifecycle && ./gradlew bOS


tikurahul (approved):

dlam:
@googlebot I fixed it.

PR #103, opened 2020-12-07
Avoid exception when calling Activity.reportFullyDrawn()

simonschiller:
## Proposed Changes

  - Avoid a bug on API 19 when calling `Activity.reportFullyDrawn()`

I'm not quite sure if this class is the right place for this workaround, the code could also live in other places, e.g.:
- `androidx.core.app.ComponentActivity`
- `androidx.activity.ComponentActivity`
- `androidx.appcompat.app.AppCompatActivity`

## Testing

Test: Manually tested. If your test harness is also executed on devices with API 19, I could write an automated test for this change if you want.

## Issues Fixed

Fixes: 163239764

### DIFF
--- a/fragment/fragment/src/main/java/androidx/fragment/app/FragmentActivity.java
+++ b/fragment/fragment/src/main/java/androidx/fragment/app/FragmentActivity.java
@@ -19,15 +19,12 @@ package androidx.fragment.app;
 import static androidx.activity.result.contract.ActivityResultContracts.StartIntentSenderForResult;
 import static androidx.annotation.RestrictTo.Scope.LIBRARY_GROUP_PREFIX;
 
-import android.Manifest;
 import android.annotation.SuppressLint;
 import android.app.Activity;
 import android.content.Context;
 import android.content.Intent;
 import android.content.IntentSender;
-import android.content.pm.PackageManager;
 import android.content.res.Configuration;
-import android.os.Build;
 import android.os.Bundle;
 import android.os.Handler;
 import android.os.Parcelable;
@@ -55,7 +52,6 @@ import androidx.annotation.Nullable;
 import androidx.annotation.RestrictTo;
 import androidx.core.app.ActivityCompat;
 import androidx.core.app.SharedElementCallback;
-import androidx.core.content.ContextCompat;
 import androidx.lifecycle.Lifecycle;
 import androidx.lifecycle.LifecycleOwner;
 import androidx.lifecycle.LifecycleRegistry;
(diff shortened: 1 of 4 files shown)


dlam:
Just a heads up that we are in the middle of transitioning to `androidx-main` instead of `androidx-master-dev`, so you may want to base your PR on top of there instead as I'll be disabling PR mirroring to androidx-master-dev this week.

dlam:
(But if this gets in in the next few days you won't run into any issues)

fragment/fragment/src/main/java/androidx/fragment/app/FragmentActivity.java
```diff
@@ -645,6 +649,19 @@ public void startActivityFromFragment(@NonNull Fragment fragment,
...
+    public void reportFullyDrawn() {
+        // On KitKat (API 19) the Activity.reportFullyDrawn() method requires the
+        // UPDATE_DEVICE_STATS permission, otherwise it throws an exception. Instead of throwing,
+        // we fall back to a no-op call here.
+        if (Build.VERSION.SDK_INT == Build.VERSION_CODES.KITKAT && ContextCompat
+                .checkSelfPermission(this, Manifest.permission.UPDATE_DEVICE_STATS)
+                != PackageManager.PERMISSION_GRANTED) {
+            return;
+        }
+        super.reportFullyDrawn();
```
ianhanniballake: Note that this should be guarded with a `Build.VERSION.SDK_INT >= 19` so that `reportFullyDrawn()` can be called on all API levels.

That means it may be more appropriate to re-arrange this method to be something more like:
```
if (Build.VERSION.SDK_INT >= 20) {
  super.reportFullyDrawn();
} else if (Build.VERSION.SDK_INT == 19 && ContextCompat...) {
  // Your current code and comment re: the specific behavior of API 19
}
```

fragment/fragment/src/main/java/androidx/fragment/app/FragmentActivity.java
```diff
@@ -645,6 +649,19 @@ public void startActivityFromFragment(@NonNull Fragment fragment,
         fragment.startActivityForResult(intent, requestCode, options);
     }
 
+    @Override
+    public void reportFullyDrawn() {
+        // On KitKat (API 19) the Activity.reportFullyDrawn() method requires the
+        // UPDATE_DEVICE_STATS permission, otherwise it throws an exception. Instead of throwing,
+        // we fall back to a no-op call here.
+        if (Build.VERSION.SDK_INT == Build.VERSION_CODES.KITKAT && ContextCompat
```
ianhanniballake: The general convention in AndroidX code is to use the numbers themselves, rather than `Build.VERSION_CODES` values. Please use 19 here.

ianhanniballake (changes requested):
Please move this code to `androidx.activity.ComponentActivity` and write a test for this (at the very least, our internal CI runs on API 19 devices and will help serve as a regression test that this method doesn't crash).

ianhanniballake (approved):

dlam:
Sorry for the churn - do you mind actually basing your PR on top of `androidx-master-dev`, looks like it's still a couple days out before we fully switch

simonschiller:
Sure, no problem :heavy_check_mark:

dlam (approved):

ianhanniballake (approved):

dlam (approved):

dlam:
jcenter 502'ing :|

PR #113, opened 2020-12-19
attr/listChoiceBackgroundIndicator misleading documentation

vitiasr:
It is a drawable used as a background for activated list items, not selected ones. There is a difference between a selected and an activated state as stated in Android's documentation. A choice in a list represents a non-transient active state, as opposed to, for example, a transient selection of files in a file manager's list (the one which triggers an ActionMode). This mistake stems from the AOSP itself.

## Testing

Test: Isn't required


dlam:
Sorry for the delay in review, we're not currently taking PRs into appcompat at the moment so the recommended way to get traction on a proposed change to appcompat is to do it through AOSP with the repo tool.

Here's some documentation to get started through that flow: https://android.googlesource.com/platform/frameworks/support/+/refs/heads/androidx-main/README.md#checking-out-the-code

PR #116, opened 2021-01-09
Update PageFetcherSnapshotTest asserts with containsExactly

claraf3:
## Proposed Changes

Update assertThat tests from using isEqualTo to containsExactly.

## Testing

### DIFF
--- a/paging/common/src/test/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
+++ b/paging/common/src/test/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
@@ -467,7 +467,11 @@ class PageFetcherSnapshotTest {
 
             advanceUntilIdle()
             assertThat(fetcherState.newEvents()).containsExactly(
-                LoadStateUpdate<PageEvent<Int>>(loadType = REFRESH, fromMediator = false, loadState = Loading),
+                LoadStateUpdate<PageEvent<Int>>(
+                    loadType = REFRESH,
+                    fromMediator = false,
+                    loadState = Loading
+                ),
                 createRefresh(range = 50..54)
             )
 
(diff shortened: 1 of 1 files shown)


paging/common/src/test/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
```diff
@@ -249,15 +234,13 @@ class PageFetcherSnapshotTest {
...
-            listOf<PageEvent<Int>>(
-                LoadStateUpdate(APPEND, false, Loading),
-                createAppend(
-                    pageOffset = 1,
-                    range = 99..99,
-                    startState = NotLoading.Incomplete,
-                    endState = NotLoading.Complete
-                )
+        assertThat(fetcherState.newEvents()).containsExactly(
+            LoadStateUpdate<PageEvent<Int>>(APPEND, false, Loading),
```
dlam: Are the generics still needed to be explicitly declared here? If not, can we remove them also while we're doing this cleanup?

claraf3: If I don't provide it, "Not enough information to infer type variable T" error shows up. It appears to be required for LoadStateUpdate and Drop events.

dlam:
Thanks for doing this cleanup - can you check if we can also remove the explicit generic declarations?

paging/common/src/test/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
```diff
@@ -110,14 +108,12 @@ class PageFetcherSnapshotTest {
...
-        assertThat(fetcherState.newEvents()).isEqualTo(
-            listOf<PageEvent<Int>>(
-                LoadStateUpdate(PREPEND, false, Loading),
-                createPrepend(
-                    pageOffset = -1,
-                    range = 0..0,
-                    startState = NotLoading.Complete
-                )
+        assertThat(fetcherState.newEvents()).containsExactly(
+            LoadStateUpdate<PageEvent<Int>>(PREPEND, false, Loading),
```
dlam: Shouldn't this be just `LoadStateUpdate<Int>(..)` ?

claraf3: You're right! Updating now.

dlam:
Build should be fixed now - can you rebase your changes on top of latest updates on androidx-main and see if the checks pass now?

dlam (approved):

dlam (approved):

PR #122, opened 2021-01-22
Add fromSavedState handle method to generated nav args

simonschiller:
## Proposed Changes

  - Generate a `fromSavedState` method for all nav args classes, similar to the `fromBundle` one.

## Testing

Test: Adapted existing tests to also include the new method

## Issues Fixed

Fixes: 136967621

### DIFF
--- a/navigation/navigation-safe-args-generator/src/main/kotlin/androidx/navigation/safe/args/generator/java/JavaNavWriter.kt
+++ b/navigation/navigation-safe-args-generator/src/main/kotlin/androidx/navigation/safe/args/generator/java/JavaNavWriter.kt
@@ -248,6 +248,52 @@ class JavaNavWriter(private val useAndroidX: Boolean = true) : NavWriter<JavaCod
             addStatement("return $N", result)
         }.build()
 
+        val fromSavedStateHandleMethod = MethodSpec.methodBuilder("fromSavedStateHandle").apply {
+            addAnnotation(annotations.NONNULL_CLASSNAME)
+            addModifiers(Modifier.PUBLIC, Modifier.STATIC)
+            addAnnotation(specs.suppressAnnotationSpec)
+            val savedStateHandle = "savedStateHandle"
+            addParameter(
+                ParameterSpec.builder(SAVED_STATE_HANDLE_CLASSNAME, savedStateHandle)
+                    .addAnnotation(specs.androidAnnotations.NONNULL_CLASSNAME)
+                    .build()
+            )
+            returns(className)
+            val result = "__result"
+            addStatement("$T $N = new $T()", className, result, className)
+            args.forEach { arg ->
+                beginControlFlow("if ($N.contains($S))", savedStateHandle, arg.name).apply {
+                    addStatement("$T $N", arg.type.typeName(), arg.sanitizedName);
+                    addStatement("$N = $N.get($S)", arg.sanitizedName, savedStateHandle, arg.name)
+                    addNullCheck(arg, arg.sanitizedName)
+                    addStatement(
+                        "$result.$N.put($S, $N)",
+                        specs.hashMapFieldSpec,
+                        arg.name,
+                        arg.sanitizedName
+                    )
+                }
+                if (arg.defaultValue == null) {
+                    nextControlFlow("else")
+                    addStatement(
+                        "throw new $T($S)", java.lang.IllegalArgumentException::class.java,
+                        "Required argument \"${arg.name}\" is missing and does " +
+                            "not have an android:defaultValue"
+                    )
+                } else {
+                    nextControlFlow("else")
+                    addStatement(
+                        "$result.$N.put($S, $L)",
+                        specs.hashMapFieldSpec,
+                        arg.name,
+                        arg.defaultValue.write()
+                    )
+                }
+                endControlFlow()
+            }
+            addStatement("return $N", result)
+        }.build()
+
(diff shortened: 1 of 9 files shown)


navigation/navigation-safe-args-generator/src/main/kotlin/androidx/navigation/safe/args/generator/java/JavaNavWriter.kt
```diff
@@ -248,6 +248,52 @@ class JavaNavWriter(private val useAndroidX: Boolean = true) : NavWriter<JavaCod
             addStatement("return $N", result)
         }.build()
 
+        val fromSavedStateHandleMethod = MethodSpec.methodBuilder("fromSavedStateHandle").apply {
```
danysantiago, line 225: There is a lot of shared code between this method spec and `fromBundleMethod`, lets try to combine it by creating a local function with the common code and then invoking it with the distinct, something like this:
```
fun MethodSpec.Builder.commonFromArgGetCode(
    fromVariableName: String,
    resultVariableName: String,
    arg: Argument,
    argGetBlock: MethodSpec.Builder.() -> Unit
) {
    beginControlFlow("if ($N.contains($S))", fromVariableName, arg.name).apply {
        argGetBlock()
        addNullCheck(arg, arg.sanitizedName)
        addStatement(
            "$resultVariableName.$N.put($S, $N)",
            specs.hashMapFieldSpec,
            arg.name,
            arg.sanitizedName
        )
    }
    if (arg.defaultValue == null) {
        nextControlFlow("else")
        addStatement(
            "throw new $T($S)", java.lang.IllegalArgumentException::class.java,
            "Required argument \"${arg.name}\" is missing and does " +
                "not have an android:defaultValue"
        )
    } else {
        nextControlFlow("else")
        addStatement(
            "$resultVariableName.$N.put($S, $L)",
            specs.hashMapFieldSpec,
            arg.name,
            arg.defaultValue.write()
        )
    }
    endControlFlow()
}
```

then you can use it like this:

```
val fromSavedStateHandleMethod = MethodSpec.methodBuilder("fromSavedStateHandle").apply {
    addAnnotation(annotations.NONNULL_CLASSNAME)
    addModifiers(Modifier.PUBLIC, Modifier.STATIC)
    addAnnotation(specs.suppressAnnotationSpec)
    val savedStateHandle = "savedStateHandle"
    addParameter(
        ParameterSpec.builder(SAVED_STATE_HANDLE_CLASSNAME, savedStateHandle)
            .addAnnotation(specs.androidAnnotations.NONNULL_CLASSNAME)
            .build()
    )
    returns(className)
    val result = "__result"
    addStatement("$T $N = new $T()", className, result, className)
    args.forEach { arg ->
        commonFromArgGetCode(
            fromVariableName = savedStateHandle,
            resultVariableName = result,
            arg= arg
        ) {
            addStatement("$T $N", arg.type.typeName(), arg.sanitizedName)
            addStatement("$N = $N.get($S)", arg.sanitizedName, savedStateHandle, arg.name)
        }
    }
    addStatement("return $N", result)
}.build()
```

navigation/navigation-safe-args-generator/src/main/kotlin/androidx/navigation/safe/args/generator/java/JavaNavWriter.kt
```diff
@@ -248,6 +248,52 @@ class JavaNavWriter(private val useAndroidX: Boolean = true) : NavWriter<JavaCod
...
+                ParameterSpec.builder(SAVED_STATE_HANDLE_CLASSNAME, savedStateHandle)
+                    .addAnnotation(specs.androidAnnotations.NONNULL_CLASSNAME)
+                    .build()
+            )
+            returns(className)
+            val result = "__result"
+            addStatement("$T $N = new $T()", className, result, className)
+            args.forEach { arg ->
+                beginControlFlow("if ($N.contains($S))", savedStateHandle, arg.name).apply {
+                    addStatement("$T $N", arg.type.typeName(), arg.sanitizedName)
```
danysantiago: Looks like the variable declaration and initialization can be done in the same statement

danysantiago (changes requested):
Thanks for making this PR!

navigation/navigation-safe-args-generator/src/main/kotlin/androidx/navigation/safe/args/generator/java/JavaNavWriter.kt
simonschiller: Done for the `JavaNavWriter`.

 I also checked the `KotlinNavWriter` and the only thing that could make sense here is the `else` branch where we read default values, but given that it's only ~7 lines I'd keep it duplicated because I think it makes it more readable that way.

navigation/navigation-safe-args-generator/src/main/kotlin/androidx/navigation/safe/args/generator/java/JavaNavWriter.kt
simonschiller: Not anymore with the newest changes using the common `addReadSingleArgBlock` function 😄

danysantiago (approved):
Looks good!

jbw0033 (approved):

romrell4:
Did this never get merged in? [The docs](https://developer.android.com/jetpack/androidx/releases/navigation) refer to this PR as being the place where this functionality was introduced, but now I'm using androidx.navigation 2.5.3, and I'm still not seeing the `MyFragmentArgs.fromSavedStateHandle` function generated

PR #123, opened 2021-01-24
Add initial infrastructure for FragmentStrictMode

simonschiller:
## Proposed Changes

  - Add initial infrastructure for `FragmentStrictMode`
  
Found this in the issue tracker and think it's a great idea. I tried to keep the first version as lean and as closely aligned to `StrictMode` as possible. I also plan on creating some follow-up PRs to add a few checks (see issue tracker), as soon as the infrastructure is discussed and somewhat agreed upon. For now the classes are restricted to the library, until we have the first few checks.

## Testing

Test: See `FragmentStrictModeTest`

## Issues Fixed

Fixes: 153737341

### DIFF
--- a/fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt
+++ b/fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt
@@ -1,0 +1,84 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.fragment.app.strictmode
+
+import androidx.fragment.app.Fragment
+import androidx.fragment.app.executePendingTransactions
+import androidx.fragment.app.test.FragmentTestActivity
+import androidx.test.core.app.ActivityScenario
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.filters.MediumTest
+import androidx.testutils.withActivity
+import com.google.common.truth.Truth.assertThat
+import org.junit.After
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+
+@MediumTest
+@RunWith(AndroidJUnit4::class)
+public class FragmentStrictModeTest {
+    private lateinit var originalPolicy: FragmentStrictMode.Policy
+
+    @Before
+    public fun setup() {
+        originalPolicy = FragmentStrictMode.getDefaultPolicy()
+    }
+
+    @After
+    public fun teardown() {
+        FragmentStrictMode.setDefaultPolicy(originalPolicy)
+    }
+
+    @Test
+    public fun policyHierarchy() {
+        var lastTriggeredPolicy = ""
+        val violation = object : Violation() {}
+
+        fun policy(name: String) = FragmentStrictMode.Policy.Builder()
+            .penaltyListener { lastTriggeredPolicy = name }
+            .build()
+
+        with(ActivityScenario.launch(FragmentTestActivity::class.java)) {
+            val fragmentManager = withActivity { supportFragmentManager }
+
+            val parentFragment = Fragment()
+            fragmentManager.beginTransaction()
+                .add(parentFragment, "parentFragment")
+                .commit()
+            executePendingTransactions()
+
+            val childFragment = Fragment()
+            parentFragment.childFragmentManager.beginTransaction()
+                .add(childFragment, "childFragment")
+                .commit()
+            executePendingTransactions()
+
+            FragmentStrictMode.setDefaultPolicy(policy("Default policy"))
+            FragmentStrictMode.onPolicyViolation(childFragment, violation)
+            assertThat(lastTriggeredPolicy).isEqualTo("Default policy")
+
+            fragmentManager.strictModePolicy = policy("Parent policy")
+            FragmentStrictMode.onPolicyViolation(childFragment, violation)
+            assertThat(lastTriggeredPolicy).isEqualTo("Parent policy")
+
+            parentFragment.childFragmentManager.strictModePolicy = policy("Child policy")
+            FragmentStrictMode.onPolicyViolation(childFragment, violation)
+            assertThat(lastTriggeredPolicy).isEqualTo("Child policy")
+        }
+    }
+}
--- a/fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt
+++ b/fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt
@@ -47,9 +47,10 @@ public class FragmentStrictModeTest {
 
     @Test
     public fun penaltyDeath() {
-        FragmentStrictMode.setDefaultPolicy(FragmentStrictMode.Policy.Builder()
+        val policy = FragmentStrictMode.Policy.Builder()
             .penaltyDeath()
-            .build())
+            .build()
+        FragmentStrictMode.setDefaultPolicy(policy)
 
         var violation: Violation? = null
         try {
--- a/fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt
+++ b/fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt
@@ -16,7 +16,6 @@
 
 package androidx.fragment.app.strictmode
 
-import androidx.fragment.app.Fragment
 import androidx.fragment.app.StrictFragment
 import androidx.fragment.app.executePendingTransactions
 import androidx.fragment.app.test.FragmentTestActivity
@@ -73,13 +72,13 @@ public class FragmentStrictModeTest {
         with(ActivityScenario.launch(FragmentTestActivity::class.java)) {
             val fragmentManager = withActivity { supportFragmentManager }
 
-            val parentFragment = Fragment()
+            val parentFragment = StrictFragment()
             fragmentManager.beginTransaction()
                 .add(parentFragment, "parentFragment")
                 .commit()
             executePendingTransactions()
 
-            val childFragment = Fragment()
+            val childFragment = StrictFragment()
             parentFragment.childFragmentManager.beginTransaction()
                 .add(childFragment, "childFragment")
                 .commit()
--- a/fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt
+++ b/fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt
@@ -17,6 +17,7 @@
 package androidx.fragment.app.strictmode
 
 import androidx.fragment.app.Fragment
+import androidx.fragment.app.StrictFragment
 import androidx.fragment.app.executePendingTransactions
 import androidx.fragment.app.test.FragmentTestActivity
 import androidx.test.core.app.ActivityScenario
@@ -24,6 +25,7 @@ import androidx.test.ext.junit.runners.AndroidJUnit4
 import androidx.test.filters.MediumTest
 import androidx.testutils.withActivity
 import com.google.common.truth.Truth.assertThat
+import com.google.common.truth.Truth.assertWithMessage
 import org.junit.After
 import org.junit.Before
 import org.junit.Test
@@ -42,6 +44,21 @@ public class FragmentStrictModeTest {
     @After
     public fun teardown() {
         FragmentStrictMode.setDefaultPolicy(originalPolicy)
+    }
+
+    @Test
+    public fun penaltyDeath() {
+        FragmentStrictMode.setDefaultPolicy(FragmentStrictMode.Policy.Builder()
+            .penaltyDeath()
+            .build())
+
+        var violation: Violation? = null
+        try {
+            FragmentStrictMode.onPolicyViolation(StrictFragment(), object : Violation() {})
+        } catch (thrown: Violation) {
+            violation = thrown
+        }
+        assertWithMessage("No exception thrown on policy violation").that(violation).isNotNull()
     }
 
     @Test
--- a/fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java
+++ b/fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java
@@ -3470,6 +3470,7 @@ public abstract class FragmentManager implements FragmentResultOwner {
 
     /** Returns the current policy for this FragmentManager. If no policy is set, returns null. */
     @Nullable
+    @RestrictTo(RestrictTo.Scope.LIBRARY) // TODO: Make API public as soon as we have a few checks
     public FragmentStrictMode.Policy getStrictModePolicy() {
         return mStrictModePolicy;
     }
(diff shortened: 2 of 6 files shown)


fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
```diff
@@ -0,0 +1,197 @@
...
+ * Policy.Builder#penaltyLog} you can watch the output of <code>adb logcat</code> while you
+ * use your application to see the violations as they happen.
+ */
+@SuppressLint("SyntheticAccessor")
+@RestrictTo(RestrictTo.Scope.LIBRARY) // TODO: Make API public as soon as we have a few checks
+public final class FragmentStrictMode {
+    private static final String TAG = "FragmentStrictMode";
+    private static Policy defaultPolicy = Policy.LAX;
+
+    private enum Flag {
```
simonschiller, line 48: If we're concerned about performance, this could also be a bitmask instead

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
```diff
@@ -0,0 +1,197 @@
...
+                FragmentManager fragmentManager = fragment.getParentFragmentManager();
+                if (fragmentManager.getStrictModePolicy() != null) {
+                    return fragmentManager.getStrictModePolicy();
+                }
+            }
+            fragment = fragment.getParentFragment();
+        }
+        return defaultPolicy;
+    }
+
```
simonschiller, line 185: Sample hook for violations could look like this:
```java
public static void onDemoViolation(@NonNull Fragment fragment) {
    Policy policy = getNearestPolicy(fragment);
    if (policy.flags.contains(Flag.DETECT_DEMO)) {
        onPolicyViolation(policy, new DemoViolation());
    }
}
```
This would be called from wherever the violation is triggered.

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
```diff
@@ -0,0 +1,197 @@
...
+             */
+            @NonNull
+            public Builder penaltyDeath() {
+                flags.add(Flag.PENALTY_DEATH);
+                return this;
+            }
+
+            /** Call #{@link OnViolationListener#onViolation} for every violation. */
+            @NonNull
+            public Builder penaltyListener(@NonNull OnViolationListener listener) {
```
simonschiller, line 136: `StrictMode` also requires an `Executor` here on which the listener is called. Could possibly also make sense here :thinking:

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
```diff
@@ -0,0 +1,197 @@
...
+            Log.d(TAG, "FragmentStrictMode policy violation: ", violation);
+        }
+
+        if (policy.listener != null) {
+            policy.listener.onViolation(violation);
+        }
+
+        if (policy.flags.contains(Flag.PENALTY_DEATH)) {
+            Log.e(TAG, "FragmentStrictMode policy violation with PENALTY_DEATH - shutting down.");
+            Process.killProcess(Process.myPid());
```
jbw0033: Killing the process might be extreme here, I think we should throw the Violation instead.

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
jbw0033: No, the Flags are find and make this clear.

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
```diff
@@ -0,0 +1,197 @@
...
+ *
+ * <p>You can decide what should happen when a violation is detected. For example, using {@link
+ * Policy.Builder#penaltyLog} you can watch the output of <code>adb logcat</code> while you
+ * use your application to see the violations as they happen.
+ */
+@SuppressLint("SyntheticAccessor")
+@RestrictTo(RestrictTo.Scope.LIBRARY) // TODO: Make API public as soon as we have a few checks
+public final class FragmentStrictMode {
+    private static final String TAG = "FragmentStrictMode";
+    private static Policy defaultPolicy = Policy.LAX;
```
jbw0033, line 46: Let's do Policy.NONE instead here

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
jbw0033: If we can skip it here we should. Ideally this is as simple as possible.

jbw0033 (changes requested):

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
simonschiller: I got this from `StrictMode`, that's how it reacts to `penaltyDeath` violations: https://cs.android.com/android/platform/superproject/+/master:frameworks/base/core/java/android/os/StrictMode.java;l=2308-2312?q=strictmode

My guess as to why it isn't an exception would be to bypass any `Thread.UncaughtExceptionHandler` that might be registered. But I'd also be fine with just throwing here.

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
simonschiller: Did this for the sake of consistency to `StrictMode` (see [here](https://developer.android.com/reference/android/os/StrictMode.VmPolicy#LAX)), but also fine with changing it if you prefer.

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
```diff
@@ -0,0 +1,204 @@
...
+     */
+    @RestrictTo(RestrictTo.Scope.LIBRARY) // TODO: Make API public as soon as we have a few checks
+    public interface OnViolationListener {
+
+        /** Called on a VM policy violation. */
+        void onViolation(@NonNull Violation violation);
+    }
+
+    /**
+     * {@link FragmentStrictMode} policy applied to a certain thread.
```
ianhanniballake: 'to a certain FragmentManager', not to a certain thread.

fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt
```diff
@@ -0,0 +1,84 @@
...
+        val violation = object : Violation() {}
+
+        fun policy(name: String) = FragmentStrictMode.Policy.Builder()
+            .penaltyListener { lastTriggeredPolicy = name }
+            .build()
+
+        with(ActivityScenario.launch(FragmentTestActivity::class.java)) {
+            val fragmentManager = withActivity { supportFragmentManager }
+
+            val parentFragment = Fragment()
```
ianhanniballake: Use `StrictFragment()` rather than just `Fragment()` here and elsewhere as that gives us some extra debugging information for tests.

fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java
```diff
@@ -3465,6 +3468,25 @@ static int reverseTransit(int transit) {
...
+    /** Returns the current policy for this FragmentManager. If no policy is set, returns null. */
+    @Nullable
+    @RestrictTo(RestrictTo.Scope.LIBRARY) // TODO: Make API public as soon as we have a few checks
+    public FragmentStrictMode.Policy getStrictModePolicy() {
+        return mStrictModePolicy;
+    }
+
+    /**
+     * Sets the policy for what actions should be detected, as well as the penalty if such actions
+     * occur. This policy is specific to this FragmentManager and all children of it. Pass null to
```
ianhanniballake: Let's use the same wording as [`setFragmentFactory()`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java;l=3352):
```
The {@link Fragment#getChildFragmentManager() child FragmentManager} of all Fragments
in this FragmentManager will also use this policy if one is not explicitly set.
```

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
ianhanniballake: I think there's some interesting implications around this, particularly when we want all listeners to be triggered prior to processing the `penaltyDeath` that might inform the structure we want to provide here, even if we don't expose this directly.

For example, do we really want a direct executor to be the default? I could imagine that using the `FragmentHostCallback`'s `Handler` (which, for `FragmentActivity` is the main thread) would be another valid default that would align with most/all of the actual cases we have in mind.

Do you think this work is feasible to do (without a massive rewrite) as a follow up CL? If so, I am okay with just making it more explicit in the Javadoc here with what thread this is being called on (right now, it seems like 'the thread that triggered the policy violation').

One other case I'd like for us to keep in mind is what coroutine friendly API could we build. I don't expect that as part of this initial CL (it is big enough as is), but I'd hate to close that avenue off entirely.

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
```diff
@@ -0,0 +1,204 @@
...
+                    return fragmentManager.getStrictModePolicy();
+                }
+            }
+            fragment = fragment.getParentFragment();
+        }
+        return defaultPolicy;
+    }
+
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    static void onPolicyViolation(@NonNull Fragment fragment, @NonNull Violation violation) {
```
ianhanniballake, line 187: I'm not particularly enthusiastic about this static call from anywhere as I think it precludes the host Handler / Executor discussion above.

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
```diff
@@ -0,0 +1,204 @@
...
+        onPolicyViolation(getNearestPolicy(fragment), violation);
+    }
+
+    private static void onPolicyViolation(@NonNull Policy policy, @NonNull Violation violation) {
+        if (policy.flags.contains(Flag.PENALTY_LOG)) {
+            Log.d(TAG, "FragmentStrictMode policy violation: ", violation);
```
ianhanniballake: FragmentStrictMode is already the tag, you can remove it from the string itself (and the trailing `: `) - just "Policy Violation". I think it would be very helpful to print out the causing Fragment with the log message though so that developers can tie this log message back to a particular fragment.

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
ianhanniballake: `Process.killProcess` means that any crash analytics won't pick up these exceptions, which I think is a big downside. Let's just throw.

ianhanniballake (changes requested):

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
simonschiller: Moving it to the `Handler` of the `FragmentHostCallback` sounds like a good idea to me. The only thing that might cause problems is the guarantee that the `penaltyDeath` handling will happen after all listeners, so we might need to move this onto the `Handler` as well. 

Anyways, should be doable without a big rewrite. For now I added a comment and I'll work on a follow-up PR for this once it is landed.

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
simonschiller: Depends on what we should run on the `Handler`/`Executor` I think. We could either

1. just trigger the listeners or
2. do the whole processing (logging, checking policy, listeners, ...)

on the `Executor`. In either case, I think we can easily switch get the right `Handler`, since we get the violating `Fragment` as parameter. Another option I can think of would be to make `FragmentStrictMode` a singleton and keep a reference to it in each `Fragment`, but I'm not sure if that's actually better.

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
ianhanniballake: Yeah, I expect that we should do a check similar to what [`ensureExecReady()` does](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java;l=1947) to either 1) synchronously trigger listeners / penalty death if we're already on the right looper or 2) post to the `Handler` with two messages: one for dispatching to listeners, the second to trigger the penalty death.

SGTM to do this in a follow up CL, preferably attached to the same underlying issue.

fragment/fragment/src/main/java/androidx/fragment/app/strictmode/FragmentStrictMode.java
ianhanniballake: I'd be okay if the follow up CL we talked about had another version of this that took in the `Handler` did the synchronous / posted dispatch I mentioned in my other comment. That would give the right behavior without necessarily precluding this static method style.

ianhanniballake (approved):

dlam:
Looks like a couple lint errors got thrown, but somehow weren't caught by gh workflows - I think the workflow got confused about which files changed exactly and may have completely skipped them :|

```
$SUPPORT/fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt:50:45: Missing newline after "("
$SUPPORT/fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt:52:20: Missing newline before ")"
```

Could you fix these lint warnings?

You can check using `./gradlew ktlint` from the `fragment` folder.

simonschiller:
Sorry about that, should be fixed now. I actually got a mail notification about the build failure, but the UI showed all green. Weird that this didn't show up in the PR overview :thinking:

fragment/fragment/src/androidTest/java/androidx/fragment/app/strictmode/FragmentStrictModeTest.kt
```diff
@@ -47,9 +47,10 @@ public class FragmentStrictModeTest {
 
     @Test
     public fun penaltyDeath() {
-        FragmentStrictMode.setDefaultPolicy(FragmentStrictMode.Policy.Builder()
+        val policy = FragmentStrictMode.Policy.Builder()
             .penaltyDeath()
-            .build())
+            .build()
+        FragmentStrictMode.setDefaultPolicy(policy)
```
ianhanniballake, line 53: What Lint actually wanted you to do was:
```
FragmentStrictMode.setDefaultPolicy(
    FragmentStrictMode.Policy.Builder()
        .penaltyDeath()
        .build()
)
```

simonschiller: Sure, I just thought this would be more readable. You want me to change it?

dlam: We actually have a `codestyles.xml` that enforces the style ian mentioned just to have some consistency, does that not automatically configure for you when starting studio via `./gradlew studio`?

simonschiller: I used `./gradlew studio`, but could be that I just didn't auto-format before committing.

jbw0033 (approved):

simonschiller:
Is there still something I should do before this can be merged?

dlam:
Interestingly, Copybara failed to mirror your latest commit due to a flake. Let me manually re-trigger.

jbw0033:
Adding the new files means we also needed to update our jetifier files. Added that change to be submitted along with this one. (https://android-review.googlesource.com/c/platform/frameworks/support/+/1592897) Should be able to get this in.

PR #127, opened 2021-02-07
Navigation: Allow parsing a `NavType#ReferenceType` from String

JvmName:
If Bundle contents can be extracted as a Reference, allow parsing the Bundle contents as a Reference. 

When navigating via deeplink URI, the user is required to string-ify navigation arguments in order to append them to the URI as query parameters. This PR supports retrieving a Reference type from a Bundle 

## Proposed Changes
  - replace `NavType#ReferenceType::parseValue`'s method functionality with `Integer.parseInt`

## Testing
Updated `NavTypeTest` to validate that reference types can be inserted and retrieved from a Bundle. 

## Issues Fixed

Fixes: b/179166693


navigation/navigation-common/src/androidTest/res/values/ids.xml
```diff
@@ -0,0 +1,19 @@
...
+
+  Unless required by applicable law or agreed to in writing, software
+  distributed under the License is distributed on an "AS IS" BASIS,
+  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+  See the License for the specific language governing permissions and
+  limitations under the License.
+  -->
+<resources>
+    <item type="id" name="nav_id_reference" />
+    <item type="id" name="nav_id_reference_2" />
```
jbw0033: Is this resource needed?

jbw0033 (changes requested):

navigation/navigation-common/src/androidTest/res/values/ids.xml
JvmName: ah, apologies -- that's from when I thought there was a `Reference[]` type. Will remove

JvmName: done in 1c76b48

jbw0033 (approved):

navigation/navigation-common/src/main/java/androidx/navigation/NavType.java
```diff
@@ -317,8 +317,7 @@ public Integer get(@NonNull Bundle bundle, @NonNull String key) {
         @NonNull
         @Override
         public Integer parseValue(@NonNull String value) {
-            throw new UnsupportedOperationException(
-                    "References don't support parsing string values.");
+            return Integer.parseInt(value);
```
ianhanniballake: Let's match the implementation of `IntType` and ensure we can also handle hex encoded values as well. Write a test for this case as well

ianhanniballake (changes requested):

navigation/navigation-common/src/main/java/androidx/navigation/NavType.java
JvmName: done in 4d26376

JvmName:
> `Error: Unable to deliver Web Hook Error: Request failed with status code 502`

Looks like some part of the build chain is having a bad day -- I'll attempt to rebuild tomorrow, but these changes compile on my local machine FWIW

ianhanniballake (approved):

ianhanniballake:
Yep, we're having some internal infrastructure fun times at the moment. Code looks good here though!

PR #130, opened 2021-02-16
Update LoadStateUpdate to contain full CombinedLoadStates

raduschirliu:
## Proposed Changes
Updated `LoadStateUpdate` to contain the full `CombinedLoadState` object instead of only `LoadType, fromMediator, LoadState`.

## Testing

## Issues Fixed

Fixes: [b/177947936](https://issuetracker.google.com/issues/177947936)


paging/common/src/test/kotlin/androidx/paging/TestPagingSourceExt.kt
```diff
@@ -70,4 +70,4 @@ private fun pages(
         originalPageOffset = pageOffset,
         data = ITEMS.slice(range)
     )
-)
+)
```
dlam: Can we revert this to keep the newline?

paging/common/src/test/kotlin/androidx/paging/ProcessPageEventCallbackCapture.kt
```diff
@@ -50,9 +56,15 @@ class ProcessPageEventCallbackCapture : ProcessPageEventCallback {
         }
     }
 
+    /*
```
dlam: Remove this unused comment block

paging/common/src/test/kotlin/androidx/paging/ProcessPageEventCallbackCapture.kt
```diff
@@ -22,11 +22,17 @@ sealed class PresenterEvent
 data class ChangeEvent(val position: Int, val count: Int) : PresenterEvent()
 data class InsertEvent(val position: Int, val count: Int) : PresenterEvent()
 data class RemoveEvent(val position: Int, val count: Int) : PresenterEvent()
+/*
```
dlam: Remove this unused comment block

paging/common/src/test/kotlin/androidx/paging/PagePresenterTest.kt
```diff
@@ -549,9 +549,12 @@ class PagePresenterTest {
 
     companion object {
         val IDLE_EVENTS = listOf<PresenterEvent>(
+            StateEvent(CombinedLoadStates.IDLE_SOURCE)
+            /*
```
dlam: Unused comment block here?

paging/common/src/main/kotlin/androidx/paging/PagePresenter.kt
```diff
@@ -274,11 +273,13 @@ internal class PagePresenter<T : Any>(
             }
 
             // Dropping from prepend direction implies NotLoading(endOfPaginationReached = false).
+            /*
```
dlam: Unused comment block here

paging/common/src/main/kotlin/androidx/paging/PagePresenter.kt
```diff
@@ -205,9 +200,13 @@ internal class PagePresenter<T : Any>(
                 }
             }
         }
+        /*
```
dlam: Unused comment block here

paging/common/src/main/kotlin/androidx/paging/PagePresenter.kt
```diff
@@ -317,11 +318,13 @@ internal class PagePresenter<T : Any>(
             }
 
             // Dropping from append direction implies NotLoading(endOfPaginationReached = false).
+            /*
```
dlam: Unused comment block here

paging/common/src/main/kotlin/androidx/paging/PagePresenter.kt
```diff
@@ -339,6 +342,7 @@ internal class PagePresenter<T : Any>(
         fun onChanged(position: Int, count: Int)
         fun onInserted(position: Int, count: Int)
         fun onRemoved(position: Int, count: Int)
-        fun onStateUpdate(loadType: LoadType, fromMediator: Boolean, loadState: LoadState)
+        // fun onStateUpdate(loadType: LoadType, fromMediator: Boolean, loadState: LoadState)
```
dlam: Remove

paging/common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt
```diff
@@ -77,6 +77,7 @@ abstract class PagingDataDiffer<T : Any>(
             differCallback.onRemoved(position, count)
         }
 
+        /*
```
dlam: Unused comment block

paging/common/src/main/kotlin/androidx/paging/PageEvent.kt
```diff
@@ -216,6 +199,16 @@ internal sealed class PageEvent<T : Any> {
              */
             internal fun canDispatchWithoutInsert(loadState: LoadState, fromMediator: Boolean) =
                 loadState is LoadState.Loading || loadState is LoadState.Error || fromMediator
+
+            internal fun canDispatchWithoutInsert(combinedLoadStates: CombinedLoadStates): Boolean {
```
dlam: This doesn't seem correct - this was to check whether a LoadStateUpdate should be deferred until next Insert or if it can be sent as a standalone update.

Since we no longer know the previous state due to sending entire snapshots instead of incremental updates, it's no longer possible to do this check here because sending snapshots means we don't know what changed.

We need to move this bit into `PageFetcher` or `PageFetcherSnapshot` itself so it can understand what the previous state was.

dlam: Also should add a test for this since all tests are passing atm although to my understanding this logic isn't correct

paging/common/src/main/kotlin/androidx/paging/PageEvent.kt
```diff
@@ -188,25 +188,8 @@ internal sealed class PageEvent<T : Any> {
...
 
     data class LoadStateUpdate<T : Any>(
-        val loadType: LoadType,
-        val fromMediator: Boolean,
-        val loadState: LoadState // TODO: consider using full state object here
+        val combinedLoadStates: CombinedLoadStates
     ) : PageEvent<T>() {
-        init {
-            // endOfPaginationReached for local refresh is driven by null values in next/prev keys.
-            require(
```
dlam, line 197: Since we can no longer rely on this, we should create some tests around this case to ensure NotLoading can only be set within insert events except in the case of remote state?

dlam (changes requested):

paging/common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt
raduschirliu: Sorry for all the unused comment blocks everywhere, I thought I checked thoroughly enough and removed them all....

paging/common/src/main/kotlin/androidx/paging/PageEvent.kt
raduschirliu: `PageFetcher` and `PageFetcherSnapshot` already seem to be using the correct version of the function, as they have the previous state already. Would it by chance make more sense to do these checks in `CachedPageEventFlow`, as that's the only place this incorrect logic was used?

dlam: This check is to ensure we only update to `LoadState.NotLoading `on Insert events, except in the case of Remote events.

I meant that the previous state is not passed in here, so it's impossible to tell what changed (if a loadType changed to NotLoading). For example, `PageFetcher` is still using this in `dispatchIfValid`.

In order to make this check do the right thing we need to pass in the previous state.

raduschirliu: Ahh okay I see, I think that makes a bit more sense, thank you for clarifying! Just one more question on this quick though:

In the case of the `PageFetcher`'s `dispatchIfValid` function for example, would it still need to pass in the previous state? Since `dispatchIfValid` is only being called from the coroutine right below it (here: https://github.com/androidx/androidx/pull/130/files#diff-4a907aa603257fb93c262b5dd0eab145ff03f06387f7732fc9f77619331a0a2bL145), does it not already perform the required checks before it's called? It seems to be checking if the state has changed first, and if so, calling `dispatchIfValid` with the new state. So in `dispatchIfValid` if we check if the state is `LoadState.NotLoading`, wouldn't it be guaranteed for it to have just changed to that state for any loadType? Additionally, since they're all remote event updates, wouldn't they all be valid to dispatch here regardless? Or am I misunderstanding how that works here?

Sorry for the excessive questions on this one, just trying to make sure I understand the logic here properly :sweat_smile:

dlam: I just realized you also kept the old one - the code you linked above just calls into the override above, so that code path isn't affected. Although that isn't the only place this check is used. It's also used in `CachedPageEventFlow`. Either way, it would be better to make it robust rather than depend on its current use-sites because the latter will change over time.

Btw, instead of sending 3 updates there, you should just send one and call into this helper.

No worries on asking too many questions - it's important to get this bit right.

Regardless of the correctness of the use-sites though, this function not doing the correct thing seems mostly like a source of bugs down the road. Previously we were checking only what changed, now we are checking if any source states are `NotLoading`. I'm actually surprised the tests pass here so we probably should add some tests for this.

We're able to dispatch a load event dependent on what changed - so even if `source.refresh` is `Loading`, if it was already previously `Loading` that shouldn't prevent us from dispatching load state updates, because we might be dispatching the update to inform remote state, which is fine to do as a `LoadStateUpdate`.

The purpose of this, is so that source events is kept synchronous with `Insert` `PageEvents`, which means that code that reacts to `LoadState` can depend on the UI showing the loaded items.

paging/common/src/main/kotlin/androidx/paging/PageFetcher.kt
```diff
@@ -145,16 +141,17 @@ internal class PageFetcher<Key : Any, Value : Any>(
                 accessor.state.collect {
```
dlam, line 141: Btw - now that we send the whole CombinedLoadStates, let's just send a single event instead of three here.

raduschirliu: Sounds good!

paging/common/src/main/kotlin/androidx/paging/PageEvent.kt
raduschirliu: Yup, I had also noticed and fixed the use of it in `CachedPageEventFlow`. That all makes sense, I just made those changes now! Let me know if the logic seems correct this time, and I'll start creating tests around it.

Thank you again for going more in-depth on how that should work, I appreciate it :)

paging/common/src/main/kotlin/androidx/paging/PageFetcher.kt
```diff
@@ -143,18 +139,16 @@ internal class PageFetcher<Key : Any, Value : Any>(
...
-                    }
-                    if (prev.prepend != it.prepend) {
                         loadStates.set(PREPEND, true, it.prepend)
-                        dispatchIfValid(PREPEND, it.prepend)
-                    }
-                    if (prev.append != it.append) {
                         loadStates.set(APPEND, true, it.append)
-                        dispatchIfValid(APPEND, it.append)
+
+                        dispatchIfValid(prevCombinedLoadStates)
```
raduschirliu, line 149: @dlam Would we still want to check if the load states are valid to be dispatched here just to be thorough, or should we just send them without checking? Since as far as I understand, if there are any changes between the `prev` load states and the new states from the accessor, they'll be remote changes and will all be valid to be dispatched?

dlam: These checks have always been around just to be thorough because it's very easy to mess up, so we should keep it.

dlam:
This has been done separately now!

PR #131, opened 2021-02-19
Call FragmentStrictMode listeners on main thread of host

simonschiller:
## Proposed Changes

  - Don't call listeners on the violating thread anymore, instead use the main thread of the host.
  - Follow-up to #123, see [here](https://github.com/androidx/androidx/pull/123#discussion_r568251886) and [here](https://github.com/androidx/androidx/pull/123#discussion_r570648776)

## Testing

Test: `FragmentStrictModeTest#listenerCalledOnCorrectThread`

## Issues Fixed

Fixes: 153737341

### DIFF
--- a/fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java
+++ b/fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java
@@ -63,6 +63,7 @@ import androidx.annotation.StringRes;
 import androidx.collection.ArraySet;
 import androidx.core.os.CancellationSignal;
 import androidx.fragment.R;
+import androidx.fragment.app.strictmode.FragmentStrictMode;
 import androidx.lifecycle.Lifecycle;
 import androidx.lifecycle.LifecycleEventObserver;
 import androidx.lifecycle.LifecycleOwner;
@@ -2896,8 +2897,10 @@ public abstract class FragmentManager implements FragmentResultOwner {
         mLaunchedFragments = new ArrayDeque<>(fms.mLaunchedFragments);
     }
 
+    /** @hide */
     @NonNull
-    FragmentHostCallback<?> getHost() {
+    @RestrictTo(RestrictTo.Scope.LIBRARY)
+    public FragmentHostCallback<?> getHost() {
         return mHost;
     }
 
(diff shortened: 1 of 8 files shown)


simonschiller:
Please note that I moved the `FragmentStrictMode` class from `androidx.fragment.app.strictmode` to `androidx.fragment.app`, so that we can access package-private methods and fields. This probably also requires changes inside Jetifier, like mentioned here: https://github.com/androidx/androidx/pull/123#issuecomment-781546779

ianhanniballake:
> Please note that I moved the `FragmentStrictMode` class from `androidx.fragment.app.strictmode` to `androidx.fragment.app`, so that we can access package-private methods and fields. This probably also requires changes inside Jetifier, like mentioned here: [#123 (comment)](https://github.com/androidx/androidx/pull/123#issuecomment-781546779)

Instead of doing this, can we make [`FragmentManager.getHost()`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java;l=2914) `@hide` + `@RestrictTo(RestrictTo.Scope.LIBRARY)` and the same for [`FragmentHostCallback.getHandler()`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:fragment/fragment/src/main/java/androidx/fragment/app/FragmentHostCallback.java;l=250). That would allow you to use `fragment.getParentFragmentManager().getHost().getHandler()` from the `strictmode` package.

I definitely prefer it in a separate package if possible.

fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java
```diff
@@ -2897,8 +2897,10 @@ void restoreSaveState(@Nullable Parcelable state) {
         mLaunchedFragments = new ArrayDeque<>(fms.mLaunchedFragments);
     }
 
+    /** @hide */
     @NonNull
-    FragmentHostCallback<?> getHost() {
+    @RestrictTo(RestrictTo.Scope.LIBRARY)
```
ianhanniballake, line 2901: Nit here and elsewhere: Always add the `@RestrictTo` annotation as the topmost annotation (i.e., directly below the Javadoc comment where the `@hide` is so that they are visually next to each other in the code).

ianhanniballake (approved):

fragment/fragment/src/main/java/androidx/fragment/app/FragmentManager.java
simonschiller: Makes sense - fixed in 4c0f698

PR #133, opened 2021-02-21
Pojo skipped verification

marcorei:
## Proposed Changes

This change allows a `@Query` method annotated with `@SkipQueryVerification` to compile successfully if it returns a Pojo (or `List<Pojo>` or `Flow<List<Pojo>>` ...). 

- Changes the condition under which `TypeAdapterStore.findRowAdapter` tries to create a `PojoAdapter` to include cases where the verification step was skipped
- Removes the `runtimeQueryPlaceholder` flag which solved a case also covered by this change
- Adds extension methods `XType.isNotVoidObject()` and `XType.isNotKotlinUnit()`

## Testing

Test: Added new `skipVerificationPojo` test to `QueryMethodProcessorTest`; and updated `QueryMethodProcessorText.pojoTest` to not expect compilation failure when skipping verification

### DIFF
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XType.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XType.kt
@@ -233,5 +233,4 @@ internal object KnownTypeNames {
     val BOXED_LONG = TypeName.LONG.box()
     val BOXED_BYTE = TypeName.BYTE.box()
     val KOTLIN_UNIT = ClassName.get("kotlin", "Unit")
-    val STRING = ClassName.get("java.lang", "String")
 }
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XType.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XType.kt
@@ -227,8 +227,18 @@ fun XType.isKotlinUnit(): Boolean = typeName == KnownTypeNames.KOTLIN_UNIT
  */
 fun XType.isByte(): Boolean = typeName == TypeName.BYTE || typeName == KnownTypeNames.BOXED_BYTE
 
+/**
+ * Returns `true` if this represents a [String] type.
+ */
 fun XType.isString(): Boolean = typeName == KnownTypeNames.STRING
 
+/**
+ * Returns `true` if represents very likely a Pojo and none of
+ * the following types:
+ * - Primitive or boxed primitive
+ * - String
+ * - Void
+ */
 fun XType.isPojoByExclusion(): Boolean =
     !isVoid() &&
     !isString() &&
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XType.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XType.kt
@@ -227,10 +227,19 @@ fun XType.isKotlinUnit(): Boolean = typeName == KnownTypeNames.KOTLIN_UNIT
  */
 fun XType.isByte(): Boolean = typeName == TypeName.BYTE || typeName == KnownTypeNames.BOXED_BYTE
 
+fun XType.isString(): Boolean = typeName == KnownTypeNames.STRING
+
+fun XType.isPojoByExclusion(): Boolean =
+    !isVoid() &&
+    !isString() &&
+    !typeName.isPrimitive &&
+    !typeName.isBoxedPrimitive
+
 internal object KnownTypeNames {
     val BOXED_VOID = TypeName.VOID.box()
     val BOXED_INT = TypeName.INT.box()
     val BOXED_LONG = TypeName.LONG.box()
     val BOXED_BYTE = TypeName.BYTE.box()
     val KOTLIN_UNIT = ClassName.get("kotlin", "Unit")
+    val STRING = ClassName.get("java.lang", "String")
 }
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XType.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XType.kt
@@ -227,24 +227,6 @@ fun XType.isKotlinUnit(): Boolean = typeName == KnownTypeNames.KOTLIN_UNIT
  */
 fun XType.isByte(): Boolean = typeName == TypeName.BYTE || typeName == KnownTypeNames.BOXED_BYTE
 
-/**
- * Returns `true` if this represents a [String] type.
- */
-fun XType.isString(): Boolean = typeName == KnownTypeNames.STRING
-
-/**
- * Returns `true` if represents very likely a Pojo and none of
- * the following types:
- * - Primitive or boxed primitive
- * - String
- * - Void
- */
-fun XType.isPojoByExclusion(): Boolean =
-    !isVoid() &&
-    !isString() &&
-    !typeName.isPrimitive &&
-    !typeName.isBoxedPrimitive
-
 internal object KnownTypeNames {
     val BOXED_VOID = TypeName.VOID.box()
     val BOXED_INT = TypeName.INT.box()
--- a/room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt
+++ b/room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt
@@ -532,9 +532,10 @@ class TypeAdapterStore private constructor(
                 rowAdapterLogs?.writeTo(context)
                 return rowAdapter
             }
-            if (query.runtimeQueryPlaceholder) {
+            if (query.runtimeQueryPlaceholder || query.skippedVerification) {
                 // just go w/ pojo and hope for the best. this happens for @RawQuery where we
                 // try to guess user's intention and hope that their query fits the result.
+                // or when @SkipVerification was used on a query that returns a pojo.
                 val pojo = PojoProcessor.createFor(
                     context = context,
                     element = typeElement,
--- a/room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt
+++ b/room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt
@@ -550,12 +550,11 @@ class TypeAdapterStore private constructor(
                 ).process()
                 return PojoRowAdapter(
                     context = context,
-                    info = query.resultInfo,
+                    info = null,
                     pojo = pojo,
                     out = typeMirror
                 )
             }
-
             return null
         } else {
             val singleColumn = findCursorValueReader(typeMirror, null) ?: return null
--- a/room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt
+++ b/room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt
@@ -19,11 +19,13 @@ package androidx.room.solver
 import androidx.room.compiler.processing.XType
 import androidx.room.compiler.processing.isArray
 import androidx.room.compiler.processing.isEnum
-import androidx.room.compiler.processing.isPojoByExclusion
 import androidx.room.ext.CommonTypeNames
 import androidx.room.ext.GuavaBaseTypeNames
 import androidx.room.ext.isEntityElement
 import androidx.room.ext.isNotByte
+import androidx.room.ext.isNotKotlinUnit
+import androidx.room.ext.isNotVoid
+import androidx.room.ext.isNotVoidObject
 import androidx.room.parser.ParsedQuery
 import androidx.room.parser.SQLTypeAffinity
 import androidx.room.processor.Context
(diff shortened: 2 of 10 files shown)


room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt
```diff
@@ -532,9 +532,10 @@ class TypeAdapterStore private constructor(
                 rowAdapterLogs?.writeTo(context)
                 return rowAdapter
             }
-            if (query.runtimeQueryPlaceholder) {
+            if (query.runtimeQueryPlaceholder || query.skippedVerification) {
```
danysantiago: Instead of introducing a new mutable variable we can probably just check if `resultInfo` is not null since that is an indicator that query validation was skipped. We can probably also remove the `runtimeQueryPlaceholder` and rely on the same indicator. Might be good to also update the comment on `resultInfo` to mention `@ SkipQueryVerification`.

marcorei: I agree with this sentiment and initially tried an approach checking if `resultInto` instead of the flag. I had about 20 tests failing which made me think that there might be more cases in which `resultInfo` is not set.

Ideally I'd prefer to handle the Pojo case in line [513 where the EntityProcessor is created in case of a `@SkippedVerification` `@Entity` query](https://github.com/androidx/androidx/blob/f0998b618369cc10568f7cb038dadd2c74d46efd/room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt#L505). 
It might be possible to detect that a `PojoProcessor` is needed by method of elimination looking at the `XType`, i.e. if `typeElement` is not Void, not a primitive or String it's likely a Pojo.

danysantiago: Where the failures from the `room-compiler` side? I think those failures would be expected since `QueryMethodProcessorTest` and some others tests expect that when verification is off, compilation fails with a 'Don't know how to convert XObject from Cursor', in other words, no adapter is found ([see code](https://github.com/androidx/androidx/blob/androidx-main/room/compiler/src/test/kotlin/androidx/room/processor/QueryMethodProcessorTest.kt#L1115)). This change updates that behaviour to now use a Pojo adapter and hope for the best if verification is off.

I think those tests would certainly need to be updated. Right now they are probably passing because `skippedVerification` is missed by them since those compiler tests directly create the processors under test and the verifier is not passed into it as if `@SkipQueryVerification` was placed in the `@Database`, but the method itself does not have the annotation. Which is actually a case not covered by the `skippedVerification` hence my request to use `resultInfo`.

Let me know if you want to try it, otherwise I can probably take over.

marcorei: You were right, I completely missed the fact that those were all pojo tests.

marcorei:
@danysantiago Gave it another try
- Adjusted the tests to not fail on skipped verification for pojo if an adapter was returned
- Removed both `skippedVerification` and `runtimeQueryPlaceholder` flags
- Moved the pojo adapter creation next to entity adapter creation and added some type checks

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XType.kt
```diff
@@ -227,10 +227,29 @@ fun XType.isKotlinUnit(): Boolean = typeName == KnownTypeNames.KOTLIN_UNIT
...
+ * - Primitive or boxed primitive
+ * - String
+ * - Void
+ */
+fun XType.isPojoByExclusion(): Boolean =
+    !isVoid() &&
+    !isString() &&
+    !typeName.isPrimitive &&
+    !typeName.isBoxedPrimitive
+
```
danysantiago: I don't think these methods are needed, but if they are lets move them to https://github.com/androidx/androidx/blob/androidx-main/room/compiler/src/main/kotlin/androidx/room/ext/xtype_ext.kt since they are extensions as interpreted by Room and not part of a general processing ab starction API.

room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt
```diff
@@ -511,6 +512,24 @@ class TypeAdapterStore private constructor(
                         ).process()
                     )
                 }
+
+                // use pojo adapter if we exclude other types.
+                // this happens when @RawQuery or @SkipVerification is used.
+                if (typeMirror.isPojoByExclusion()) {
```
danysantiago: It should be enough to only check for `isVoid`, `isVoidObject` and `isKotlinUnit`, everything else, primitives, String, arrays is handled by a `SingleColumnRowAdapter` and one of the built-in type converters. This is done slightly below in this same method ([here](https://github.com/androidx/androidx/blob/androidx-main/room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt#L522)). Which also means we need to keep this logic down there where `runtimeQueryPlaceholder` used to be so that we try to find a single row adapter first before attempting a Pojo one.

room/compiler/src/test/kotlin/androidx/room/processor/QueryMethodProcessorTest.kt
```diff
@@ -1106,18 +1106,12 @@ class QueryMethodProcessorTest(val enableVerification: Boolean) {
                     handler(null, parsedQuery, invocation)
                 }
             } else {
-                assertThat(adapter, nullValue())
+                assertThat(adapter, notNullValue())
             }
         }
         if (enableVerification) {
```
danysantiago: The `return` can be lifted now with the `if`:
```
return if (enableVerification) {
```

room/compiler/src/test/kotlin/androidx/room/processor/DaoProcessorTest.kt
```diff
@@ -211,6 +211,23 @@ class DaoProcessorTest(private val enableVerification: Boolean) {
         }
     }
 
+    @Test
+    fun skipQueryVerificationPojo() {
```
danysantiago: Can you please move this test to `QueryMethodProcessorTest` since that is more closer to where the behaviour is changing rather than the `DaoProcessor`. The test method `singleQueryMethod` should also help you further verify this by asserting the result binder and its row adapter to be a PojoAdapter.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XType.kt
marcorei: Agreed, they're only needed if the pojo case is checked upfront

room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt
marcorei: Thanks for pointing this out, my previous implementation was also missing the checks for `isVoidObject` and `isKotlinUnit`. After reading your comment I agree that it is better to move the Pojo logic back down. 
The original thought was that it might improve readability, but that was hardly the case.

marcorei:
Implemented the suggestions.

~~`pojo_badQuery` is now failing, because an additional error and warning is being generated.~~

- ~~Error: `The columns returned by the query does not have the fields [maxAge] in foo.bar.MyClass.Pojo even though they are annotated as non-null or primitive. Columns returned by the query: []`~~
- ~~Warning: `foo.bar.MyClass.Pojo has some fields [MAX(ageColumn), name] which are not returned by the query. If they are not supposed to be read from the result, you can mark them with @Ignore annotation. You can suppress this warning by annotating the method with @SuppressWarnings(RoomWarnings.CURSOR_MISMATCH). Columns returned by the query: . Fields in foo.bar.MyClass.Pojo: MAX(ageColumn), name.`~~

~~@danysantiago I wanted to quickly check in with you before adding them to the test because I am not quite sure why they are only being thrown now – and if want them there. They are not necessarily adding useful information.~~

Nevermind, it was a typo :(
There's one less error now and that one makes sense, I reckon.

room/compiler/src/test/kotlin/androidx/room/processor/QueryMethodProcessorTest.kt
```diff
@@ -734,6 +733,30 @@ class QueryMethodProcessorTest(val enableVerification: Boolean) {
...
+                parsedQuery.returnType.typeName,
+                `is`(COMMON.NOT_AN_ENTITY_TYPE_NAME as TypeName)
+            )
+            val adapter = parsedQuery.queryResultBinder.adapter
+            assertThat(checkNotNull(adapter))
+            assertThat(adapter::class, `is`(PojoRowAdapter::class))
+            val rowAdapter = adapter.rowAdapter
+            assertThat(checkNotNull(rowAdapter))
+            assertThat(rowAdapter::class, `is`(PojoRowAdapter::class))
+        }
```
danysantiago: You need to invoke `.compilesWithoutError()`, otherwise it does not really run anything. Once you do that you might notice that the first assertion, `assertThat(adapter::class, `is`(PojoRowAdapter::class))`, is not quite right and likely needs to be asserting that the 'result adapter' is a `SingleEntityQueryResultAdapter`.

room/compiler/src/test/kotlin/androidx/room/processor/QueryMethodProcessorTest.kt
```diff
@@ -946,12 +969,7 @@ class QueryMethodProcessorTest(val enableVerification: Boolean) {
         ) { _, _, _ ->
         }?.failsToCompile()
             ?.withErrorContaining("no such column: age")
-            ?.and()?.withErrorContaining(
-                cannotFindQueryResultAdapter(
-                    ClassName.get("foo.bar", "MyClass", "Pojo")
-                )
-            )
-            ?.and()?.withErrorCount(2)
+            ?.and()?.withErrorCount(1)
```
danysantiago: I don't think this needs to be removed, when db verification is 'off' the `pojoTest` does not return the 'assertion subject' which means this 'withError' assertions are only when verification is ON, for which we want Room to fail since with column verification the Pojo does not match the result.

room/compiler/src/test/kotlin/androidx/room/processor/QueryMethodProcessorTest.kt
```diff
@@ -1106,19 +1124,13 @@ class QueryMethodProcessorTest(val enableVerification: Boolean) {
...
+            assertion
         } else {
-            assertion.failsToCompile()
-                .withErrorContaining(
-                    cannotFindQueryResultAdapter(
-                        ClassName.get("foo.bar", "MyClass", "Pojo")
-                    )
-                )
-            return null
+            null
```
danysantiago, line 1139: before this `null` there needs to be a `assertion.compilesWithoutError()`, otherwise we are really not running compilation tests for when the verification is off

room/compiler/src/main/kotlin/androidx/room/solver/TypeAdapterStore.kt
```diff
@@ -532,9 +535,13 @@ class TypeAdapterStore private constructor(
                 rowAdapterLogs?.writeTo(context)
                 return rowAdapter
             }
-            if (query.runtimeQueryPlaceholder) {
-                // just go w/ pojo and hope for the best. this happens for @RawQuery where we
-                // try to guess user's intention and hope that their query fits the result.
+
+            // use pojo adapter as a last resort.
+            // this happens when @RawQuery or @SkipVerification is used.
+            if (typeMirror.isNotVoid() &&
```
danysantiago: It might be redundant, but just to make it clear can you please also add a `query.resultInfo == null `, this should also fix the need to change `pojo_badQuery`, and just to make it clear, we only want to try this pojo guessing if verification is off to make sure user opt-in to this.

danysantiago (commented):
Just a few more comments, but it's looking good, thanks for working on this.

marcorei:
Adding `resultInfo == null` solved the the issue with the bad query test indeed. Thanks for helping me with this PR.

danysantiago (approved):
Looks good to me, thanks!

danysantiago:
By the way, would be amazing if you can update the PR's description, since that is what gets converted as the commit message when squashing to merge.

marcorei:
Updated the description. 
Can I ignore the build failures in the other projects?

danysantiago:
I think so, @dlam might know more, the way merging actually work is that we [mirror the CL Gerrit](https://android-review.googlesource.com/c/platform/frameworks/support/+/1605977) and presubmit is run there, once merged then a bot comes and closes this PR.

dlam:
Feel free to ignore the build failures in the other projects for now - we can ignore the github failures as long as presubmit passed in the Gerrit mirror.

marcorei:
Hm, the presubmit seems to fail as well. Please let me know if I can help fixing that.

dlam:
You should be able to just rebase on top of `androidx-main` to get the fixes!

PR #137, opened 2021-03-09
Adds KotlinPoet integration to Room Processing

gfreivasc:
## Proposed Changes

Adds `write()` method that accepts KotlinPoet's an instance of
`FileSpec`. Also adds KotlinPoet as an API dependency and an extension
function similar to that of JavaPoet.

Also verifies if processing backend is set for generating kotlin code
and fails for KotlinPoet if it isn't.

## Testing

Test: ./gradlew test connectedCheck without benchmarks

## Fixes
Fixes: [b/182195680](https://issuetracker.google.com/issues/182195680)

### DIFF
--- a/room/compiler-processing-testing/src/test/java/androidx/room/compiler/processing/util/GeneratedCodeMatchTest.kt
+++ b/room/compiler-processing-testing/src/test/java/androidx/room/compiler/processing/util/GeneratedCodeMatchTest.kt
@@ -20,6 +20,9 @@ import com.google.common.truth.Truth.assertThat
 import com.squareup.javapoet.JavaFile
 import com.squareup.javapoet.TypeName
 import com.squareup.javapoet.TypeSpec
+import com.squareup.kotlinpoet.BOOLEAN
+import com.squareup.kotlinpoet.FileSpec
+import com.squareup.kotlinpoet.TypeSpec as KTypeSpec
 import org.junit.Test
 import org.junit.runner.RunWith
 import org.junit.runners.Parameterized
@@ -107,6 +110,72 @@ class GeneratedCodeMatchTest internal constructor(
             actual = Line(
                 pos = 4,
                 content = "boolean bar;"
+            )
+        )
+        assertThat(result.exceptionOrNull()).hasMessageThat().contains(mismatch.toString())
+    }
+
+    @Test
+    fun successfulGeneratedKotlinCodeMatch() {
+        // java environment will not generate kotlin files
+        if (runTest.toString() == "java") return
+
+        val file = FileSpec.builder("foo.bar", "Baz")
+            .addType(KTypeSpec.classBuilder("Baz").build())
+            .build()
+        runTest { invocation ->
+            if (invocation.processingEnv.findTypeElement("foo.bar.Baz") == null) {
+                invocation.processingEnv.filer.write(file)
+            }
+            invocation.assertCompilationResult {
+                generatedSource(
+                    Source.kotlin("foo/bar/Baz.kt", file.toString())
+                )
+            }
+        }
+    }
+
+    @Test
+    fun missingGeneratedKotlinCode_mismatch() {
+        // java environment will not generate kotlin files
+        if (runTest.toString() == "java") return
+
+        val generated = FileSpec.builder("foo.bar", "Baz")
+            .addType(
+                KTypeSpec.classBuilder("Baz")
+                    .addProperty("bar", BOOLEAN)
+                    .build()
+            )
+            .build()
+        val expected = FileSpec.builder("foo.bar", "Baz")
+            .addType(
+                KTypeSpec.classBuilder("Baz")
+                    .addProperty("foo", BOOLEAN)
+                    .build()
+            )
+            .build()
+
+        val result = runCatching {
+            runTest { invocation ->
+                if (invocation.processingEnv.findTypeElement("foo.bar.Baz") == null) {
+                    invocation.processingEnv.filer.write(generated)
+                }
+                invocation.assertCompilationResult {
+                    generatedSource(
+                        Source.kotlin("foo/bar/Baz.kt", expected.toString())
+                    )
+                }
+            }
+        }
+
+        val mismatch = SourceFileMismatch(
+            expected = Line(
+                pos = 6,
+                content = "val foo: Boolean"
+            ),
+            actual = Line(
+                pos = 6,
+                content = "val bar: Boolean"
             )
         )
         assertThat(result.exceptionOrNull()).hasMessageThat().contains(mismatch.toString())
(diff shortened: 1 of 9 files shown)


gfreivasc:
@googlebot I signed it!

yigit (changes requested):
overall lgtm, thanks!
Can you add a test to GeneratedCodeMatchTest to ensure we find these files properly as well.

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/compiler-processing-testing/src/test/java/androidx/room/compiler/processing/util/GeneratedCodeMatchTest.kt

gfreivasc:
Sure, working on it

gfreivasc:
@yigit Is there anyway I can get the current generation path? I've stumbled upon two situations:

- Java environment will not generate the Kotlin file, I could either ignore that scenario since it targets Java codebases (?).
- `Sources.kotlin()` API isn't as simple as it does not take the qualified name, and internally the matcher expects the full path of the generated code instead of just the relative path, as seen on [Source.kt#L132](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/Source.kt;drc=48f734a99aa4adc5364f5d4b7827c688e1637d83;l=132).

I am investigating what can be done about the latter but not sure how to handle the java scenario yet.

yigit:
> @yigit Is there anyway I can get the current generation path? I've stumbled upon two situations:
> 
> * Java environment will not generate the Kotlin file, I could either ignore that scenario since it targets Java codebases (?).
We would need to solve it eventually but does not have to be in this CL
> * `Sources.kotlin()` API isn't as simple as it does not take the qualified name, and internally the matcher expects the full path of the generated code instead of just the relative path, as seen on [Source.kt#L132](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/Source.kt;drc=48f734a99aa4adc5364f5d4b7827c688e1637d83;l=132).
> 
> I am investigating what can be done about the latter but not sure how to handle the java scenario yet.

We can maybe ignore kotlin code verification for now and instead throw an exception that it is not supported if an assertion receives a kotlin file w/ the java processor (to avoid false positives).
XTestInvocation has an `isKsp` property that can be used in the test side to avoid calling that method if desired.

I'm also not against changing Source.kotlin to require a qName but that would be inconsistent with how kotlin works so I'd rather don't do that for now. We can resolve it later.

gfreivasc:
> We can maybe ignore kotlin code verification for now and instead throw an exception that it is not supported if an assertion receives a kotlin file w/ the java processor (to avoid false positives).
> XTestInvocation has an `isKsp` property that can be used in the test side to avoid calling that method if desired.

I've managed to run the tests by verifying if TestRunner's `toString()` is `"java"` and skipping the test if positive. This allows to validate for kapt which is expected to work.

> I'm also not against changing Source.kotlin to require a qName but that would be inconsistent with how kotlin works so I'd rather don't do that for now. We can resolve it later.

I agree, and modified it to take the relative path instead of the absolute. All tests are passing (the CI looks to have broken but they're passing locally).

room/compiler-processing-testing/src/test/java/androidx/room/compiler/processing/util/GeneratedCodeMatchTest.kt
```diff
@@ -111,4 +114,70 @@ class GeneratedCodeMatchTest internal constructor(
         )
         assertThat(result.exceptionOrNull()).hasMessageThat().contains(mismatch.toString())
     }
+
+    @Test
+    fun successfulGeneratedKotlinCodeMatch() {
+        // java environment will not generate kotlin files
+        if (runTest.toString() == "java") return
```
yigit:
```suggestion
        if (runTest.toString() == "java") throw AssumptionViolatedException()
```
so that test is reported as skippped instead of succeeded.

room/compiler-processing-testing/src/test/java/androidx/room/compiler/processing/util/GeneratedCodeMatchTest.kt
```diff
@@ -111,4 +114,70 @@ class GeneratedCodeMatchTest internal constructor(
...
+                    Source.kotlin("foo/bar/Baz.kt", file.toString())
+                )
+            }
+        }
+    }
+
+    @Test
+    fun missingGeneratedKotlinCode_mismatch() {
+        // java environment will not generate kotlin files
+        if (runTest.toString() == "java") return
```
yigit: same here. eventually we should probaby change TestRunner instead of equals check but not important for this PR.
it also probably makes sense to move this to a helper function `assumeCanGenerateKotlinCode` that throws assumption violation.

yigit (commented):
lgtm, just small nits in the test.

you can ignore presubmit, i think we broke playground again, it will work fine in AOSP if it is passing locally (just don't rebase until we fix playground:) )

yigit:
> lgtm, just small nits in the test.
> 
> you can ignore presubmit, i think we broke playground again, it will work fine in AOSP if it is passing locally (just don't rebase until we fix playground:) )

it should be fixed now: https://android-review.googlesource.com/c/platform/frameworks/support/+/1624289

gfreivasc:
> > lgtm, just small nits in the test.
> > you can ignore presubmit, i think we broke playground again, it will work fine in AOSP if it is passing locally (just don't rebase until we fix playground:) )
> 
> it should be fixed now: https://android-review.googlesource.com/c/platform/frameworks/support/+/1624289

Ok, I will try rebasing now. I also applied precondition checks for running KotlinPoet on javaAP without kapt. I can remove those if you don't think that's applicable. Will push the changes once tests pass.

Edit: playground still broken to me after rebase

yigit (approved):
lgtm, thanks for adding the kapt parameter check as well.

if CL is fine, it should pass in AOSP which is what matters anyways. github might block it but if it is the only one, we'll skip.

Approving so that it gets mirrored into AOSP

yigit:
the aosp build failed due to some unused imports.

```
Detected these failing tasks: [':room:room-compiler-processing:ktlint']
> Task :room:room-compiler-processing:ktlint
$SUPPORT/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacFiler.kt:22:1: Unused import

> Task :room:room-compiler-processing:ktlint
"checkstyle" report written to $OUT_DIR/androidx/room/room-compiler-processing/build/reports/ktlint/ktlint-checkstyle-report.xml

```

You can validate these locally as described here:
https://github.com/androidx/androidx/blob/androidx-main/CONTRIBUTING.md#validating-changes-locally

Looks like the lint changes are still breaking playground. Locally, you can workaround it by commenting the source sets in `../lint-checks/build.gradle`.

PR #139, opened 2021-03-12
[FragmentStrictMode] Detect calls to Fragment#setTargetFragment

simonschiller:
## Proposed Changes

  - Detect calls to `Fragment#setTargetFragment`

## Testing

Test: See `FragmentStrictModeTest#detectSetTargetFragment`

## Issues Fixed

Fixes: 153737745


simonschiller:
Right now, this check only detects calls to `Fragment#setTargetFragment`. We could also extend it to detect calls to `Fragment#getTargetFragment` and `Fragment#getTargetRequestCode`, as those methods belong to the same category. What do you think?

jbw0033:
We should do all of the methods.

The long term goal for strict mode is to remove any APIs that cause violations. So if you have strict mode on and your app has no problems, then when you upgrade to say Fragment 2.0 where all the APIs that cause violations have been removed, it should be like a no-op. So although, calling the `getters` without the `setter` is not very useful, we should make sure we are actually catching those calls and reporting them as violations.

simonschiller:
Makes sense, I adapted the PR (as well as #140).

jbw0033 (approved):

PR #141, opened 2021-03-12
[FragmentStrictMode] Detect <fragment> tag usage

simonschiller:
## Proposed Changes

  - Detect `<fragment>` tag usage inside XML layouts

## Testing

Test: See `FragmentStrictModeTest#detectFragmentTagUsage`

## Issues Fixed

Fixes: 153738235

### DIFF
--- a/fragment/fragment/src/main/java/androidx/fragment/app/FragmentLayoutInflaterFactory.java
+++ b/fragment/fragment/src/main/java/androidx/fragment/app/FragmentLayoutInflaterFactory.java
@@ -27,6 +27,7 @@ import android.view.ViewGroup;
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 import androidx.fragment.R;
+import androidx.fragment.app.strictmode.FragmentStrictMode;
 
 class FragmentLayoutInflaterFactory implements LayoutInflater.Factory2 {
     private static final String TAG = FragmentManager.TAG;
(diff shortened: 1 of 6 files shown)


fragment/fragment/src/main/java/androidx/fragment/app/FragmentLayoutInflaterFactory.java
```diff
@@ -130,6 +131,7 @@ public View onCreateView(@Nullable final View parent, @NonNull String name,
                         + "re-attached via the <fragment> tag: id=0x" + Integer.toHexString(id));
             }
         }
+        FragmentStrictMode.onFragmentTagUsage(fragment);
```
jbw0033, line 134: I think we should make the call on line 74 before we check the ID and tag. The idea being that if you are using strict mode, we should throw this before we even check if you are using <fragment> tag correctly.

jbw0033 (changes requested):

fragment/fragment/src/main/java/androidx/fragment/app/FragmentLayoutInflaterFactory.java
simonschiller: I wanted to do that originally, but the problem there is that we don't have a `Fragment` instance yet. If there is a reliable way to traverse up the `FragmentManager` hierarchy without a concrete `Fragment` instance, that would also work, but I couldn't find one. We need the traversal to figure out the correct policy.

jbw0033: There is a package private findFragmentManager that finds the FragmentManager associated with a given view, but I don't think that helps here much.

jbw0033 (approved):

PR #143, opened 2021-03-16
Adds element deferring for KSP to Room Processing

gfreivasc:
## Proposed Changes

Creates a map of `XTypeElement` to their originating `KSAnnotated` that can
be used to return elements deferred on `XProcessingStep` back to the KSP
`SymbolProcessor`.

## Testing

```
Test: ./gradlew \
    test connectedCheck \
    -x :room:room-benchmark:cC \
```

## Issues Fixed

Fixes: [b/182822821](https://issuetracker.google.com/182822821)

### DIFF
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
@@ -22,6 +22,7 @@ import androidx.room.compiler.processing.ksp.KspProcessingEnv
 import com.google.auto.common.BasicAnnotationProcessor
 import com.google.auto.common.MoreElements
 import com.google.common.collect.SetMultimap
+import com.google.devtools.ksp.symbol.KSAnnotated
 import com.google.devtools.ksp.symbol.KSClassDeclaration
 import javax.annotation.processing.ProcessingEnvironment
 import javax.lang.model.element.Element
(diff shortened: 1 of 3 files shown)


room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
```diff
@@ -66,18 +67,22 @@ interface XProcessingStep {
...
+        val elementMap = mutableMapOf<XTypeElement, KSAnnotated>()
         val args = annotations().associateWith { annotation ->
             val elements = env.resolver.getSymbolsWithAnnotation(
                 annotation.java.canonicalName
             ).filterIsInstance<KSClassDeclaration>()
                 .map {
-                    env.requireTypeElement(it.qualifiedName!!.asString())
+                    val element = env.requireTypeElement(it.qualifiedName!!.asString())
+                    elementMap[element] = it
+                    element
```
yigit: you could also cast this to KspTypeElement to get the declaration out of it but this soolution is fine too.

yigit (approved):
Thank you!

yigit:
btw we need to update room as well but i can also do it as a followup.

https://github.com/androidx/androidx/blob/androidx-main/room/compiler/src/main/kotlin/androidx/room/RoomKspProcessor.kt#L56

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
gfreivasc: I think it's more cohesive to do that, indeed

gfreivasc:
Changed the implementation to use the reference that `KspTypeElement` holds. also changes KspRoomProcessor to make use of the API changes.

dlam:
Build should be passing again now - if you rebase this checks should pass.

PR #144, opened 2021-03-17
[Navigation] Fix deeplink domain parsing being case sensitive

bentrengrove:
## Proposed Changes

The current regex pattern matcher was doing a case sensitive check of the domain and arguments. Simply changing it to case insensitive seems to be all that's required to fix this bug. It is important that parameters are still case sensitive but this is still the case as parameter matching is done by the `Uri` code not the `NavDeepLink` code. I have written tests to cover these cases.

If it is deemed too risky to just run the pattern matcher with `CASE_INSENSITIVE` I believe the regex will have to be redesigned not to use `\Q\E`.

## Testing

- Tested using the sample project and deep linking via `adb shell am start -a android.intent.action.VIEW -d https://www.Iana.org/domains/second`
- Added new unit tests to the project for the different case insensitive and sensitive sections

## Issues Fixed
https://issuetracker.google.com/issues/153829033
Fixes: b/153829033

### DIFF
--- a/navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDeepLinkTest.kt
+++ b/navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDeepLinkTest.kt
@@ -1076,7 +1076,7 @@ class NavDeepLinkTest {
             .that(matchArgs?.containsKey("path"))
             .isFalse()
     }
-    
+
     @Test
     fun deepLinkCaseSensitiveParams() {
         val deepLinkString = "$DEEP_LINK_EXACT_HTTP/?myParam={param}"
--- a/navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDeepLinkTest.kt
+++ b/navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDeepLinkTest.kt
@@ -1057,7 +1057,7 @@ class NavDeepLinkTest {
 
         val id = 2
         val matchArgs = deepLink.getMatchingArguments(
-            Uri.parse("${DEEP_LINK_EXACT_HTTPS.toUpperCase()}/users/${id}/posts"),
+            Uri.parse("${DEEP_LINK_EXACT_HTTPS.toUpperCase()}/users/$id/posts"),
             mapOf("id" to intArgument())
         )
         assertWithMessage("Args should not be null")
@@ -1075,9 +1075,10 @@ class NavDeepLinkTest {
 
         val id = 2
         val matchArgs = deepLink.getMatchingArguments(
-            Uri.parse(deepLinkArgument
-                .replace("{id}", id.toString())
-                .replace("users", "Users")
+            Uri.parse(
+                deepLinkArgument
+                    .replace("{id}", id.toString())
+                    .replace("users", "Users")
             ),
             mapOf("id" to intArgument())
         )
@@ -1098,8 +1099,10 @@ class NavDeepLinkTest {
         val deepLinkUpper = deepLinkString
             .replace("myParam", "MYPARAM")
             .replace("{param}", param.toString())
-        val matchArgs = deepLink.getMatchingArguments(Uri.parse(deepLinkUpper),
-            mapOf("param" to intArgument()))
+        val matchArgs = deepLink.getMatchingArguments(
+            Uri.parse(deepLinkUpper),
+            mapOf("param" to intArgument())
+        )
 
         assertWithMessage("Args should be not be null")
             .that(matchArgs)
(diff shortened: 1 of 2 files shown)


navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDeepLinkTest.kt
```diff
@@ -1049,4 +1049,72 @@ class NavDeepLinkTest {
             .that(matchArgs?.getInt("id"))
             .isEqualTo(id)
     }
+
+    @Test
+    fun deepLinkCaseInsensitiveDomain() {
```
jbw0033: Does this case actually fail before the added change? If not, let's omit it.

navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDeepLinkTest.kt
```diff
@@ -1049,4 +1049,72 @@ class NavDeepLinkTest {
...
+        val deepLinkString = DEEP_LINK_EXACT_HTTP
+        val deepLink = NavDeepLink(deepLinkString)
+
+        val deepLinkUpper = deepLinkString.toUpperCase()
+        assertThat(deepLink.matches(Uri.parse(deepLinkUpper)))
+    }
+
+    @Test
+    fun deepLinkCaseInsensitiveDomainRegexMatching() {
```
jbw0033: deepLinkCaseInsensitiveDomainWithPath

navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDeepLinkTest.kt
```diff
@@ -1049,4 +1049,72 @@ class NavDeepLinkTest {
...
+        assertWithMessage("Args should not be null")
+            .that(matchArgs)
+            .isNotNull()
+        assertWithMessage("Args should contain the id")
+            .that(matchArgs?.getInt("id"))
+            .isEqualTo(id)
+    }
+
+    @Test
+    fun deepLinkCaseInsensitiveArguments() {
```
jbw0033: deepLinkCaseInsensitivePath

navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDeepLinkTest.kt
```diff
@@ -1049,4 +1049,72 @@ class NavDeepLinkTest {
...
+        assertWithMessage("Args should not be null")
+            .that(matchArgs)
+            .isNotNull()
+        assertWithMessage("Args should contain the id")
+            .that(matchArgs?.getInt("id"))
+            .isEqualTo(id)
+    }
+
+    @Test
+    fun deepLinkCaseSensitiveParams() {
```
jbw0033: deepLinkCaseSensitiveQueryParams

navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDeepLinkTest.kt
```diff
@@ -1049,4 +1049,72 @@ class NavDeepLinkTest {
...
+        val deepLinkString = "$DEEP_LINK_EXACT_HTTP/?myParam={param}"
+        val deepLink = NavDeepLink(deepLinkString)
+
+        val param = 2
+        val deepLinkUpper = deepLinkString
+            .replace("myParam", "MyParam")
```
jbw0033: Make the 2nd param all uppercase "MYPARAM" so this is easier to see.

navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDeepLinkTest.kt
```diff
@@ -1049,4 +1049,72 @@ class NavDeepLinkTest {
...
+            .replace("{param}", param.toString())
+        val matchArgs = deepLink.getMatchingArguments(Uri.parse(deepLinkUpper),
+            mapOf("param" to intArgument()))
+
+        assertWithMessage("Args should be not be null")
+            .that(matchArgs)
+            .isNotNull()
+        assertWithMessage("Args bundle should be empty")
+            .that(matchArgs!!.isEmpty)
```
jbw0033: Use ? instead of !!.

navigation/navigation-common/src/main/java/androidx/navigation/NavDeepLink.kt
```diff
@@ -422,7 +422,7 @@ public class NavDeepLink internal constructor(
             // specifically escape any .* instances to ensure
             // they are still treated as wildcards in our final regex
             val finalRegex = uriRegex.toString().replace(".*", "\\E.*\\Q")
-            pattern = Pattern.compile(finalRegex)
+            pattern = Pattern.compile(finalRegex, Pattern.CASE_INSENSITIVE)
```
jbw0033, line 425: @ianhanniballake Since this is on the finalRegex, it is making the entire pattern including params case insensitive, but based on the test above, it looks like params still aren't matching, which is what we want. Wanted your thoughts on this method, vs always checking the domain in the lowercase or something similar.

jbw0033 (changes requested):

jbw0033 (approved):

ianhanniballake:
Thanks for fixing this!

PR #153, opened 2021-04-03
[Navigation] Make deep link path arguments to not include slash

osipxd:
## Proposed Changes

Slash is used for URI path splitting. Arguments should not include more than one path section.
I purpose to replace `(.+?)` with `([^/]+?)` in resulting regexp.

## Testing

Added corresponing tests to `NavDestinationAndroidTest`

## Issues Fixed

Fixes: [b/184072811](https://issuetracker.google.com/issues/184072811)

### DIFF
--- a/navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDestinationAndroidTest.kt
+++ b/navigation/navigation-common/src/androidTest/java/androidx/navigation/NavDestinationAndroidTest.kt
@@ -112,6 +112,26 @@ class NavDestinationAndroidTest {
     }
 
     @Test
+    fun matchDeepLinkBestMatchPathTail() {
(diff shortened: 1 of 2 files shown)


osipxd:
@googlebot I signed it!

ianhanniballake (approved):

geak:
This breaks cases when argument needs to have a slash in it:

scheme://foo/{page}

page = bar/xyz

So DL scheme://foo/bar/xyz doesn't work 😢

ianhanniballake:
If you want special characters included as part of an argument, you should be encoding them via `Uri.encode`. Navigation will automatically decode them for you as part of its parsing.

525411:
Merge full

PR #154, opened 2021-04-03
[Room][Compiler Processing] Support getting non type elements

elihart:
## Proposed Changes
Currently the Room compiler processing abstraction is set up to use the pattern of processing steps, including using auto-common for javac processing. Our existing processors (and I assume many other people's) do not use this approach and I'd prefer not to have to convert to that pattern. While annotated elements can be looked up directly with XRoundEnv, that currently only works for type elements.

This PR adds to XRoundEnv a new function to look up annotated elements of all kinds, not just type elements. This allows processors more flexibility in how they use the compiler processing (and mirrors the patterns we use with javac processing in our libraries).

The implementation is fairly straightforward, I just created new functions to wrap each possible type of Element and Symbol, leveraging existing classes.

I believe this could benefit from caching, but that could be added in the future.

I also did not attempt to modify XProcessingStep to support elements of all kinds, since the pattern it enforces seems directed towards starting with just type elements.

If this direction is OK I can also update the readme to cover this use case.

## Testing
Test: XRoundEnvTest

Wrote new unit tests that check annotated functions and properties in both java and kotlin are wrapped with the appropriate `X` class.

## Issues Fixed

Bug: b/184438287

### DIFF
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
@@ -88,7 +88,7 @@ internal class JavacProcessingStepDelegate(
         val xEnv = JavacProcessingEnv(env)
         annotations().forEach { annotation ->
             val elements = elementsByAnnotation[annotation].mapNotNull { element ->
-                xEnv.wrapElement(element, annotationName = { annotation })
+                xEnv.wrapAnnotatedElement(element, annotation)
             }.toSet()
             converted[annotation] = elements
         }
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
@@ -18,22 +18,16 @@ package androidx.room.compiler.processing
 
 import androidx.room.compiler.processing.javac.JavacElement
 import androidx.room.compiler.processing.javac.JavacProcessingEnv
+import androidx.room.compiler.processing.ksp.KspElement
 import androidx.room.compiler.processing.ksp.KspProcessingEnv
-import androidx.room.compiler.processing.ksp.KspTypeElement
 import com.google.auto.common.BasicAnnotationProcessor
-import com.google.auto.common.MoreElements
 import com.google.common.collect.ImmutableSetMultimap
 import com.google.devtools.ksp.symbol.KSAnnotated
-import com.google.devtools.ksp.symbol.KSClassDeclaration
 import javax.annotation.processing.ProcessingEnvironment
 import javax.lang.model.element.Element
-import javax.tools.Diagnostic
 
 /**
- * Specialized processing step which only supports annotations on TypeElements.
- *
- * We can generalize it but for now, Room only needs annotations on TypeElements to start
- * processing.
+ * Processing step to simplify processing a set of annotations.
  */
 interface XProcessingStep {
     /**
@@ -46,8 +40,8 @@ interface XProcessingStep {
      */
     fun process(
         env: XProcessingEnv,
-        elementsByAnnotation: Map<String, List<XTypeElement>>
-    ): Set<XTypeElement>
+        elementsByAnnotation: Map<String, Set<XElement>>
+    ): Set<XElement>
 
     /**
      * The set of annotation qualified names processed by this step.
@@ -69,17 +63,12 @@ interface XProcessingStep {
 
     fun executeInKsp(env: XProcessingEnv): List<KSAnnotated> {
         check(env is KspProcessingEnv)
+        val round = XRoundEnv.create(env)
         val args = annotations().associateWith { annotation ->
-            val elements = env.resolver.getSymbolsWithAnnotation(
-                annotation
-            ).filterIsInstance<KSClassDeclaration>()
-                .map {
-                    env.requireTypeElement(it.qualifiedName!!.asString())
-                }
-            elements
+            round.getElementsAnnotatedWith(annotation)
         }
         return process(env, args)
-            .map { (it as KspTypeElement).declaration }
+            .map { (it as KspElement).declaration }
     }
 }
 
@@ -93,23 +82,14 @@ internal class JavacProcessingStepDelegate(
     override fun process(
         elementsByAnnotation: ImmutableSetMultimap<String, Element>
     ): Set<Element> {
-        val converted = mutableMapOf<String, List<XTypeElement>>()
+        val converted = mutableMapOf<String, Set<XElement>>()
         // create a new x processing environment for each step to ensure it can freely cache
         // whatever it wants and we don't keep elements references across rounds.
         val xEnv = JavacProcessingEnv(env)
         annotations().forEach { annotation ->
             val elements = elementsByAnnotation[annotation].mapNotNull { element ->
-                if (MoreElements.isType(element)) {
-                    xEnv.wrapTypeElement(MoreElements.asType(element))
-                } else {
-                    xEnv.delegate.messager.printMessage(
-                        Diagnostic.Kind.ERROR,
-                        "Unsupported element type: ${element.kind}",
-                        element
-                    )
-                    null
-                }
-            }
+                xEnv.wrapElement(element, annotationName = { annotation })
+            }.toSet()
             converted[annotation] = elements
         }
         val result = delegate.process(xEnv, converted)
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XRoundEnv.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XRoundEnv.kt
@@ -16,7 +16,6 @@
 
 package androidx.room.compiler.processing
 
-import androidx.annotation.VisibleForTesting
 import androidx.room.compiler.processing.javac.JavacProcessingEnv
 import androidx.room.compiler.processing.javac.JavacRoundEnv
 import androidx.room.compiler.processing.ksp.KspProcessingEnv
@@ -29,23 +28,11 @@ import kotlin.reflect.KClass
  *
  * @see javax.annotation.processing.RoundEnvironment
  */
-@VisibleForTesting
 interface XRoundEnv {
     /**
      * The root elements in the round.
      */
     val rootElements: Set<XElement>
-
-    /**
-     * Returns the set of [XTypeElement]s that are annotated with the given [annotationQualifiedName].
-     *
-     * Note, Elements that are not Type elements (such as functions) are ignored by this.
-     *
-     * @see getElementsAnnotatedWith
-     */
-    fun getTypeElementsAnnotatedWith(annotationQualifiedName: String): Set<XTypeElement>
-
-    fun getTypeElementsAnnotatedWith(klass: KClass<out Annotation>): Set<XTypeElement>
 
     /**
      * Returns the set of [XElement]s that are annotated with the given annotation [klass].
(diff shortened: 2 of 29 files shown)


room/compiler-processing/src/main/java/androidx/room/compiler/processing/XRoundEnv.kt
```diff
@@ -36,10 +37,23 @@ interface XRoundEnv {
...
-     * Returns the set of [XElement]s that are annotated with the given [annotationQualifiedName].
+     * Returns the set of [XTypeElement]s that are annotated with the given [annotationQualifiedName].
+     *
+     * Note, Elements that are not Type elements (such as functions) are ignored by this.
+     *
+     * @see getElementsAnnotatedWith
      */
     fun getTypeElementsAnnotatedWith(annotationQualifiedName: String): Set<XTypeElement>
 
+    fun getTypeElementsAnnotatedWith(klass: KClass<out Annotation>): Set<XTypeElement>
```
yigit: lets remove these two and we can change the clients in Room to check for type via: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/compiler-processing/src/main/java/androidx/room/compiler/processing/XElement.kt;l=45?q=XElement.kt

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacProcessingEnv.kt
```diff
@@ -183,6 +185,46 @@ internal class JavacProcessingEnv(
         } as T
     }
 
+    // TODO: Add a cache layer
```
yigit: we can skip these caches. particularly, these get cached in the TypeElement anyways and it is very unlikely to find an executable element without a type element reference first.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacProcessingEnv.kt
```diff
@@ -183,6 +185,46 @@ internal class JavacProcessingEnv(
...
+                    env = this,
+                    containing = enclosingType,
+                    element = element
+                )
+            }
+            else -> error("Unsupported kind ${element.kind} of executable element $element")
+        }
+    }
+
+    // TODO: Add a cache layer
```
yigit: ditto

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacProcessingEnv.kt
```diff
@@ -183,6 +185,46 @@ internal class JavacProcessingEnv(
...
+    fun wrapVariableElement(element: VariableElement): JavacVariableElement {
+        return when (val enclosingElement = element.enclosingElement) {
+            is ExecutableElement -> {
+                val executableElement = wrapExecutableElement(enclosingElement)
+
+                executableElement.parameters.find { param ->
+                    param.element == element
```
yigit: java AP does not implement equals for these elements. In this case, i assume it will still work since it will be the same element instance, so we can also do === here. (== fine too)

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacRoundEnv.kt
```diff
@@ -47,4 +52,51 @@ internal class JavacRoundEnv(
             env.wrapTypeElement(MoreElements.asType(it))
         }.toSet()
     }
+
+    override fun getTypeElementsAnnotatedWith(klass: KClass<out Annotation>): Set<XTypeElement> {
+        return getTypeElementsAnnotatedWith(
+            annotationQualifiedName = klass.java.name
```
yigit: i think we want `klass.java.canonicalName` here?
this seems to be the same for most cases but in some cases, it seems to return the binary representation?
From `.name` docs:

```
* <p> Examples:
     * <blockquote><pre>
     * String.class.getName()
     *     returns "java.lang.String"
     * byte.class.getName()
     *     returns "byte"
     * (new Object[3]).getClass().getName()
     *     returns "[Ljava.lang.Object;"
     * (new int[3][4][5][6][7][8][9]).getClass().getName()
     *     returns "[[[[[[[I"
     * </pre></blockquote>
```

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacRoundEnv.kt
```diff
@@ -47,4 +52,51 @@ internal class JavacRoundEnv(
...
+                is TypeElement -> {
+                    env.wrapTypeElement(element)
+                }
+                is ExecutableElement -> {
+                    env.wrapExecutableElement(element)
+                }
+                is PackageElement -> {
+                    error(
+                        "Cannot get elements with annotation ${annotationName()}. Package " +
+                            "elements are not supported by KSP."
```
yigit: maybe instead of KSP, we can put XProcessing to avoid confusion? (since they'll hit this code path only with JavaAP)

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspFieldElement.kt
```diff
@@ -69,4 +69,17 @@ internal class KspFieldElement(
...
+
+    companion object {
+        fun create(
+            env: KspProcessingEnv,
+            declaration: KSPropertyDeclaration
+        ): KspFieldElement {
+            return KspFieldElement(
+                env = env,
+                declaration = declaration,
+                containing = declaration.requireEnclosingTypeElement(env)
```
yigit, line 81: i wonder what happens with these when the field's enclosing element is a File ?
We may want to check for that and throw an error as unsupported, otherwise, the exception may not be very clear.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspRoundEnv.kt
```diff
@@ -37,4 +42,37 @@ internal class KspRoundEnv(
...
+        )
+    }
+
+    override fun getElementsAnnotatedWith(klass: KClass<out Annotation>): Set<XElement> {
+        return getElementsAnnotatedWith(
+            annotationQualifiedName = klass.qualifiedName ?: error("No qualified name for $klass")
+        )
+    }
+
+    override fun getElementsAnnotatedWith(annotationQualifiedName: String): Set<XElement> {
```
yigit, line 40: this is not related to xprocessing but it might make sense to add some tests w/ AnnotationTargets to ensure KSP handles these gracefuly (and if not file a bug).

room/compiler-processing/src/test/java/androidx/room/compiler/processing/XRoundEnvTest.kt
```diff
@@ -0,0 +1,113 @@
...
+        )
+
+        runProcessorTest(listOf(source)) { testInvocation ->
+            val annotatedElements = testInvocation.roundEnv.getElementsAnnotatedWith(
+                OtherAnnotation::class
+            )
+
+            val targetElement = testInvocation.processingEnv.requireTypeElement("Baz")
+
+            Truth.assertThat(
```
yigit: nit: we use static import for Truth (though i wouldn't be surprised if there are other instances in the codebase :) ).

yigit (commented):
thank you!

overall lgtm, just small nit comments.

yigit:
Btw, the commit message needs:

Bug: b/xxx
Test: <test name you've added>

for the bug, i've created b/184438287

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacRoundEnv.kt
elihart: Sure, I can change it. I thought about this and `name` should be the same as canonicalName for all types. the `canonicalName` implementation seems to use some reflection and other java processing to compute the value, while `name` is implemented natively so I figured it was more efficient to use that, but it is likely an over optimization if at all

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspFieldElement.kt
elihart: Good catch, this fails for top level properties and functions in KSP (javac works). I'll add a clearer error message and file a bug for this support. I don't need it personally for our usages but it seems like a good thing to support.

elihart: I created b/184526463 for this

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspRoundEnv.kt
elihart: What is the ideal way for KSP to handle a wrong target? I tried adding a unit test, and compilation does fail because of that, but it seems like KSP does run first before compilation failure (at least in the xprocessing test harness), which could cause confusing errors.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
```diff
@@ -18,22 +18,16 @@ package androidx.room.compiler.processing
...
 import com.google.devtools.ksp.symbol.KSAnnotated
-import com.google.devtools.ksp.symbol.KSClassDeclaration
 import javax.annotation.processing.ProcessingEnvironment
 import javax.lang.model.element.Element
-import javax.tools.Diagnostic
 
 /**
- * Specialized processing step which only supports annotations on TypeElements.
- *
- * We can generalize it but for now, Room only needs annotations on TypeElements to start
```
yigit, line 35: nice to see this removed :) thanks!!!

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacProcessingEnv.kt
```diff
@@ -183,6 +188,69 @@ internal class JavacProcessingEnv(
         } as T
     }
 
+    internal inline fun wrapElement(
```
yigit: passing annotation here feels very weird.
I see you are passing it in to have a better error message.

I have two possible options:

a) rename method to `wrapAnnotatedElement` and also pass the annotationName as a parameter (rather then callback)
b) define an explict exception and catch it in the call site to append the annotation name to the user visible error.

I'm equal on both options so feel free to pick one or a third one.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspFieldElement.kt
yigit: thanks. we should eventually model packages here as well but we'll do it once we have a user :)

room/compiler/src/test/kotlin/androidx/room/writer/SQLiteOpenHelperWriterTest.kt
```diff
@@ -214,7 +215,8 @@ class SQLiteOpenHelperWriterTest {
             sources = sources + databaseCode
         ) { invocation ->
             val db = invocation.roundEnv
-                .getTypeElementsAnnotatedWith(androidx.room.Database::class.qualifiedName!!)
+                .getElementsAnnotatedWith(androidx.room.Database::class.qualifiedName!!)
+                .filterIsInstance<XTypeElement>()
```
yigit, line 219: thanks for doing these, sorry i didn't notice it would be this many :facepalm:

yigit (approved):
Just one small comment.
I'll approve this so that it gets synced to aosp where I can run the full presubmit test
(we used to run more checks in AOSP, idk if it is still the case)

yigit:
the build failed with:

```
e: $SUPPORT/room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/ProcessorTestExt.kt: (141, 43): Not enough information to infer type variable V
e: $SUPPORT/room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/ProcessorTestExt.kt: (142, 39): Unresolved reference: getTypeElementsAnnotatedWith

```

Which does not make sense since it passed here, very likely we've conflicted :/

https://android-review.googlesource.com/c/platform/frameworks/support/+/1665205

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacProcessingEnv.kt
elihart: Agreed, I can do a). I missed this comment originally, sorry!

PR #155, opened 2021-04-06
[Room][Compiler Processing] Originating element improvements

elihart:
## Proposed Changes
Right now the compiler processing library only exposes a way to set originating elements when generating a file with JavaPoet, and KotlinPoet should be supported for the case of creating kotlin files. Additionally when writing a file with KSP the incremental mode defaults to isolating with no way to specify that it should be aggregating.

This change:
  - Adds a new boolean parameter to XFiler `write` functions to optionally specify isolating vs aggregating behavior (defaults to isolating)
  - Adds a kotlinpoet extension function to add originating elements.
  - Exposes a public function to convert XElement to an originating element in case users need to access it directly

## Testing

Test: OriginatingElementsTest

Unit tests that check originating elements are created from XElement and added to kotlin and java poet type builders.

## Issues Fixed

Fixes: https://issuetracker.google.com/issues/184568922

### DIFF
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/JavaPoetExt.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/JavaPoetExt.kt
@@ -51,8 +51,9 @@ internal fun TypeMirror.safeTypeName(): TypeName = if (kind == TypeKind.NONE) {
  * Adds the given element as the originating element for compilation.
  * see [TypeSpec.Builder.addOriginatingElement].
  */
-fun TypeSpec.Builder.addOriginatingElement(element: XElement) {
+fun TypeSpec.Builder.addOriginatingElement(element: XElement): TypeSpec.Builder {
     element.originatingElement()?.let(this::addOriginatingElement)
+    return this
 }
 
 internal fun TypeName.rawTypeName(): TypeName {
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/JavaPoetExt.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/JavaPoetExt.kt
@@ -50,7 +50,7 @@ internal fun TypeMirror.safeTypeName(): TypeName = if (kind == TypeKind.NONE) {
  * see [TypeSpec.Builder.addOriginatingElement].
  */
 fun TypeSpec.Builder.addOriginatingElement(element: XElement): TypeSpec.Builder {
-    element.originatingElement()?.let(this::addOriginatingElement)
+    element.originatingElementForPoet()?.let(this::addOriginatingElement)
     return this
 }
 
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/JavaPoetExt.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/JavaPoetExt.kt
@@ -52,11 +52,7 @@ internal fun TypeMirror.safeTypeName(): TypeName = if (kind == TypeKind.NONE) {
  * see [TypeSpec.Builder.addOriginatingElement].
  */
 fun TypeSpec.Builder.addOriginatingElement(element: XElement) {
-    if (element is JavacElement) {
-        this.addOriginatingElement(element.element)
-    } else if (element is KspElement) {
-        element.containingFileAsOriginatingElement()?.let(this::addOriginatingElement)
-    }
+    element.originatingElement()?.let(this::addOriginatingElement)
 }
 
 internal fun TypeName.rawTypeName(): TypeName {
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/JavaPoetExt.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/JavaPoetExt.kt
@@ -15,9 +15,7 @@
  */
 package androidx.room.compiler.processing
 
-import androidx.room.compiler.processing.javac.JavacElement
 import androidx.room.compiler.processing.javac.JavacExecutableElement
-import androidx.room.compiler.processing.ksp.KspElement
 import androidx.room.compiler.processing.ksp.KspMethodElement
 import androidx.room.compiler.processing.ksp.KspMethodType
 import com.squareup.javapoet.ClassName
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XElement.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XElement.kt
@@ -17,6 +17,7 @@
 package androidx.room.compiler.processing
 
 import androidx.room.compiler.processing.javac.JavacElement
+import androidx.room.compiler.processing.ksp.KSFileAsOriginatingElement
 import androidx.room.compiler.processing.ksp.KspElement
 import javax.lang.model.element.Element
 import kotlin.contracts.contract
@@ -83,13 +84,14 @@ fun XElement.isConstructor(): Boolean {
  * Attempts to get a Javac [Element] representing the originating element for attribution
  * when writing a file for incremental processing.
  *
- * Note, instead of using this directly you can instead use the [addOriginatingElement] extension
- * functions for JavaPoet and KotlinPoet Type builders.
+ * In KSP a [KSFileAsOriginatingElement] will be returned, which is a synthetic javac element
+ * that allows us to pass originating elements to JavaPoet and KotlinPoet, and later extract
+ * the KSP file when writing with [XFiler].
  */
-fun XElement.originatingElement(): Element? {
+internal fun XElement.originatingElementForPoet(): Element? {
     return when (this) {
         is JavacElement -> element
         is KspElement -> containingFileAsOriginatingElement()
-        else -> null
+        else -> error("Originating element not implemented for ${this.javaClass}")
     }
 }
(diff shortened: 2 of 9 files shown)


room/compiler-processing/src/main/java/androidx/room/compiler/processing/JavaPoetExt.kt
```diff
@@ -51,12 +51,9 @@ internal fun TypeMirror.safeTypeName(): TypeName = if (kind == TypeKind.NONE) {
  * Adds the given element as the originating element for compilation.
  * see [TypeSpec.Builder.addOriginatingElement].
  */
-fun TypeSpec.Builder.addOriginatingElement(element: XElement) {
-    if (element is JavacElement) {
```
elihart, line 55: Pulled this implementation out to share with the kotlin poet extension, and to make it public in case any one wants their own custom file writing.

I also changed this to return the builder, which is the pattern that javapoet uses and allows better chaining

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KSFileAsOriginatingElement.kt
```diff
@@ -30,7 +30,7 @@ import javax.lang.model.type.TypeMirror
  * This wrapper class helps us wrap a KSFile as an originating element and KspFiler unwraps it to
  * get the actual KSFile out of it.
  */
-internal class KSFileAsOriginatingElement(
+internal data class KSFileAsOriginatingElement(
```
elihart, line 33: Made this a data class to support equality for testing purposes

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XElement.kt
```diff
@@ -75,3 +78,18 @@ fun XElement.isConstructor(): Boolean {
...
 }
+
+/**
+ * Attempts to get a Javac [Element] representing the originating element for attribution
+ * when writing a file for incremental processing.
+ *
+ * Note, instead of using this directly you can instead use the [addOriginatingElement] extension
+ * functions for JavaPoet and KotlinPoet Type builders.
+ */
+fun XElement.originatingElement(): Element? {
```
yigit: So far, I don't think we ever exposed JavaAP types from the API.
Maybe, it is better to have a more explicit name not to get confused with XElement? Maybe something like:

`originatingJavaAPElement`
or
`originatingElementForJavaPoet`
or
`originatingElementForPoet` to also cover for KotlinPoet?


might also make sense to move this to a real method in XElement interface so that if we ever have another type of it, we won't forget to implement this. But I'm fine with this implementation as well, just an idea.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XFiler.kt
```diff
@@ -23,11 +23,22 @@ import com.squareup.kotlinpoet.FileSpec
  * Code generation interface for XProcessing.
  */
 interface XFiler {
-    fun write(javaFile: JavaFile)
+    /**
+     * @param aggregating Specifies whether this
```
yigit: lets make this an enum (can still have the default).
Boolean makes it harder to read if not called w/ named parameters.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KSFileAsOriginatingElement.kt
yigit: sgtm

yigit (commented):
overall lgtm, just minor nits.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XElement.kt
elihart: sure, I can move `originatingElementForPoet` to the interface

elihart: actually I realized that publicly exposing this isn't actually that helpful, since if someone did want to use it for some custom writing they wouldn't be able to do so in KSP since they can't access the `KSFileAsOriginatingElement`. We would need to provide a more thought out abstraction for that, which seems overkill for now since it is unlikely someone is using neither javapoet or kotlinpoet, so I'll just go back to having this be internal

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XElement.kt
```diff
@@ -75,3 +79,19 @@ fun XElement.isConstructor(): Boolean {
...
+
+/**
+ * Attempts to get a Javac [Element] representing the originating element for attribution
+ * when writing a file for incremental processing.
+ *
+ * In KSP a [KSFileAsOriginatingElement] will be returned, which is a synthetic javac element
+ * that allows us to pass originating elements to JavaPoet and KotlinPoet, and later extract
+ * the KSP file when writing with [XFiler].
+ */
+internal fun XElement.originatingElementForPoet(): Element? {
```
yigit, line 91: oh was this always internal? i thought it was going to be public. anyways, i prefer internal unless we have to make it public (mostly because KSFileAsOriginatingElement is not a true implementation)

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XElement.kt
```diff
@@ -75,3 +79,19 @@ fun XElement.isConstructor(): Boolean {
...
+    return when (this) {
+        is JavacElement -> element
+        is KspElement -> containingFileAsOriginatingElement()
+        else -> error("Originating element not implemented for ${this.javaClass}")
```
yigit: nit: is not implemented

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacFiler.kt
```diff
@@ -22,11 +22,13 @@ import com.squareup.kotlinpoet.FileSpec
 import javax.annotation.processing.ProcessingEnvironment
 
 internal class JavacFiler(val processingEnv: ProcessingEnvironment) : XFiler {
-    override fun write(javaFile: JavaFile) {
+
+    // "aggregating" is ignored in javac, and only applicable in KSP
+    override fun write(javaFile: JavaFile, aggregating: Boolean) {
```
yigit: forgot to push this change? i saw you made it an enum above but it is still a boolean here?

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspFiler.kt
```diff
@@ -32,23 +32,24 @@ internal class KspFiler(
     private val delegate: CodeGenerator,
     private val messager: XMessager,
 ) : XFiler {
-    override fun write(javaFile: JavaFile) {
+    override fun write(javaFile: JavaFile, aggregating: Boolean) {
```
yigit: ditto

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XElement.kt
elihart: you're right, it was public but I changed it back to internal since I had the same thought https://github.com/androidx/androidx/pull/155#discussion_r608116580

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacFiler.kt
elihart: Oh, I forgot to update the implementations

yigit (approved):

yigit:
The build failed with: (in `AutoMigrationWriterTest`)

```
Caused by: java.lang.IllegalStateException: Originating element is not implemented for class androidx.room.compiler.processing.XElement$MockitoMock$1838500372
	at androidx.room.compiler.processing.XElementKt.originatingElementForPoet(XElement.kt:95)
	at androidx.room.compiler.processing.JavaPoetExtKt.addOriginatingElement(JavaPoetExt.kt:53)
	at androidx.room.writer.AutoMigrationWriter.createTypeSpecBuilder(AutoMigrationWriter.kt:50)
	at androidx.room.writer.ClassWriter.write(ClassWriter.kt:68)
	at androidx.room.writer.AutoMigrationWriterTest$validAutoMigrationWithDefaultValue$1.invoke(AutoMigrationWriterTest.kt:77)
	at androidx.room.writer.AutoMigrationWriterTest$validAutoMigrationWithDefaultValue$1.invoke(AutoMigrationWriterTest.kt:32)
```

you should be able to repro with a rebase. I thought your change wouldn't change the behavior so maybe i missed that.

yigit:
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/compiler/src/test/kotlin/androidx/room/writer/AutoMigrationWriterTest.kt;l=76?q=AutoMigrationWriterTest

elihart:
Fixed. The tests started failing because I changed the originating element function throw an error if an unknown type is encountered, which failed for a mocked XElement - fixed by using a real implementation in the test

elihart:
Navigation lint seems to be failing for unrelated reasons, not sure I can do anything about it besides try rebasing later and hope its fixed upstream

```
/home/runner/work/androidx/androidx/navigation/navigation-dynamic-features-fragment/src/main/java/androidx/navigation/dynamicfeatures/fragment/ui/AbstractProgressFragment.kt:70: Error: Upgrade Fragment version to at least 1.3.0. [InvalidFragmentVersionForActivityResult]
    private val intentSenderLauncher = registerForActivityResult(

yigit:
Don't worry about it, i can override it in aosp.
Navigation/ fragments have a weird build structure with prebuilts that
breaks time to time in android X.

PR #158, opened 2021-04-14
Add support for Firebase Test Lab

yigit:
This CL introduces a basic setup in AndroidX to run instrumentation
tests on Firebase Test Lab.

This is limited to only application tests for now because FTL requires
two APKs. We can probably work around it if we want to start using FTL
extensively.

Bug: n/a
Test: not sure how to test this so for now, none :)

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -569,6 +569,25 @@ jobs:
           name: artifacts_${{ env.artifact-id }}
           path: ~/dist
 
+      - name: Set up Cloud SDK
+        uses: google-github-actions/setup-gcloud@master
+        with:
+          project_id: ${{ secrets.GCP_PROJECT_ID }}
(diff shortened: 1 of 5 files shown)


digitalbuddha:
Any reason to not use flank or fladle? Both make it easier to download artifacts and shard tests. For instance getting code coverage from ftl. As for library testing, what I do is keep a stub ftl_app application and send that, it doesn't matter what the application apk is when you are sending a lib test apk. Lastly, you can use the affected module detector to selectively assemble apks. https://github.com/dropbox/AffectedModuleDetector/blob/e4e80174988ca64ef7cf2b840a34801de0aade94/affectedmoduledetector/src/main/kotlin/com/dropbox/affectedmoduledetector/AffectedModuleDetectorPlugin.kt#L64 we have a fladle config that wraps this at dropbox. Happy to push it public as well.

tikurahul (approved):

PR #159, opened 2021-04-14
Fix intermittent InvalidationTracker issues in JournalMode.TRUNCATE

bubenheimer:
## Proposed Changes

  - Fixes issue in Room JournalMode.TRUNCATE where the InvalidationTracker callback is sometimes invoked invalidly, too late, or not at all.
  - Adds InvalidationTrackerBehavioralTest

## Testing

Test: ./gradlew test connectedCheck -x :room:room-benchmark:cC -x :room:integration-tests:room-incremental-annotation-processing:test

## Issues Fixed

Fixes: b/154040286

### DIFF
--- a/room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/InvalidationTrackerBehavioralTest.java
+++ b/room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/InvalidationTrackerBehavioralTest.java
@@ -155,7 +155,7 @@ public class InvalidationTrackerBehavioralTest {
                                 //
                                 // If we see non-null here instead of null due to a race then
                                 // our long delay was just too short and we'll need to adjust it
-                                // because the test will have failed. latch.countdown() happens
+                                // because the test will have failed. latch.countDown() happens
                                 // too late in this case but it has no particular effect.
                                 final CountDownLatch latch = mLatch;
                                 if (latch == null) {
@@ -181,7 +181,7 @@ public class InvalidationTrackerBehavioralTest {
                 try {
                     // Resets latch and updates missedInvalidations when change notification failed
                     for (int i = 0; i < iterations; ++i) {
-                        // The Counter table exists just to make InvalidationTracker's life more
+                        // The Counter1 table exists just to make InvalidationTracker's life more
                         // difficult, we are not interested in notifications from this one;
                         // inserts may trigger undefined invalidation callback behavior,
                         // depending on table update timing
@@ -210,7 +210,7 @@ public class InvalidationTrackerBehavioralTest {
                         });
 
                         // Use sufficient delay to give invalidation tracker ample time to catch up;
-                        // this would need to be increased if the test has false positives.
+                        // this would need to be increased if the test had false positives.
                         try {
                             if (!latch.await(10L, TimeUnit.SECONDS)) {
                                 // The tracker still has not been called, log an error
--- a/room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/InvalidationTrackerBehavioralTest.java
+++ b/room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/InvalidationTrackerBehavioralTest.java
@@ -75,13 +75,13 @@ public class InvalidationTrackerBehavioralTest {
     }
 
     @Test
-    public void testInsertsJournalModeTruncate() throws ExecutionException, InterruptedException {
+    public void testInserts_JournalModeTruncate() throws ExecutionException, InterruptedException {
         testInserts(RoomDatabase.JournalMode.TRUNCATE);
     }
 
     @Test
     @SdkSuppress(minSdkVersion = Build.VERSION_CODES.JELLY_BEAN)
-    public void testInsertsJournalModeWAL() throws ExecutionException, InterruptedException {
+    public void testInserts_JournalModeWAL() throws ExecutionException, InterruptedException {
         testInserts(RoomDatabase.JournalMode.WRITE_AHEAD_LOGGING);
     }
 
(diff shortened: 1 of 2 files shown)


bubenheimer:
The remaining failing checks look like upstream problems

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/InvalidationTrackerBehavioralTest.java
```diff
@@ -0,0 +1,254 @@
...
+        mExecutorService.shutdown();
+    }
+
+    @Test
+    public void testInserts() throws ExecutionException, InterruptedException {
+        testInserts(RoomDatabase.JournalMode.TRUNCATE);
+        // Use AUTOMATIC rather than WRITE_AHEAD_LOGGING to support running this test on older
+        // SDKs; effectively old SDKs will exercise the TRUNCATE case a second time this way, but
+        // who cares, it won't hurt
+        testInserts(RoomDatabase.JournalMode.AUTOMATIC);
```
danysantiago: Lets have two different `@Test` and the one with `WRITE_AHEAD_LOGGING` should be annotated with `@SdkSuppress(minSdkVersion = Build.VERSION_CODES.JELLY_BEAN)`, that way we don't do double work on older APIs and the test with WAL will only run in APIs where WAL is available.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/InvalidationTrackerBehavioralTest.java
```diff
@@ -0,0 +1,254 @@
...
+    abstract static class Counter1Dao {
+        @Insert
+        abstract void insert(Counter1 entity);
+    }
+
+    @Dao
+    abstract static class Counter2Dao {
+        @Insert
+        abstract void insert(Counter2 entity);
+    }
```
danysantiago: Lets just have one `@Dao` call it `CounterDao` with both insert methods to reduce some noise in the test, organization is not as important for test code.

room/integration-tests/testapp/src/androidTest/java/androidx/room/integration/testapp/test/InvalidationTrackerBehavioralTest.java
```diff
@@ -0,0 +1,254 @@
...
+                        if (!latch.await(10L, TimeUnit.SECONDS)) {
+                            // The tracker still has not been called, log an error
+                            missedInvalidations.incrementAndGet();
+                        }
+
+                        latch = null;
+                    }
+                } catch (InterruptedException e) {
+                    throw new RuntimeException(e);
+                }
```
danysantiago: If `InterruptedException ` will just throw then move it to just wrap `sleep` and `await`.

danysantiago (commented):
Thanks! I am glad you  were able to workaround your issue with the checkout and got a chance to make this PR.

bubenheimer:
Thanks for the review, code changed as requested

bubenheimer:
I've been thinking of a few more details in the test that I'd like to change if you're on board with any or all of them. All except the first are non-functional changes. I would submit them as separate commits, I have them ready. Sorry for the extra commotion.

- missing removal of ValidationTracker.Observer: messy, should likely be changed
- use a local CountDownLatch variable in table insertion loop instead of unnecessary extra reference to volatile field within a single iteration
- inner class fields should follow 'm' prefix naming convention
- use test method naming convention separating case being tested with underscore
- change DB file naming to align with test class name
- clarify a few comments
- improve test failure explanations

danysantiago:
All of those sounds great, do go ahead adding them.

bubenheimer:
I also wonder what you guys' expectations might be regarding test speed. Right now this test takes about 11 seconds in the emulator on my dated Macbook Pro. I see some more or less reasonable options to improve speeds:

- Reduce iterations from 30 to 10: may be reasonable assuming that the test is run repeatedly over time and is allowed to fail in a flaky manner. Also, does AOSP have a good means to run a parameterized test with (internal) repetitions passed in, or possibly pulled from an external configuration? Only running the whole test repeatedly may be a bad idea (vs. running the inner loop repeatedly), because it may not trigger the flaky behavior.
- Remove some of the staggered delay options (for delay between inserts into the 2 tables): I'd be curious to know what to assume for system clock resolution across various test device hardware & the emulator, now and in the foreseeable future; I've never specifically looked into this and could not find much usable info online. This could inform the removal of delay options. What I mean is, for example, if system clock resolution is unlikely to get down to 10ns or 100ns on any device then those delay tiers make little sense.
- Randomize delay between inserts into the 2 tables: may be reasonable if a single random delay is exercised repeatedly within a single test execution and logged upon failure, to make failures easy to reproduce. Assuming that really major test flakiness is acceptable, so that problems will reveal themselves occasionally over repeated test runs.

danysantiago:
Our build servers read the `@LargeTest` annotation and those are set to timeout after 100 seconds, so I think it is alright that the test takes a bit.

> does AOSP have a good means to run a parameterized test with (internal) repetitions...

AFAIK we don't have this, the `@FlakyTest` actually just makes it so that the test is not run in pre-submit as to not block anyone, but will continuously run on post-submit and if a bot sees it fails to much it will let us know.

>  what to assume for system clock resolution across various test device hardware ...

That is a fantastic question, one thing I can mention is we don't use emulators for android tests in CI, and as new devices are added their hardware will likely be better so keeping those tiers on the long run might make sense.

danysantiago (approved):

danysantiago:
Hey @bubenheimer, can you please rebase your PR? It seems at the time the CI build was failing and it should be fixed now. To pickup the changes you'll need to rebase so a new run starts and if its all green we'll be able to merge.

bubenheimer:
Compose compiler looks broken...

danysantiago:
So sorry to ask again, but can you rebase again. We had to revert some changes to the workflows (https://github.com/androidx/androidx/commit/7f774818067cac3dae6986cc2944dc775c5f9717) which should fix the build now and hopefully merge.

bubenheimer:
Something weird happened there, the IDE seems to have merged changes for some reason instead of rebase, I thought I was using the same workflow as before. LMK if it's still acceptable.

bubenheimer:
I tried to fix it by dropping the merge & doing another rebase, but now something seems screwed up.

bubenheimer:
I rebased one more time, looks like some other changes got in my way at the most recent push. Keeping my fingers crossed. Looks like it's up to the rest of the CI passing now.

PR #160, opened 2021-04-17
[Navigation] Make trailing slash in deep links not required

osipxd:
## Proposed Changes

Allow deep link: `myapp://example.io/messages/`  
To match both of the links:
- `myapp://example.io/messages/`
- `myapp://example.io/messages`

If a deep link pattern contains a trailing slash, make it not required.
If a deep link pattern does not contain a trailing slash, add `\?` to the pattern.

## Testing

Corresponding tests added to NavDeepLinkTest

## Issues Fixed

https://issuetracker.google.com/issues/185532465
Fixes: 185532465

### DIFF
--- a/navigation/navigation-common/src/main/java/androidx/navigation/NavDeepLink.kt
+++ b/navigation/navigation-common/src/main/java/androidx/navigation/NavDeepLink.kt
@@ -85,8 +85,10 @@ public class NavDeepLink internal constructor(
         }
         if (appendPos < uri.length) {
             // Use Pattern.quote() to treat the input string as a literal
-            uriRegex.append(Pattern.quote(uri.substring(appendPos)))
+            // Also remove trailing slashes if it exists so it should be optional
+            uriRegex.append(Pattern.quote(uri.substring(appendPos).trimEnd('/')))
         }
+        uriRegex.append("/?")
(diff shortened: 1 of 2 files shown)


navigation/navigation-common/src/main/java/androidx/navigation/NavDeepLink.kt
```diff
@@ -85,8 +85,10 @@ public class NavDeepLink internal constructor(
         }
         if (appendPos < uri.length) {
             // Use Pattern.quote() to treat the input string as a literal
-            uriRegex.append(Pattern.quote(uri.substring(appendPos)))
+            // Also remove trailing slashes if it exists so it should be optional
+            uriRegex.append(Pattern.quote(uri.substring(appendPos).trimEnd('/')))
```
osipxd, line 89: Is it OK to use `trimEnd` here?

jbw0033: Using it here seems fine.

jbw0033 (approved):

jbw0033:
As stated in https://issuetracker.google.com/185532465, we can't do this since there should be a distinction between having a slash and not have a slash.

PR #162, opened 2021-04-17
WIP expand FTL to fetch results into dist

yigit:
This CL adds support for specifying a Bucket to keep results, which
defaults to a generic bucket I've created for FTL tests.

FTL outputs a bunch of things into this bucket, including a video of the
test running. All of it is pulled back into the out folder after the
test runs. I've also configured a finalizer task to copy that output
into the DIST folder.

Bug: n/a
Test: production :(


yigit:
abandoning this for now. will re-send once the target branch merges into main.

PR #163, opened 2021-04-17
[Room][Compiler Processing] Support getting elements by package name

elihart:
## Proposed Changes
KSP's recent version added the new API "Resolver.getDeclarationsFromPackage()" which is conceptually similar to javac's "Elements.getPackageElement". The Room compiler processing library can include an abstraction in XProcessingEnv to surface both of these and return XElements contained in a given package.

This PR adds the function `getElementsFromPackage` to XProcessingEnv, and has the implementations for KSP and Javac simply use their appropriate package lookup. This was very straightforward; the main source of code changes was to handle cases where xprocessing does not support certain element types, in which case those elements are skipped. I introduced a new exception to throw in this case, which callers can catch.

I'm personally partial to a "result" type monad to represent success and failure in a safer, clearer way, but Room doesn't seem to use that pattern.

## Testing

Test: Updated XRoundEnvTest.kt to test that elements from packages in both local sources and binaries are queried correctly.

## Issues Fixed

Fixes: https://issuetracker.google.com/issues/185609823

### DIFF
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
@@ -151,4 +151,14 @@ interface XProcessingEnv {
             resolver = resolver
         )
     }
+
+    /**
+     * Returns [XTypeElement]s with the given package name. Note that this call can be expensive.
+     *
+     * @param packageName the package name to look up.
+     *
+     * @return A list of [XTypeElement] with matching package name. This will return declarations
+     * from both dependencies and source.
+     */
+    fun getTypeElementsFromPackage(packageName: String): List<XTypeElement>
 }
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
@@ -161,4 +161,6 @@ interface XProcessingEnv {
      * from both dependencies and source.
      */
     fun getTypeElementsFromPackage(packageName: String): List<XTypeElement>
+
+    // TODO: Add support for getting top level members in a package
 }
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacProcessingEnv.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacProcessingEnv.kt
@@ -73,6 +73,16 @@ internal class JavacProcessingEnv(
         return typeElementStore[qName]
     }
 
+    override fun getTypeElementsFromPackage(packageName: String): List<XTypeElement> {
+        // Note, to support Java Modules we would need to use "getAllPackageElements",
+        // but that is only available in Java 9+.
+        val packageElement = delegate.elementUtils.getPackageElement(packageName)
+
+        return packageElement.enclosedElements
+            .filterIsInstance<TypeElement>()
+            .map { wrapTypeElement(it) }
+    }
+
     override fun findType(qName: String): XType? {
         // check for primitives first
         PRIMITIVE_TYPES[qName]?.let {
--- a/room/compiler-processing/src/test/java/androidx/room/compiler/processing/XRoundEnvTest.kt
+++ b/room/compiler-processing/src/test/java/androidx/room/compiler/processing/XRoundEnvTest.kt
@@ -228,11 +228,9 @@ class XRoundEnvTest {
 
         runProcessorTest(listOf(source)) { testInvocation ->
             val elements = testInvocation.processingEnv.getTypeElementsFromPackage("foo")
-
             val targetElement = testInvocation.processingEnv.requireTypeElement(
                 "foo.Baz"
             )
-
             assertThat(
                 elements
             ).contains(targetElement)
--- a/room/compiler-processing/src/test/java/androidx/room/compiler/processing/XRoundEnvTest.kt
+++ b/room/compiler-processing/src/test/java/androidx/room/compiler/processing/XRoundEnvTest.kt
@@ -216,6 +216,46 @@ class XRoundEnvTest {
         }
     }
 
+    @Test
+    fun getElementsFromPackageIncludesSources() {
+        val source = Source.kotlin(
+            "foo/Baz.kt",
+            """
+            package foo
+            class Baz 
+            """.trimIndent()
+        )
+
+        runProcessorTest(listOf(source)) { testInvocation ->
+            val elements = testInvocation.processingEnv.getTypeElementsFromPackage("foo")
+
+            val targetElement = testInvocation.processingEnv.requireTypeElement(
+                "foo.Baz"
+            )
+
+            assertThat(
+                elements
+            ).contains(targetElement)
+        }
+    }
+
+    @Test
+    fun getElementsFromPackageIncludesBinaries() {
+        runProcessorTest { testInvocation ->
+            val kspElements = testInvocation.processingEnv.getTypeElementsFromPackage(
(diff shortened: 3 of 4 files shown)


room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/UnsupportedElementException.kt
```diff
@@ -0,0 +1,22 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.room.compiler.processing.javac
+
+/**
+ * Thrown when processing utils don't yet support a certain Element.
+ */
+class UnsupportedElementException(msg: String) : RuntimeException(msg)
```
elihart: I created a separate class for javac vs ksp, but that does seem unnecessary. Happy to combine them if that is better

yigit:
hmm, I'm not a big fan of using exceptions for control flow, though may not be possible to avoid it here (so no strong stance there). If we go that way, merging the exception makes sense.

Though i have a higher level question.
What is the use case where you needed this?
I'm specifically asking because given that we don't model top level functions/properties yet, nor packages, the only feasible item to fetch here seems like the TypeElements. 
So if this was `getTypeElementsFromPackage`, would that work for your use case? If so, we would also avoid the exception since we model them already. Then the API would be fully functional on what it promises.
Gabrial was looking into the top level members problem so once it is done; then we can extend functionality to include them as well?

elihart:
> hmm, I'm not a big fan of using exceptions for control flow, though may not be possible to avoid it here (so no strong stance there). If we go that way, merging the exception makes sense.

Agreed. If there is a Monad class that you are interested in introducing I'm happy to replace the exceptions with that.

> What is the use case where you needed this?

This KSP ticket explains the use case https://github.com/google/ksp/issues/372

Our use cases do only require TypeElements, so your proposal would work for now. However, we also need general top level function/property support (good to hear that is being worked on); if that is coming soon then do you think it is worth creating `getTypeElementsFromPackage` now and then making a breaking change to switch to to all elements in the near future?

I suppose I  took this approach since I assumed top level function/property support would be available soon, but understand if you don't want to misrepresent the API until then.

yigit:
> Our use cases do only require TypeElements, so your proposal would work for now. However, we also need general top level function/property support (good to hear that is being worked on); if that is coming soon then do you think it is worth creating `getTypeElementsFromPackage` now and then making a breaking change to switch to to all elements in the near future?

I was actually thinking it might be better to have `getTypeElements, getFunctions` etc, similar to how `XTypeElement` does not have `getEnclosedElements` but instead has `getFields, methods`. In fact, we could even consider moving these into an `XPackageElement` class instead of the `XProcessingEnv`.

So once we model top level funcitons etc, we can add them as separate getters. That seems like a better API to me and specifically in this case, it avoids any confusion such that the method does what its name promises.

elihart:
Yeah, that API seems nicer and could save unnecessary work wrapping unneeded elements. The only thing that concerns me though is that ` resolver.getDeclarationsFromPackage(packageName)` is expensive, and calling it multiple times to first get types, then properties and functions, etc would be inefficient. As long we we can cache ` resolver.getDeclarationsFromPackage(packageName)` that seems fine, which lends itself even better to a `XPackageElement`abstraction (although we would need to make sure that the cached values don't hide generated sources from incremental rounds)

Not sure I want to tackle a `XPackageElement` API in this PR. I could just do `getTypeElementsFromPackage` for now and we can add support for the rest later?

elihart:
I updated this to only introduce `XProcessingEnv.getTypeElementsFromPackage` for now. This simplified the approach since it no longer needs to wrap all element types, so I removed all of the exception handling I had before.

This at least adds parity with exiting javac package element lookup, and meets our use case for now. In the future it could be expanded to support top level members, and we may want to consider caching the element lookup if it is expensive to repeat for both types and members.

room/compiler-processing/src/test/java/androidx/room/compiler/processing/XRoundEnvTest.kt
```diff
@@ -222,6 +222,46 @@ class XRoundEnvTest {
...
+            assertThat(
+                elements
+            ).contains(targetElement)
+        }
+    }
+
+    @Test
+    fun getElementsFromPackageIncludesBinaries() {
+        runProcessorTest { testInvocation ->
+            val kspElements = testInvocation.processingEnv.getTypeElementsFromPackage(
```
yigit, line 243: nice hack :D i usually compile a library in test to achieve this but this is good enough.

yigit (approved):

elihart:
Lint failed on `fatal: Invalid revision range 0f32fe709402389b9f2cddca3679df0f805ba62d..HEAD` - I'm not sure if there was a separate underlying issue that I am not spotting in the github checks.

I have rebased locally, but the build now fails with what seems to be an issue on master
```
> Task :androidx-plugin:gradle-plugin:compileKotlin FAILED
e: /Users/elihart/repos/androidx/buildSrc/src/main/kotlin/androidx/build/ReportLibraryMetricsTask.kt: (31, 12): Unresolved reference: json
e: /Users/elihart/repos/androidx/buildSrc/src/main/kotlin/androidx/build/ReportLibraryMetricsTask.kt: (62, 20): Unresolved reference: JSONObject
e: /Users/elihart/repos/androidx/buildSrc/src/main/kotlin/androidx/build/metalava/CheckApiEquivalenceTask.kt: (20, 27): Unresolved reference: io
e: /Users/elihart/repos/androidx/buildSrc/src/main/kotlin/androidx/build/metalava/CheckApiEquivalenceTask.kt: (100, 10): Unresolved reference: FileUtils
```

yigit:
yea the lint stuff here is buggy. we need to fix that, meanwhile, rebasing from androidx-main should fix it.
but anyways, presubmit here doess not matter if it passes the aosp preseubmit.

sorry i lost track of these, is this ready to merge or are you planning any changes?

elihart:
This is ready to merge! I have nothing else to add.

That error I shared above is happening after rebasing with `androidx-main` though so I'm not sure what to do about that. I just pushed a rebased version of the commit

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacProcessingEnv.kt
```diff
@@ -73,6 +73,16 @@ internal class JavacProcessingEnv(
         return typeElementStore[qName]
     }
 
+    override fun getTypeElementsFromPackage(packageName: String): List<XTypeElement> {
```
yigit, line 76: FYI i just merged support for top level members so we can support them as a followup to this.

yigit (approved):

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
```diff
@@ -151,4 +151,14 @@ interface XProcessingEnv {
...
+
+    /**
+     * Returns [XTypeElement]s with the given package name. Note that this call can be expensive.
+     *
+     * @param packageName the package name to look up.
+     *
+     * @return A list of [XTypeElement] with matching package name. This will return declarations
+     * from both dependencies and source.
+     */
+    fun getTypeElementsFromPackage(packageName: String): List<XTypeElement>
```
danysantiago, line 163: Leave a `TODO: Add support for top level members` ?

yigit:
build failure:

```
https://android-review.googlesource.com/c/platform/frameworks/support/+/1691387/1/room/compiler-processing/src/test/java/androidx/room/compiler/processing/XRoundEnvTest.kt#231
```
Just an extra space

elihart:
Thanks, addressed.

danysantiago (approved):

PR #164, opened 2021-04-19
[Room][Compiler Processing] Add additional type element information

elihart:
## Proposed Changes
Adds an interface function to get element kdoc. Implements this for javac, and uses null for KSP until https://github.com/google/ksp/issues/392 is fixed

Also surfaces more XTypeElement information, such as whether the type is a data class, a fun interface, a companion object, etc.

I implemented everything exposed in kotlin class metadata that seemed reasonable in order to future proof it a bit.

I was not able to implement fun interface checking for KSP since it is not yet surfaced in KSP - https://github.com/google/ksp/issues/393

## Testing

Test: Added cases to XTypeElementTest.modifiers

Note, I wasn't able to test the `expect` modifier. First I had to add support for custom compiler arguments so I could enable multiplatform via `"-Xmulti-platform"` but to get the test sources to compile I had to include a `actual` implementation as well, and since both the expected and actual implementations were in the same sources it seems the processor only picks up the actual. Omitting "actual" implementation doesn't work as compilation fails with "no actual implementation found in module", and I'm not sure how to have the compiler processing testing work with multiple module sources. The implementation of the "expect" modifier is very simple so I'm not sure how important the test is.

## Issues Fixed

Fixes: https://issuetracker.google.com/issues/185672887

### DIFF
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt
@@ -37,7 +37,11 @@ abstract class XMessager {
         onPrintMessage(kind, msg, element)
     }
 
-    protected abstract fun onPrintMessage(kind: Diagnostic.Kind, msg: String, element: XElement? = null)
+    protected abstract fun onPrintMessage(
+        kind: Diagnostic.Kind,
+        msg: String,
+        element: XElement? = null
+    )
 
     fun addMessageWatcher(watcher: XMessager) {
         watchers.add(watcher)
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt
@@ -37,7 +37,7 @@ abstract class XMessager {
         onPrintMessage(kind, msg, element)
     }
 
-    abstract fun onPrintMessage(kind: Diagnostic.Kind, msg: String, element: XElement? = null)
+    protected abstract fun onPrintMessage(kind: Diagnostic.Kind, msg: String, element: XElement? = null)
 
     fun addMessageWatcher(watcher: XMessager) {
         watchers.add(watcher)
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
@@ -65,9 +65,47 @@ interface XTypeElement : XHasModifiers, XElement, XMemberContainer {
     fun isInterface(): Boolean
 
     /**
-     * Returns `true` if this [XTypeElement] is declared as a Kotlin `object`
+     * Returns `true` if this [XTypeElement] represents a Kotlin functional interface,
+     * i.e. marked with the keyword `fun`
+     */
+    fun isFunctionalInterface(): Boolean
+
+    /**
+     * Returns `true` if this [XTypeElement] represents an ordinary class. ie not an enum, object,
+     * annotation, interface, or other type of specialty class.
+     */
+    fun isClass(): Boolean
+
+    /**
+     * Returns `true` if this [XTypeElement] represents a Kotlin data class
+     */
+    fun isDataClass(): Boolean
+
+    /**
+     * Returns `true` if this [XTypeElement] represents a Kotlin value class
+     */
+    fun isValueClass(): Boolean
+
+    /**
+     * Returns `true` if this [XTypeElement] represents a class with the Kotlin `expect` keyword
+     */
+    fun isExpect(): Boolean
+
+    /**
+     * Returns `true` if this [XTypeElement] represents a Kotlin annotation class or a Java
+     * annotation type.
+     */
+    fun isAnnotationClass(): Boolean
+
+    /**
+     * Returns `true` if this [XTypeElement] is a non-companion `object` in Kotlin
      */
     fun isKotlinObject(): Boolean
+
+    /**
+     * Returns `true` if this [XTypeElement] is declared as a Kotlin `companion object`
+     */
+    fun isCompanionObject(): Boolean
 
     /**
      * All fields, including private supers.
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacTypeElement.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacTypeElement.kt
@@ -81,18 +81,16 @@ internal sealed class JavacTypeElement(
     override fun isExpect() = kotlinMetadata?.isExpect() == true
 
     override fun isAnnotationClass(): Boolean {
-        return (element.kind == ElementKind.ANNOTATION_TYPE) ||
-            kotlinMetadata?.isAnnotationClass() == true
+        return kotlinMetadata?.isAnnotationClass()
+            ?: (element.kind == ElementKind.ANNOTATION_TYPE)
     }
 
     override fun isClass(): Boolean {
-        return (element.kind == ElementKind.CLASS) ||
-            kotlinMetadata?.isClass() == true
+        return kotlinMetadata?.isClass() ?: (element.kind == ElementKind.CLASS)
     }
 
     override fun isInterface(): Boolean {
-        return (element.kind == ElementKind.INTERFACE) ||
-            kotlinMetadata?.isInterface() == true
+        return kotlinMetadata?.isInterface() ?: (element.kind == ElementKind.INTERFACE)
     }
 
     override fun findPrimaryConstructor(): JavacConstructorElement? {
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacTypeElement.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacTypeElement.kt
@@ -21,7 +21,6 @@ import androidx.room.compiler.processing.XFieldElement
 import androidx.room.compiler.processing.XHasModifiers
 import androidx.room.compiler.processing.XMethodElement
 import androidx.room.compiler.processing.XTypeElement
-import androidx.room.compiler.processing.javac.JavacTypeElement.JavacEnumTypeElement
 import androidx.room.compiler.processing.javac.kotlin.KotlinMetadataElement
 import com.google.auto.common.MoreElements
 import com.google.auto.common.MoreTypes
@@ -58,8 +57,6 @@ internal sealed class JavacTypeElement(
         element.enclosingType(env)
     }
 
-    override fun isInterface() = element.kind == ElementKind.INTERFACE
-
     private val _allFieldsIncludingPrivateSupers by lazy {
         element.getAllFieldsIncludingPrivateSupers(
             env.elementUtils
@@ -77,6 +74,26 @@ internal sealed class JavacTypeElement(
     }
 
     override fun isKotlinObject() = kotlinMetadata?.isObject() == true
+    override fun isCompanionObject() = kotlinMetadata?.isCompanionObject() == true
+    override fun isDataClass() = kotlinMetadata?.isDataClass() == true
+    override fun isValueClass() = kotlinMetadata?.isValueClass() == true
+    override fun isFunctionalInterface() = kotlinMetadata?.isFunctionalInterface() == true
+    override fun isExpect() = kotlinMetadata?.isExpect() == true
+
+    override fun isAnnotationClass(): Boolean {
+        return (element.kind == ElementKind.ANNOTATION_TYPE) ||
+            kotlinMetadata?.isAnnotationClass() == true
+    }
+
+    override fun isClass(): Boolean {
+        return (element.kind == ElementKind.CLASS) ||
+            kotlinMetadata?.isClass() == true
+    }
+
+    override fun isInterface(): Boolean {
+        return (element.kind == ElementKind.INTERFACE) ||
+            kotlinMetadata?.isInterface() == true
+    }
 
     override fun findPrimaryConstructor(): JavacConstructorElement? {
         val primarySignature = kotlinMetadata?.findPrimaryConstructorSignature() ?: return null
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinClassMetadataUtils.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinClassMetadataUtils.kt
@@ -193,7 +193,14 @@ private class ConstructorReader(val result: MutableList<KmConstructor>) : KmClas
 internal class KotlinMetadataClassFlags(val classMetadata: KotlinClassMetadata.Class) {
 
     private val flags: Flags by lazy {
-        classMetadata.toKmClass().flags
+        var theFlags: Flags = 0
+        classMetadata.accept(object : KmClassVisitor() {
+            override fun visit(flags: Flags, name: ClassName) {
+                theFlags = flags
+                super.visit(flags, name)
+            }
+        })
+        return@lazy theFlags
     }
 
     fun isObject(): Boolean = Flag.Class.IS_OBJECT(flags)
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinClassMetadataUtils.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinClassMetadataUtils.kt
@@ -39,6 +39,7 @@ import kotlinx.metadata.jvm.KotlinClassMetadata
 internal interface KmExecutable {
     val parameters: List<KmValueParameter>
 }
+
 /**
  * Represents the kotlin metadata of a function
  */
@@ -68,6 +69,7 @@ internal data class KmProperty(
 ) {
     val typeParameters
         get() = type.typeArguments
+
     fun isNullable() = Flag.Type.IS_NULLABLE(type.flags)
 }
 
@@ -188,23 +190,33 @@ private class ConstructorReader(val result: MutableList<KmConstructor>) : KmClas
     }
 }
 
-internal fun KotlinClassMetadata.Class.isObject(): Boolean = ObjectReader().let {
-    this@isObject.accept(it)
-    it.isObject
+internal class KotlinMetadataClassFlags(val classMetadata: KotlinClassMetadata.Class) {
+
+    private val flags: Flags by lazy {
+        classMetadata.toKmClass().flags
+    }
+
+    fun isObject(): Boolean = Flag.Class.IS_OBJECT(flags)
+
+    fun isCompanionObject(): Boolean = Flag.Class.IS_COMPANION_OBJECT(flags)
+
+    fun isAnnotationClass(): Boolean = Flag.Class.IS_ANNOTATION_CLASS(flags)
+
+    fun isInterface(): Boolean = Flag.Class.IS_INTERFACE(flags)
+
+    fun isClass(): Boolean = Flag.Class.IS_CLASS(flags)
+
+    fun isDataClass(): Boolean = Flag.Class.IS_DATA(flags)
+
+    fun isValueClass(): Boolean = Flag.Class.IS_INLINE(flags)
+
+    fun isFunctionalInterface(): Boolean = Flag.Class.IS_FUN(flags)
+
+    fun isExpect(): Boolean = Flag.Class.IS_EXPECT(flags)
 }
 
 internal fun KotlinClassMetadata.Class.readProperties(): List<KmProperty> =
     mutableListOf<KmProperty>().apply { accept(PropertyReader(this)) }
-
-/**
- * Reads whether the given class is a kotlin object
- */
-private class ObjectReader : KmClassVisitor() {
-    var isObject: Boolean = false
-    override fun visit(flags: Flags, name: ClassName) {
-        isObject = Flag.Class.IS_OBJECT(flags)
-    }
-}
 
 /**
  * Reads the properties of a class declaration
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinMetadataElement.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinMetadataElement.kt
@@ -45,6 +45,9 @@ internal class KotlinMetadataElement(
     private val functionList: List<KmFunction> by lazy { classMetadata.readFunctions() }
     private val constructorList: List<KmConstructor> by lazy { classMetadata.readConstructors() }
     private val propertyList: List<KmProperty> by lazy { classMetadata.readProperties() }
+    private val classFlags: KotlinMetadataClassFlags by lazy {
+        KotlinMetadataClassFlags(classMetadata)
+    }
 
     private val ExecutableElement.descriptor: String
         get() = descriptor()
@@ -53,7 +56,15 @@ internal class KotlinMetadataElement(
         it.isPrimary()
     }?.descriptor
 
-    fun isObject(): Boolean = classMetadata.isObject()
+    fun isObject(): Boolean = classFlags.isObject()
+    fun isCompanionObject(): Boolean = classFlags.isCompanionObject()
+    fun isAnnotationClass(): Boolean = classFlags.isAnnotationClass()
+    fun isClass(): Boolean = classFlags.isClass()
+    fun isInterface(): Boolean = classFlags.isInterface()
+    fun isDataClass(): Boolean = classFlags.isDataClass()
+    fun isValueClass(): Boolean = classFlags.isValueClass()
+    fun isFunctionalInterface(): Boolean = classFlags.isFunctionalInterface()
+    fun isExpect(): Boolean = classFlags.isExpect()
 
     fun getFunctionMetadata(method: ExecutableElement): KmFunction? {
         check(method.kind == ElementKind.METHOD) {
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspElement.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspElement.kt
@@ -61,4 +61,10 @@ internal abstract class KspElement(
             KSFileAsOriginatingElement(it)
         }
     }
+
+    override val docComment: String? by lazy {
+        // TODO: Not yet implemented in KSP.
+        // https://github.com/google/ksp/issues/392
+        null
+    }
 }
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspTypeElement.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspTypeElement.kt
@@ -40,6 +40,7 @@ import com.google.devtools.ksp.symbol.KSPropertyDeclaration
 import com.google.devtools.ksp.symbol.Modifier
 import com.google.devtools.ksp.symbol.Origin
 import com.squareup.javapoet.ClassName
+import javax.tools.Diagnostic
 
 internal sealed class KspTypeElement(
     env: KspProcessingEnv,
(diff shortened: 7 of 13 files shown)


room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/ProcessorTestExt.kt
```diff
@@ -173,7 +175,8 @@ fun runProcessorTest(
     sources: List<Source> = emptyList(),
     classpath: List<File> = emptyList(),
     options: Map<String, String> = emptyMap(),
-    handlers: List<(XTestInvocation) -> Unit>
+    handlers: List<(XTestInvocation) -> Unit>,
+    kotlincArguments: List<String> = emptyList()
```
yigit: small nit. lets keep handlers the last argument to be consistent with the other ones that receive a handler. (even though you still need to pass this explicitly)

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/ProcessorTestExt.kt
```diff
@@ -115,12 +115,14 @@ fun runProcessorTest(
     sources: List<Source> = emptyList(),
     classpath: List<File> = emptyList(),
     options: Map<String, String> = emptyMap(),
+    kotlincArguments: List<String> = emptyList(),
```
yigit: i'm not sure if we want to have a kotlin specific parameter here. (haven't seen the rest of the PR yet). But maybe, we can make this an `argumentProvider : (CompilationType) -> List<String>` ? (just an idea)

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
```diff
@@ -65,10 +65,48 @@ interface XTypeElement : XHasModifiers, XElement {
...
+     * Returns `true` if this [XTypeElement] represents a Kotlin functional interface,
+     * i.e. marked with the keyword `fun`
+     */
+    fun isFunctionalInterface(): Boolean
+
+    /**
+     * Returns `true` if this [XTypeElement] represents an ordinary class. ie not an enum, object,
+     * annotation, interface, or other type of specialty class.
+     */
+    fun isNormalClass(): Boolean
```
yigit: is this the right word for this? I'm not sure but if there is a better word than Normal in the literature, it might be better.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
```diff
@@ -65,10 +65,48 @@ interface XTypeElement : XHasModifiers, XElement {
...
+
+    /**
+     * Returns `true` if this [XTypeElement] represents a Kotlin inline class
+     */
+    fun isInlineClass(): Boolean
+
+    /**
+     * Returns `true` if this [XTypeElement] represents a class with the Kotlin `expect` keyword
+     */
+    fun isExpect(): Boolean
```
yigit, line 92: do we need `isActual` ?

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacTypeElement.kt
```diff
@@ -77,6 +77,25 @@ internal sealed class JavacTypeElement(
...
     override fun isKotlinObject() = kotlinMetadata?.isObject() == true
+    override fun isCompanionObject() = kotlinMetadata?.isCompanionObject() == true
+    override fun isDataClass() = kotlinMetadata?.isDataClass() == true
+    override fun isInlineClass() = kotlinMetadata?.isInlineClass() == true
+    override fun isFunctionalInterface() = kotlinMetadata?.isFunctionalInterface() == true
+    override fun isExpect() = kotlinMetadata?.isExpect() == true
+
+    override fun isAnnotationClass(): Boolean {
+        return kotlinMetadata?.isAnnotationClass()
+            ?: run {
```
yigit: i guess you can omit this run?

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacTypeElement.kt
```diff
@@ -77,6 +77,25 @@ internal sealed class JavacTypeElement(
...
+                element.kind == ElementKind.ANNOTATION_TYPE
+            }
+    }
+
+    override fun isNormalClass(): Boolean {
+        return kotlinMetadata?.isNormalClass()
+            ?: run {
```
yigit: ditto

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
yigit: also, maybe we can add `isInterface` for completeness and then use `isInterface` vs `isClass` ?

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinMetadataElement.kt
```diff
@@ -54,6 +54,13 @@ internal class KotlinMetadataElement(
     }?.descriptor
 
     fun isObject(): Boolean = classMetadata.isObject()
+    fun isCompanionObject(): Boolean = classMetadata.isCompanionObject()
```
yigit: seems like we'll read these a lot, would be good to cache the value (can be a followup CL)

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspElement.kt
```diff
@@ -61,4 +61,10 @@ internal abstract class KspElement(
             KSFileAsOriginatingElement(it)
         }
     }
+
+    override val docComment: String? by lazy {
+        // Not yet implemented in KSP.
```
yigit: can you add a TODO to the beginning of this line.

also, thanks for linking the issue, i have a small bash script that parses ksp links so that we know what we are waiting for from KSP.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspTypeElement.kt
```diff
@@ -251,6 +251,38 @@ internal sealed class KspTypeElement(
...
+        return Modifier.DATA in declaration.modifiers
+    }
+
+    override fun isInlineClass(): Boolean {
+        return Modifier.INLINE in declaration.modifiers
+    }
+
+    override fun isFunctionalInterface(): Boolean {
+        // Update this once KSP supports it https://github.com/google/ksp/issues/393
+        throw UnsupportedOperationException(
```
yigit: i'm a bit conflcited on throwing because it will break a processor just by calling this even when the answer would be false. 
on the other hand returning false without checking is not great.

maybe log an error an System.error and return false instead of crashing?

room/compiler-processing/src/test/java/androidx/room/compiler/processing/XTypeElementTest.kt
```diff
@@ -224,8 +237,16 @@ class XTypeElementTest {
                 if (element.isProtected()) result.add("protected")
                 if (element.isPublic()) result.add("public")
                 if (element.isKotlinObject()) result.add("object")
+                if (element.isCompanionObject()) result.add("companion")
+                // Enable once ksp supports fun interfaces - https://github.com/google/ksp/issues/393
+                // if (element.isFunctionalInterface()) result.add("fun")
```
yigit: what i usually do for these is to still call it (if we don't throw above, we can) then in the assertions, check for `invocation.isKsp` to exclude the check from KSP (so that the javac path is tested)

room/compiler-processing/src/test/java/androidx/room/compiler/processing/XTypeElementTest.kt
```diff
@@ -235,32 +256,50 @@ class XTypeElementTest {
...
+                } else {
+                    containsExactly("abstract", "public", "annotation")
+                }
+            }
+            assertThat(getModifiers("DataClass"))
+                .containsExactly("public", "final", "normal", "data")
+            assertThat(getModifiers("InlineClass"))
+                .containsExactly("public", "final", "normal", "inline")
+            // Enable once ksp supports fun interfaces - https://github.com/google/ksp/issues/393
+//            assertThat(getModifiers("FunInterface"))
```
yigit: we can wrap this in `!invocation.isKsp`

yigit (commented):
overall lgtm, thanks!
just small nits.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
elihart: it isn't present in the metadata for some reason - maybe because kapt was only usable for JVM?

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
elihart: Kotlin API's just refer to it as a "class", although the kdoc clarifies saying it is just a "usual" class. I can just call it a class and add the interface check too, I'll be clear in the kdoc

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinMetadataElement.kt
elihart: Sure, I can just do that now

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/ProcessorTestExt.kt
elihart: sure. I didn't end up using this since I couldn't get multi-platform testing to work, but might as well keep it in if you think it's useful (otherwise I can remove it)

room/compiler-processing/src/test/java/androidx/room/compiler/processing/XTypeElementTest.kt
elihart: Good call

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt
```diff
@@ -37,7 +37,7 @@ abstract class XMessager {
         onPrintMessage(kind, msg, element)
     }
 
-    abstract fun onPrintMessage(kind: Diagnostic.Kind, msg: String, element: XElement? = null)
+    protected abstract fun onPrintMessage(kind: Diagnostic.Kind, msg: String, element: XElement? = null)
```
elihart: I realized this would pop up in auto complete when trying to print a message, which is confusing and seems error prone. I believe it's use case is only internal

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
```diff
@@ -65,10 +65,48 @@ interface XTypeElement : XHasModifiers, XElement {
...
+
+    /**
+     * Returns `true` if this [XTypeElement] represents a Kotlin data class
+     */
+    fun isDataClass(): Boolean
+
+    /**
+     * Returns `true` if this [XTypeElement] represents a Kotlin value class
+     */
+    fun isValueClass(): Boolean
```
elihart, line 87: Renamed this from inline to value - figure I might as well get ahead of the imminent 1.5.0 release/rename

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinClassMetadataUtils.kt
```diff
@@ -188,24 +190,34 @@ private class ConstructorReader(val result: MutableList<KmConstructor>) : KmClas
     }
 }
 
-internal fun KotlinClassMetadata.Class.isObject(): Boolean = ObjectReader().let {
-    this@isObject.accept(it)
-    it.isObject
+internal class KotlinMetadataClassFlags(val classMetadata: KotlinClassMetadata.Class) {
+
+    private val flags: Flags by lazy {
```
elihart, line 195: Took a different approach with the flag access and caching here, lmk what you think

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspTypeElement.kt
```diff
@@ -251,6 +252,40 @@ internal sealed class KspTypeElement(
...
+    }
+
+    override fun isValueClass(): Boolean {
+        return Modifier.INLINE in declaration.modifiers
+    }
+
+    override fun isFunctionalInterface(): Boolean {
+        // TODO: Update this once KSP supports it
+        // https://github.com/google/ksp/issues/393
+        env.messager.printMessage(
```
elihart, line 280: is this the kind of logging you meant?

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/ProcessorTestExt.kt
```diff
@@ -115,12 +115,14 @@ fun runProcessorTest(
...
     options: Map<String, String> = emptyMap(),
+    compilerArguments: (CompilationType) -> List<String> = { emptyList() },
     handler: (XTestInvocation) -> Unit
 ) = runProcessorTest(
     sources = sources,
     classpath = classpath,
     options = options,
-    handlers = listOf(handler)
+    handlers = listOf(handler),
+    compilerArguments = compilerArguments
```
yigit: nit: lets keep handlers the last argument

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/ProcessorTestExt.kt
```diff
@@ -173,7 +175,8 @@ fun runProcessorTest(
     sources: List<Source> = emptyList(),
     classpath: List<File> = emptyList(),
     options: Map<String, String> = emptyMap(),
-    handlers: List<(XTestInvocation) -> Unit>
+    compilerArguments: (CompilationType) -> List<String> = { emptyList() },
+    handlers: List<(XTestInvocation) -> Unit>,
```
yigit: ditto

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/ProcessorTestExt.kt
yigit: well not fully sure actually because it is the only function type parameter here.
though i'm thinking, maybe we should use an explicit type here, or fun interface? specifically, for other overloads that receive a handler block, idk how well IJ will resolve it.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt
yigit: probably my bad, thanks.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
yigit: probably.
KSP has one: https://github.com/google/ksp/blob/master/api/src/main/kotlin/com/google/devtools/ksp/symbol/KSExpectActual.kt#L30

but we can add it later as well.

yigit: also, in KSP all KSDeclaration's extend that interface so expect/actual might be better in XElement but again, we can refactor that later.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacTypeElement.kt
```diff
@@ -77,6 +74,23 @@ internal sealed class JavacTypeElement(
...
 
     override fun isKotlinObject() = kotlinMetadata?.isObject() == true
+    override fun isCompanionObject() = kotlinMetadata?.isCompanionObject() == true
+    override fun isDataClass() = kotlinMetadata?.isDataClass() == true
+    override fun isValueClass() = kotlinMetadata?.isValueClass() == true
+    override fun isFunctionalInterface() = kotlinMetadata?.isFunctionalInterface() == true
+    override fun isExpect() = kotlinMetadata?.isExpect() == true
+
+    override fun isAnnotationClass(): Boolean {
+        return kotlinMetadata?.isAnnotationClass() ?: (element.kind == ElementKind.ANNOTATION_TYPE)
```
yigit: minor nit, probably better to do the elements check first but feel free to keep it this way.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinClassMetadataUtils.kt
yigit: lgtm. cc: @danysantiago

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspTypeElement.kt
yigit: lgtm. it will end up in logs in the user's compilation but then it would bubble back to the developer. for now, better be safe & verbose.

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/ProcessorTestExt.kt
yigit: also, i think this is not used in this PR? do you need it on your end?

yigit (approved):

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/CompilationType.kt
```diff
@@ -0,0 +1,21 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.room.compiler.processing.util
+
+enum class CompilationType {
```
yigit: if we keep this, we should add some docs here.

yigit:
this is the failure in AOSP:

```
Detected these failing tasks: [':room:room-compiler-processing:ktlint']
> Task :room:room-compiler-processing:ktlint
$SUPPORT/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt:40:1: Exceeded max line length (100) (cannot be auto-corrected)
$SUPPORT/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt:40:43: Parameter should be on a separate line (unless all parameters can fit a single line)
$SUPPORT/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt:40:66: Parameter should be on a separate line (unless all parameters can fit a single line)
$SUPPORT/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt:40:79: Parameter should be on a separate line (unless all parameters can fit a single line)
$SUPPORT/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt:40:104: Missing newline before ")"

> Task :room:room-compiler-processing:ktlint FAILED
$SUPPORT/room/compiler-processing/src/test/java/androidx/room/compiler/processing/XTypeElementTest.kt:306:1: Unexpected blank line(s) before "}"
"checkstyle" report written to $OUT_DIR/androidx/room/room-compiler-processing/build/reports/ktlint/ktlint-checkstyle-report.xml

```

yigit:
https://android-review.googlesource.com/c/platform/frameworks/support/+/1690663

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/ProcessorTestExt.kt
elihart: I don't actually need this anymore, I guess I will just delete it unless you think it would be useful to keep. I added it because I thought it might be necessary to enable experiment flags like inline classes or to enable multiplatform, but that didn't pan out in this PR

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacTypeElement.kt
elihart: I can switch it, seems faster

yigit (approved):

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinClassMetadataUtils.kt
```diff
@@ -188,24 +190,34 @@ private class ConstructorReader(val result: MutableList<KmConstructor>) : KmClas
...
+        classMetadata.toKmClass().flags
+    }
+
+    fun isObject(): Boolean = Flag.Class.IS_OBJECT(flags)
```
danysantiago, line 206: nit: you can omit Boolean return type for all of these flag checks

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinClassMetadataUtils.kt
danysantiago: Since these are just class level flags I would avoid using `toKmClass()` since that parses *everything* (visits all properties and functions), instead you can get ahold of the flags with a small visitor:
```
private val flags: Flags by lazy {
    var theFlags: Flags = 0
    classMetadata.accept(object : KmClassVisitor() {
        override fun visit(flags: Flags, name: ClassName) {
            theFlags = flags
            super.visit(flags, name)
        }
    })
    return@lazy theFlags
}
```

yigit:
presubmit fails with:

```
Detected these failing tasks: [':room:room-compiler-processing:compileKotlin']
> Task :room:room-compiler-processing:compileKotlin FAILED
e: $SUPPORT/room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspFileMemberContainer.kt: (28, 10): Class 'KspFileMemberContainer' is not abstract and does not implement abstract member public abstract val docComment: String? defined in androidx.room.compiler.processing.ksp.KspMemberContainer
```

This is due to my recent change to support top level members :/.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/kotlin/KotlinClassMetadataUtils.kt
elihart: good tip, thanks

elihart:
I've addressed everything, thanks!

danysantiago (approved):

yigit:
this is the failure from aosp, might need a rebase, not sure why it passed here but failing there.

```
androidx.room.compiler.processing.util.CompilationResultSubject$createProcessorAssertionError$1: processor did throw an error
androidx.room.compiler.processing.util.CompilationResultSubject$createProcessorAssertionError$1: processor did throw an error
CompilationResult (with kapt)
ERROR: 0

WARNING: 0

MANDATORY_WARNING: 0

NOTE: 0

OTHER: 0

Generated files:
RAW OUTPUT:
w: /tmp/Kotlin-Compilation2496147856791129456/ksp/srcInput/kotlin/Foo.kt: (13, 1): The feature "inline classes" is experimental


Caused by: unexpected (1): class
---
expected      : [final, public, object]
but was       : [final, public, object, class]
	at androidx.room.compiler.processing.XTypeElementTest$modifiers$1.invoke(XTypeElementTest.kt:262)
	at androidx.room.compiler.processing.XTypeElementTest$modifiers$1.invoke(XTypeElementTest.kt:38)
	at androidx.room.compiler.processing.SyntheticProcessorImpl.runNextRound(SyntheticProcessor.kt:106)
	at androidx.room.compiler.processing.SyntheticJavacProcessor.doProcess(SyntheticJavacProcessor.kt:39)
	at androidx.room.compiler.processing.JavacTestProcessor.process(JavacTestProcessor.kt:51)
	at org.jetbrains.kotlin.kapt3.base.incremental.IncrementalProcessor.process(incrementalProcessors.kt:89)
	at org.jetbrains.kotlin.kapt3.base.ProcessorWrapper.process(annotationProcessing.kt:166)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.callProcessor(JavacProcessingEnvironment.java:980)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.discoverAndRunProcs(JavacProcessingEnvironment.java:896)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment$Round.run(JavacProcessingEnvironment.java:1222)
	at jdk.compiler/com.sun.tools.javac.processing.JavacProcessingEnvironment.doProcessing(JavacProcessingEnvironment.java:1334)
	at jdk.compiler/com.sun.tools.javac.main.JavaCompiler.processAnnotations(JavaCompiler.java:1258)
	at jdk.internal.reflect.GeneratedMethodAccessor118.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.jetbrains.kotlin.kapt3.base.AnnotationProcessingKt.doAnnotationProcessing(annotationProcessing.kt:81)
	at org.jetbrains.kotlin.kapt3.base.AnnotationProcessingKt.doAnnotationProcessing$default(annotationProcessing.kt:31)
	at org.jetbrains.kotlin.kapt3.AbstractKapt3Extension.runAnnotationProcessing(Kapt3Extension.kt:229)
	at org.jetbrains.kotlin.kapt3.AbstractKapt3Extension.analysisCompleted(Kapt3Extension.kt:192)
	at org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM.analyzeFilesWithJavaIntegration$invokeExtensionsOnAnalysisComplete(TopDownAnalyzerFacadeForJVM.kt:108)
	at org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM.analyzeFilesWithJavaIntegration(TopDownAnalyzerFacadeForJVM.kt:118)
	at org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM.analyzeFilesWithJavaIntegration$default(TopDownAnalyzerFacadeForJVM.kt:85)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinToJVMBytecodeCompiler$analyze$1.invoke(KotlinToJVMBytecodeCompiler.kt:517)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinToJVMBytecodeCompiler$analyze$1.invoke(KotlinToJVMBytecodeCompiler.kt:508)
	at org.jetbrains.kotlin.cli.common.messages.AnalyzerWithCompilerReport.analyzeAndReport(AnalyzerWithCompilerReport.kt:114)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinToJVMBytecodeCompiler.analyze(KotlinToJVMBytecodeCompiler.kt:508)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinToJVMBytecodeCompiler.compileModules$cli(KotlinToJVMBytecodeCompiler.kt:188)
	at org.jetbrains.kotlin.cli.jvm.compiler.KotlinToJVMBytecodeCompiler.compileModules$cli$default(KotlinToJVMBytecodeCompiler.kt:154)
	at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecute(K2JVMCompiler.kt:169)
	at org.jetbrains.kotlin.cli.jvm.K2JVMCompiler.doExecute(K2JVMCompiler.kt:52)
	at org.jetbrains.kotlin.cli.common.CLICompiler.execImpl(CLICompiler.kt:88)
	at org.jetbrains.kotlin.cli.common.CLICompiler.execImpl(CLICompiler.kt:44)
	at org.jetbrains.kotlin.cli.common.CLITool.exec(CLITool.kt:98)
	at com.tschuchort.compiletesting.KotlinCompilation.stubsAndApt(KotlinCompilation.kt:455)
	at com.tschuchort.compiletesting.KotlinCompilation.compile(KotlinCompilation.kt:632)
	at androidx.room.compiler.processing.util.runner.KaptCompilationTestRunner.compile(KaptCompilationTestRunner.kt:46)
	at androidx.room.compiler.processing.util.ProcessorTestExtKt.runTests(ProcessorTestExt.kt:44)
	at androidx.room.compiler.processing.util.ProcessorTestExtKt.runProcessorTest(ProcessorTestExt.kt:183)
	at androidx.room.compiler.processing.util.ProcessorTestExtKt.runProcessorTest(ProcessorTestExt.kt:119)
	at androidx.room.compiler.processing.util.ProcessorTestExtKt.runProcessorTest$default(ProcessorTestExt.kt:117)
	at androidx.room.compiler.processing.XTypeElementTest.modifiers(XTypeElementTest.kt:229)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	```

elihart:
Thanks, will try to get this addressed today

elihart:
Test should be fixed now - I needed to revert the ordering change I made to use kotlin metadata first since that correctly differentiates between object and class. Interface and annotation types are probably fine to check java kind first, but it seems safer to prefer the kotlin information for consistency with ksp.

PR #165, opened 2021-04-19
[Room][Compiler Processing] Adds configuration to define processor target language

gfreivasc:
## Proposed Changes

Adds configuration that defines if a processing environment should
expect to work as if on a regular Java setup or if it's Kotlin only.
This defines internal behavior as to how some things will be modelled,
for example if we can expect top level methods to be modelled as such,
or if we should synthesize a `Kt` type to hold it instead.

For JavaAP, the `XProcessingStep.asAutoCommonProcessor()` method now
expects the language to be specified. If not specified, it will default
to expect Java language. Same is applied for KSP.

## Testing
Test:

    ./gradlew \
        test connectedCheck \
        -x :room:room-benchmark:cC \

## Issues Fixed

Fixes: [b/185760236](https://issuetracker.google.com/issues/185760236)

### DIFF
--- a/room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/SyntheticKspProcessor.kt
+++ b/room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/SyntheticKspProcessor.kt
@@ -56,7 +56,8 @@ class SyntheticKspProcessor private constructor(
             options,
             resolver,
             codeGenerator,
-            logger
+            logger,
+            XProcessingEnv.Language.JAVA
         )
         val testInvocation = XTestInvocation(
             processingEnv = xEnv,
--- a/room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/SyntheticKspProcessor.kt
+++ b/room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/SyntheticKspProcessor.kt
@@ -25,10 +25,15 @@ import com.google.devtools.ksp.symbol.KSAnnotated
 
 @ExperimentalProcessingApi
 class SyntheticKspProcessor private constructor(
-    private val impl: SyntheticProcessorImpl
+    private val impl: SyntheticProcessorImpl,
+    private val targetLanguage: XProcessingEnv.Language
 ) : SymbolProcessor, SyntheticProcessor by impl {
-    constructor(handlers: List<(XTestInvocation) -> Unit>) : this(
-        SyntheticProcessorImpl(handlers)
+    constructor(
+        handlers: List<(XTestInvocation) -> Unit>,
+        targetLanguage: XProcessingEnv.Language
+    ) : this(
+        SyntheticProcessorImpl(handlers),
+        targetLanguage
     )
 
     private lateinit var options: Map<String, String>
@@ -57,7 +62,7 @@ class SyntheticKspProcessor private constructor(
             resolver,
             codeGenerator,
             logger,
-            XProcessingEnv.Language.JAVA
+            targetLanguage
         )
         val testInvocation = XTestInvocation(
             processingEnv = xEnv,
--- a/room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/runner/JavacCompilationTestRunner.kt
+++ b/room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/runner/JavacCompilationTestRunner.kt
@@ -33,7 +33,10 @@ internal object JavacCompilationTestRunner : CompilationTestRunner {
     }
 
     override fun compile(params: TestCompilationParameters): CompilationResult {
-        val syntheticJavacProcessor = SyntheticJavacProcessor(params.handlers)
+        val syntheticJavacProcessor = SyntheticJavacProcessor(
+            params.handlers,
+            params.targetLanguage
+        )
         val sources = if (params.sources.isEmpty()) {
             // synthesize a source to trigger compilation
             listOf(
--- a/room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/runner/JavacCompilationTestRunner.kt
+++ b/room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/runner/JavacCompilationTestRunner.kt
@@ -18,10 +18,12 @@ package androidx.room.compiler.processing.util.runner
 
 import androidx.room.compiler.processing.ExperimentalProcessingApi
 import androidx.room.compiler.processing.SyntheticJavacProcessor
+import androidx.room.compiler.processing.XProcessingEnv
 import androidx.room.compiler.processing.util.CompilationResult
 import androidx.room.compiler.processing.util.JavaCompileTestingCompilationResult
 import androidx.room.compiler.processing.util.Source
 import com.google.testing.compile.Compiler
+import org.jetbrains.kotlin.types.model.requireOrDescribe
 
 @ExperimentalProcessingApi
 internal object JavacCompilationTestRunner : CompilationTestRunner {
@@ -33,9 +35,12 @@ internal object JavacCompilationTestRunner : CompilationTestRunner {
     }
 
     override fun compile(params: TestCompilationParameters): CompilationResult {
+        require(params.targetLanguage != XProcessingEnv.Language.KOTLIN) {
+            "Javac processor is not meant to process Kotlin code. Use KAPT instead."
+        }
         val syntheticJavacProcessor = SyntheticJavacProcessor(
             params.handlers,
-            params.targetLanguage
+            XProcessingEnv.Language.JAVA
         )
         val sources = if (params.sources.isEmpty()) {
             // synthesize a source to trigger compilation
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
@@ -35,7 +35,7 @@ interface XProcessingEnv {
     val backend: Backend
 
     /**
-     * The target language the processor is set to expect
+     * The expected language of the sources to be processed and generated.
      */
     val targetLanguage: Language
 
--- a/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
+++ b/room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
@@ -134,6 +134,11 @@ interface XProcessingEnv {
         KSP
     }
 
+    /**
+     * The target language for XProcessingEnv.
+     * This value controls certain logic when processing code,
+     * see the documentation for [XProcessingEnv.targetLanguage] for details.
+     */
     enum class Language {
         JAVA,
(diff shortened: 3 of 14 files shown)


yigit:
for the lint failure, you need 

`Test:`

`Bug:` or `Fixes:`

stanzas (we should probably update the template)

Btw, overall lgtm

gfreivasc:
I've updated the description after it failed the test but I would need to trigger the pipeline again

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
```diff
@@ -128,11 +134,19 @@ interface XProcessingEnv {
         KSP
     }
 
+    enum class Language {
+        JAVA,
+        KOTLIN
+    }
+
```
yigit, line 146: sorry i forgot to put comments here but it would be good to add comments here.

```suggestion
    /**
     * The target language for XProcessingEnv.
     * This value controls certain logic when processing code, see the documentation for [XProcessingEnv.targetLanguage] for details.
     */
    enum class Language {
        JAVA,
        KOTLIN
    }

```

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
```diff
@@ -33,6 +33,12 @@ import kotlin.reflect.KClass
 interface XProcessingEnv {
 
     val backend: Backend
+
+    /**
+     * The target language the processor is set to expect
+     */
```
yigit, line 39: we don't have anything here yet but i think this is probably the best place to document behavior changes between the two. As we add them, we should start documenting them here (open for better suggestions though )

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
```diff
@@ -85,7 +88,7 @@ internal class JavacProcessingStepDelegate(
         val converted = mutableMapOf<String, Set<XElement>>()
         // create a new x processing environment for each step to ensure it can freely cache
         // whatever it wants and we don't keep elements references across rounds.
-        val xEnv = JavacProcessingEnv(env)
+        val xEnv = JavacProcessingEnv(env, targetLanguage)
```
yigit, line 96: i'm not sure if we can support kotlin here so we might consider eventually throwing Unsupported exception from the constructor of this but fine for this PR until we clarify what to do there.

yigit (approved):
approving to kick of aosp (not all checks run here yet)

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
gfreivasc: The processor doesn´t work with multiple rounds if it generates kotlin code, but I thought this API was fine.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
gfreivasc: Indeed, done.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingStep.kt
yigit: i was more thinking in the lines of "can we fake kotlin structure from java processor". e.g. we would need to detect things like "this class is actually generated from a KSFile" etc, which i thought would be hard.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
```diff
@@ -141,12 +160,14 @@ interface XProcessingEnv {
             options: Map<String, String>,
             resolver: Resolver,
             codeGenerator: CodeGenerator,
-            logger: KSPLogger
+            logger: KSPLogger,
+            targetLanguage: Language = Language.KOTLIN
```
yigit: we should have the same default for all and since it works like Java now, this should always be JAVA.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspProcessingEnv.kt
```diff
@@ -38,7 +38,8 @@ internal class KspProcessingEnv(
     override val options: Map<String, String>,
     codeGenerator: CodeGenerator,
     logger: KSPLogger,
-    val resolver: Resolver
+    val resolver: Resolver,
+    override val targetLanguage: XProcessingEnv.Language = XProcessingEnv.Language.KOTLIN
```
yigit: ditto. this should default to java.

though i'm wondering, given that we require these in the builders, can we remove the default here?

yigit (changes requested):

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
gfreivasc: Yeah, that´s true, changed that part.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspProcessingEnv.kt
gfreivasc: We could, but I was wondering if this could break client code. Will see if it breaks compilation or tests. If it passes I´ll remove it.

gfreivasc: [`SyntheticKspProcessor`](https://github.com/androidx/androidx/blob/788e5b8a5af3ca6102837b93ed272c60d4945f27/room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/SyntheticKspProcessor.kt#L54) seems to make use of that factory. Should I explicitly pass Java to it?

yigit: yep. at least for now until we add this language parameter to the `runProcessorTest` function.

yigit: btw you can also implement `runProcessorTest` change here but can be a followup too. whatever you prefer.

gfreivasc: Will take a look into it

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/SyntheticKspProcessor.kt
```diff
@@ -55,7 +55,8 @@ class SyntheticKspProcessor private constructor(
             options,
             resolver,
             codeGenerator,
-            logger
+            logger,
+            XProcessingEnv.Language.JAVA
```
yigit: we should probably pass this from somewhere into here instead of hardcoding. 

though after looking at the rest of the CL, you've not plugged it yet (to runProcessorTest). Were you planning to do that as a followup? It is probably better together so that there is a way to invoke it and then we can also write a runProcessorTest to ensure it is properly passed around.

yigit (changes requested):

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/SyntheticKspProcessor.kt
gfreivasc: I was thinking of doing it as a follow up, indeed. Will be doing it in this PR.

gfreivasc:
@yigit I've modified `runProcessorTest` and similar methods to take the target language as a parameter, thus removing the hardcoded value from the synthetic test processor which now takes that information from the constructor instead. Let me know if there are any further changes you think are applicable to these APIs.

Also, I've changed the PR description to be more in line with the fact that Java is the default target language regardless of whether it's running JavaAP or KSP.

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/runner/JavacCompilationTestRunner.kt
```diff
@@ -33,7 +33,10 @@ internal object JavacCompilationTestRunner : CompilationTestRunner {
     }
 
     override fun compile(params: TestCompilationParameters): CompilationResult {
-        val syntheticJavacProcessor = SyntheticJavacProcessor(params.handlers)
+        val syntheticJavacProcessor = SyntheticJavacProcessor(
+            params.handlers,
+            params.targetLanguage
+        )
```
gfreivasc: I wonder if it wouldn't be some kind of invalid scenario to have Javac targeting kotlin. The `runJavaProcessorTest()` is not supposed to run with kotlin source for what I can tell.

gfreivasc:
@yigit are these changes still relevant/applicable or should we close this PR?

yigit:
@gfreivasc sorry i lost track of this.
The changes a still relevant, last time i checked, i had some comments for change but then missed your comments....

Sorry about that, will re-review.

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/runner/JavacCompilationTestRunner.kt
yigit: yea this never runs with kotlin sources (in that case, we use kapt instead of javac).
We can make the processing env throw if language is not set to java though that would make testing a bit harder here :/ i'm fine to leave it as is for now.

yigit (approved):

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
```diff
@@ -33,6 +33,12 @@ import kotlin.reflect.KClass
 interface XProcessingEnv {
 
     val backend: Backend
+
+    /**
+     * The target language the processor is set to expect
```
danysantiago: nit: Reword a bit to clarify: `The expected language of the sources generated by the processor.`

room/compiler-processing-testing/src/main/java/androidx/room/compiler/processing/util/runner/JavacCompilationTestRunner.kt
gfreivasc: I think I have an idea of how to deal with it here. Will be changing that along with @danysantiago's suggestion.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XProcessingEnv.kt
gfreivasc: Changed the description

gfreivasc:
@yigit I made changes to make `JavacCompilationtestRunner` throw if it receives KOTLIN as a target Language, and also changed the code that calls it to avoid passing Kotlin down as a parameter. There are tests to assert these behaviors.

PR #166, opened 2021-04-20
Pull FTL results back into DIST_DIR

yigit:
This CL adds support for specifying a Bucket to keep results, which
defaults to a generic bucket I've created for FTL tests.

By default, the bucket path is inferred from github environment
variables when available, otherwise, a random UUID is chosen for uniqueness.

FTL outputs a bunch of things into this bucket, including a video of the
test running. All of it is pulled back into the out folder after the
test runs. I've also configured a finalizer task to copy that output
into the DIST folder.

Since we only rely on APK contents, I marked the task cacheable.

Bug: n/a
Test: production :(. verified build cache locally but required --build-cache

### DIFF
--- a/buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
+++ b/buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
@@ -186,7 +186,7 @@ internal class GCloudCLIWrapper(
         /**
          * The GS Bucket directory where the results will be saved
          */
-        val resultsBucketDir: String = makeResultsDir(projectPath),
+        val resultsBucketDir: String = buildRelativeResultDirPath(projectPath),
         /**
          * The local folder where we will download the test results
          */
@@ -209,13 +209,13 @@ internal class GCloudCLIWrapper(
             const val TEST_OUTPUT_FILE_NAME = "testResults.json"
 
             /**
-             * Generates a folder for test results.
+             * Generates a relative path for test results.
              *
              * If run on Github Actions CI, this method will use the environment variables to
              * create a unique path for the action.
              * If run locally, this will create a random UUID for the folder.
              */
-            fun makeResultsDir(
+            private fun buildRelativeResultDirPath(
                 projectPath: String
             ): String {
                 // github action env variables:
--- a/buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
+++ b/buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
@@ -136,11 +136,11 @@ internal class GCloudCLIWrapper(
         if (result.exitValue != 0) {
             throw GradleException(
                 """
-                    Unable to find $name CLI executable.
-                    `which $name` returned exit code ${result.exitValue}.
-                    Make sure gcloud CLI is installed, authenticated and is part of your PATH.
-                    See https://cloud.google.com/sdk/gcloud for installation instructions.
-                    """.trimIndent()
+                Unable to find $name CLI executable.
+                `which $name` returned exit code ${result.exitValue}.
+                Make sure gcloud CLI is installed, authenticated and is part of your PATH.
+                See https://cloud.google.com/sdk/gcloud for installation instructions.
+                """.trimIndent()
             )
         }
         return output.toString(Charsets.UTF_8).trim()
--- a/buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
+++ b/buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
@@ -16,6 +16,7 @@
 
 package androidx.build.ftl
 
+import androidx.build.ftl.GCloudCLIWrapper.RunTestParameters.Companion.TEST_OUTPUT_FILE_NAME
 import com.google.gson.Gson
 import com.google.gson.annotations.SerializedName
 import com.google.gson.reflect.TypeToken
@@ -24,6 +25,7 @@ import org.gradle.process.ExecOperations
 import java.io.ByteArrayOutputStream
 import java.io.File
 import java.util.Locale
+import java.util.UUID
 
 /**
  * Wrapper around GCloud CLI.
@@ -44,7 +46,7 @@ internal class GCloudCLIWrapper(
     /**
      * Path to the gcloud executable, derived from `which gcloud` call.
      */
-    private val executable: String by lazy {
+    private val gcloud: String by lazy {
         val output = ByteArrayOutputStream()
         val result = execOperations.exec {
             it.commandLine("which", "gcloud")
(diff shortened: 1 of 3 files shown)


tikurahul (approved):

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
```diff
@@ -44,53 +46,103 @@ internal class GCloudCLIWrapper(
...
+            it.commandLine("which", name)
+            it.standardOutput = output
+            it.isIgnoreExitValue = true
+        }
+        if (result.exitValue != 0) {
+            throw GradleException(
+                """
+                Unable to find $name CLI executable.
+                `which $name` returned exit code ${result.exitValue}.
+                Make sure gcloud CLI is installed, authenticated and is part of your PATH.
```
liutikas, line 139: Does this include gsutils?

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
```diff
@@ -110,6 +162,95 @@ internal class GCloudCLIWrapper(
...
+            const val TEST_OUTPUT_FILE_NAME = "testResults.json"
+
+            /**
+             * Generates a folder for test results.
+             *
+             * If run on Github Actions CI, this method will use the environment variables to
+             * create a unique path for the action.
+             * If run locally, this will create a random UUID for the folder.
+             */
+            fun makeResultsDir(
```
liutikas: This does not "make" a directory, so maybe call it something like getResultsDirectory or generateResultsDirectoryName

liutikas: Mark private. Additionally, should these just be private top level functions in this file?

buildSrc/src/main/kotlin/androidx/build/ftl/RunTestOnFTLTask.kt
```diff
@@ -153,7 +170,16 @@ abstract class RunTestOnFTLTask @Inject constructor(
                         .firstOrNull()
                         ?.outputFile
                 )
+                task.finalizedBy(copyToDistTask)
```
liutikas, line 173: Can we avoid using finalizedBy? It breaks incrementality.

Instead we should just use copyToDistTask as the anchor task and name these as such or make RunTestOnFTLTask put the results into dist directory directly without an intermediate task.

buildSrc/src/main/kotlin/androidx/build/ftl/RunTestOnFTLTask.kt
```diff
@@ -45,6 +50,7 @@ import javax.inject.Inject
  * Due to the limitations of FTL, this task only support application instrumentation tests for now.
  */
 @Suppress("UnstableApiUsage") // for gradle property APIs
+@CacheableTask
```
liutikas, line 53: Does this work if we have no `@Input`s specified?

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
```diff
@@ -110,6 +162,95 @@ internal class GCloudCLIWrapper(
...
+                        runId,
+                    )
+                } else {
+                    listOf(
+                        "local",
+                        projectPath,
+                        UUID.randomUUID().toString()
+                    )
+                }
+                return pathValues.joinToString("/")
```
liutikas, line 243: This will not work with windows, can we use File.separator instead?

buildSrc/src/main/kotlin/androidx/build/ftl/RunTestOnFTLTask.kt
yigit: i was actually not sure how to do this so was going to ask it.
Specifically, I noticed that we were doing `doLast` for regular tasks, so wasn't sure if there is a reason for that.
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:buildSrc/src/main/kotlin/androidx/build/AndroidXPlugin.kt;l=501?q=getDistributionDirectory&ss=androidx%2Fplatform%2Fframeworks%2Fsupport&start=11

I'll make the change here.

buildSrc/src/main/kotlin/androidx/build/ftl/RunTestOnFTLTask.kt
yigit: we have inputs here though?

also, what i tried was:

```
./gradlew :room:integration-tests:room-testapp-noappcompat:debugAndroidTestOnFirebaseTestLab --info --build-cache
./gradlew clean
./gradlew :room:integration-tests:room-testapp-noappcompat:debugAndroidTestOnFirebaseTestLab --info --build-cache
```
check that the second command didn't actually run the FTL test, instead the info said "FROM CACHE".
a bunch of other tasks still did run though :'( but not calling firebase would be big save here.

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
yigit: renaming to `buildRelativeResultDirPath` specifically, this does not even make a file instead just creates a relative directory path.

yigit: made it private. I'm always unsure where to scope them but it is in the companion object already and it is scoped to the parameters class so maybe that is OK?

buildSrc/src/main/kotlin/androidx/build/ftl/RunTestOnFTLTask.kt
yigit: i think separate task is better since DIST is not part of build dir, idk how well it will play with caching. 
I might make the anchor task a copy task though, that will be configured by each of these so we'll not get extra tasks while still keeping the firebase run isolated. Let me try that

yigit: oh i remembered why I used `finalizedBy`. I want it to run even when the test fails (which fails the task). If I make it a regular dependency, failure test results will not be copied to the DIST.

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
yigit: actually it will. this method builds a relative path and as far as I can remember, JVM on windows still expects paths to be separated by `/`. 
Also, this path is used for GS buckets which will need `/`, not OS dependent path.
I've not tried running this on windows though

buildSrc/src/main/kotlin/androidx/build/ftl/RunTestOnFTLTask.kt
liutikas: Let's keep it as finalizedBy for now.

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
```diff
@@ -44,53 +46,103 @@ internal class GCloudCLIWrapper(
...
-            "--app", testedApk.canonicalPath,
-            "--num-flaky-test-attempts", "3",
+            "--test", params.testApk.canonicalPath,
+            "--app", params.testedApk.canonicalPath,
+            "--num-flaky-test-attempts", "2",
+            "--results-bucket=${params.bucketName}",
+            "--results-dir=${params.resultsBucketDir}",
+            "--results-history-name=${params.projectPath}"
+        )
+        // copy the test results from the bucket to the build folder
```
liutikas: nit: use directory everywhere instead of folder

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
```diff
@@ -44,53 +46,103 @@ internal class GCloudCLIWrapper(
...
+        val localFolder = params.resultsLocalDir
```
liutikas: Why create the extra variable? Can we use this directly in execGsUtil?

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
```diff
@@ -44,53 +46,103 @@ internal class GCloudCLIWrapper(
...
+        execGsUtil(
+            "cp", "-r", params.cloudBucketPath() + "/*", localFolder.canonicalPath
         )
+        // finally, write the command response into the folder as well
+        val testResultOutput = params.resultsLocalDir.resolve(TEST_OUTPUT_FILE_NAME)
+        testResultOutput.bufferedWriter(Charsets.UTF_8).use {
```
liutikas, line 115: Do we need to close() testResultOutput?

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
```diff
@@ -44,53 +46,103 @@ internal class GCloudCLIWrapper(
     /**
      * Path to the gcloud executable, derived from `which gcloud` call.
      */
-    private val executable: String by lazy {
+    private val gcloud: String by lazy {
+        findExecutable("gcloud")
```
liutikas, line 50: If gcloud is already on PATH, why do we need to find its exact location explicitly? Does execOperations not do that for us?

liutikas (approved):

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
yigit: yea we can. :) fixed.

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
yigit: the `use` function does that. https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.io/use.html

buildSrc/src/main/kotlin/androidx/build/ftl/GCloudCLIWrapper.kt
yigit: actually i didn't try that. I can try it but looking for the executable lets me print a better error. Otherwise, we would need to parse the output of any call for "not-found" executable. If you don't mind, i'd rather keep it this way.

PR #173, opened 2021-05-03
[Navigation] Introduce navigation-compose-material and support for bottom sheet navigation destinations

jossiwolf:
## Proposed Changes

For easier integration with `compose-material`'s `ModalBottomSheetLayout`, this library provides convenience wrappers and a `BottomSheetNavigator` to be able to declare and navigate to bottom sheet destinations, similar to bottom sheet dialog destinations in vanilla navigation.
`BottomSheetNavigator` drives a `ModalBottomSheetLayout`'s `sheetContent` and state, controlling visibility according to the back stack and composing and disposing destinations that are navigated to or away from. `BottomSheetNavigator` offers the `sheetContent` as a composable that should be set as the `ModalBottomSheetLayout`'s `sheetContent`.

- Composable bottom sheet destinations can be added using `NavGraphBuilder#bottomSheet`
- The BottomSheetNavigator needs to be registered with the `NavController`'s `navigatorProvider`
- Provides a convenience wrapper around `ModalBottomSheetLayout` that accepts a `BottomSheetNavigator`
- Extracted `BackStackEntryIdViewModel` and the `SaveableStateHolder.SaveableStateProvider` extension into their own file in `navigation-compose` because they are needed in `navigation-compose-material`

~~Generating the API files for `navigation-compose-material` makes lint trip up about overloads/Java compatibility on a composable function (`src/main/java/androidx/navigation/compose/material/BottomSheet.kt:35: error: A Kotlin method with default parameter values should be annotated with @JvmOverloads for better Java interoperability`), so any pointers there would be appreciated :)~~ as a workaround I added `navigation-compose-material` to the list of exceptions in `AndroidXExtension#targetsJavaConsumers` locally to generate the API files. That might even be the correct solution, let me know what you think is best.

I also thought about making `navigation-compose-material` a submodule of `navigation-compose` - would that be better?

Please let me know if you have ideas for more tests, I plan on adding a couple more, not super happy with them yet.

Cheers!

## Testing

Test: Added tests for new behavior (`BottomSheetNavigatorTest`, `SheetContentHostTest`, `NavGraphBuilderTest`), ran with `./gradlew test connectedCheck`. More tests to be added.
Demos: Introduced BottomSheetNavDemo

## Issues Fixed

Fixes: b/180247978


buildSrc/src/main/kotlin/androidx/build/LibraryVersions.kt
```diff
@@ -88,6 +88,7 @@ object LibraryVersions {
     val MEDIA2 = Version("1.2.0-alpha02")
     val MEDIAROUTER = Version("1.3.0-alpha02")
     val NAVIGATION = Version("2.4.0-alpha01")
+    val NAVIGATION_COMPOSE_MATERIAL = Version("1.0.0-alpha01")
```
ianhanniballake: Navigation is back to using just a single version of all artifacts; we don't want or need a separate version for this artifact (it can just be part of 2.4.0-alpha01 with everything else).

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
```diff
@@ -0,0 +1,125 @@
...
+ * [androidx.navigation.compose.material.bottomSheet].
+ * <b>The [sheetContent] [Composable] will always host the latest entry of the back stack. Nested
+ * flows inside
+ * floating windows should be avoided.</b>
+ *
+ * @param sheetState The [ModalBottomSheetState] that the [BottomSheetNavigator] will use to
+ * drive the sheet state
+ */
+@OptIn(ExperimentalMaterialApi::class)
+@Navigator.Name("ComposeBottomSheetNavigator")
```
ianhanniballake: Use the same name as the `NavGraphBuilder` extension - `"bottomSheet"`

navigation/navigation-compose-material/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+    id("org.jetbrains.kotlin.android")
+}
+
+dependencies {
+    kotlinPlugin(projectOrArtifact(":compose:compiler:compiler"))
+
+    implementation(KOTLIN_STDLIB)
+    api(projectOrArtifact(":compose:runtime:runtime"))
+    api(projectOrArtifact(":compose:runtime:runtime-saveable"))
+    api(projectOrArtifact(":navigation:navigation-common-ktx"))
```
ianhanniballake: You don't need this dependency as it is transitively included in `navigation-compose`.

navigation/navigation-compose-material/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+        exclude group: "androidx.navigation", module: "navigation-common"
+    })
+    androidTestImplementation(projectOrArtifact(":compose:ui:ui-test-junit4"))
+    androidTestImplementation(ANDROIDX_TEST_RUNNER)
+    androidTestImplementation(JUNIT)
+    androidTestImplementation(TRUTH)
+}
+
+androidx {
+    name = "Compose Material Navigation"
```
ianhanniballake: The name should be changed to Navigation Compose Material (we always refer to it as Navigation Compose)

navigation/navigation-compose-material/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+    publish = Publish.SNAPSHOT_AND_RELEASE
+    mavenVersion = LibraryVersions.NAVIGATION_COMPOSE_MATERIAL
+    mavenGroup = LibraryGroups.NAVIGATION
+    inceptionYear = "2021"
+    description = "Compose Material integration with Navigation"
```
ianhanniballake: Reverse the ordering here as well - Navigation integration with Compose Material

navigation/navigation-compose-material/build.gradle
```diff
@@ -0,0 +1,62 @@
...
+    runApiTasks = new RunApiTasks.Yes()
+    legacyDisableKotlinStrictApiMode = true
```
ianhanniballake: Generally for Navigation artifacts, we've enabled strict API mode (it has caught a number of bugs for us). I'm personally not particularly concerned about visibility issues in our androidTest code so feel free to either ignore or just mark them all as public as needed (it doesn't actually matter).

navigation/navigation-compose-material/integration-tests/navigation-demos/build.gradle
```diff
@@ -0,0 +1,47 @@
...
+    implementation(projectOrArtifact(":compose:foundation:foundation"))
+    implementation(projectOrArtifact(":compose:material:material"))
+    implementation(projectOrArtifact(":navigation:navigation-compose-material"))
+    implementation(projectOrArtifact(
+            ":navigation:navigation-compose-material:navigation-compose-material-samples"
+    ))
+}
+
+androidx {
+    name = "Compose Material Navigation Demos"
```
ianhanniballake: Ditto on reordering the names to Navigation Compose Material Demos

navigation/navigation-compose-material/integration-tests/navigation-demos/src/main/java/androidx/navigation/compose/material/demos/BottomSheetNavDemo.kt
```diff
@@ -0,0 +1,96 @@
...
+    const val Feed = "FEED"
+    const val Sheet = "SHEET"
+}
+
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun BottomSheetNavDemo() {
+    val navController = rememberNavController()
+    val bottomSheetNavigator = rememberBottomSheetNavigator()
+    navController.navigatorProvider += bottomSheetNavigator
```
ianhanniballake, line 47: This is going to cause issues if this method recomposes. Perhaps what we actually want here is a `rememberNavController()` overload (/optional parameter) that takes a `NavHostController.() -> Unit` lambda where we could do this type of configuration when the NavController is created?

I'm not sure how we'd effectively match the `remember` scoping of the `NavController` and the `BottomSheetNavigator`, as generally we'd only want those to be 1:1. Perhaps not really an issue since the `remember` is ~unconditional.

navigation/navigation-compose-material/integration-tests/navigation-demos/src/main/java/androidx/navigation/compose/material/demos/BottomSheetNavDemo.kt
```diff
@@ -0,0 +1,96 @@
...
+@Composable
+private fun BottomSheet(showFeed: () -> Unit, showAnotherSheet: () -> Unit, arg: String) {
+    Text("Sheet with arg: $arg")
+    Button(onClick = showFeed) {
+        Text("Click me to navigate!")
+    }
+    Button(onClick = showAnotherSheet) {
+        Text("Click me to show another sheet!")
+    }
+}
```
ianhanniballake: Make sure to add a newline at the bottom of every new file.

navigation/navigation-compose-material/samples/build.gradle
```diff
@@ -0,0 +1,48 @@
...
+    implementation(KOTLIN_STDLIB)
+
+    compileOnly(projectOrArtifact(":annotation:annotation-sampled"))
+    implementation(projectOrArtifact(":compose:foundation:foundation"))
+    implementation(projectOrArtifact(":navigation:navigation-compose-material"))
+    implementation(projectOrArtifact(":compose:material:material"))
+}
+
+androidx {
+    name = "AndroidX Compose UI Navigation Material Integration Samples"
```
ianhanniballake, line 42: Ditto on the name and description.

navigation/navigation-compose-material/src/androidTest/java/androidx/navigation/compose/material/SheetContentHostTest.kt
```diff
@@ -0,0 +1,131 @@
...
+        composeTestRule.onNodeWithTag(bodyContentTag).performClick()
+        composeTestRule.awaitIdle()
+        assertWithMessage("Sheet should be hidden")
+            .that(sheetState.isVisible).isFalse()
+        assertWithMessage("Back stack entry should be in the dismissed entries list")
+            .that(dismissedBackStackEntries)
+            .containsExactly(backStackEntry)
+    }
+
+
```
ianhanniballake, line 217: Nit: Make sure to clean up multiple empty lines in a row.

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
```diff
@@ -0,0 +1,125 @@
...
+import kotlinx.coroutines.flow.MutableStateFlow
+import kotlinx.coroutines.flow.StateFlow
+
+/**
+ * Create and remember a [BottomSheetNavigator]
+ *
+ * @param sheetState The [ModalBottomSheetState] that the [BottomSheetNavigator] will use to
+ * drive the sheet state
+ */
+@OptIn(ExperimentalMaterialApi::class)
```
ianhanniballake, line 70: We'll need to follow up the Compose team - while we can use experimental APIs in alpha, policy states that we **cannot** move to beta with APIs that depend on experimental APIs from another artifact group. We'll need to get agreement that these APIs will become stable before Compose 1.0 before we can add APIs built on top of them.

I can do most of this work if you'd like - I'd appreciate a list of APIs we'll need to implement this fully (or if there's some minimal set we can use and leave helpers that need more experimental APIs as demos / docs only).

ianhanniballake (changes requested):
Just a high level review at the moment - I hope to have some time to dive into the details in a later review (one CL huh? lol).

It looks like there's a few more infrastructural things (like the `SaveableStateProvider`) that we should split off into separate PRs that can be merged in prior to this. Let me know if you'd like to do this or if you'd like us to do this ahead of time.

I have some ideas on how we'd like to approach each of these if you'd like to reach out to me or just list out what dependencies you need here and I can comment on each.

navigation/navigation-compose-material/build.gradle
jossiwolf: Sounds good! Heh maybe that was causing the issues with generating the lint baseline? 🤔 (yes, looks like it actually was!)

Looks like it's [disabled in `navigation-compose`](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:navigation/navigation-compose/build.gradle;l=62) too, maybe we should remove that there too?

I saw that the Compose libraries apply the Kotlin plugin through `AndroidXUiPlugin.applyAndConfigureKotlinPlugin(project)`, I think it'd be good to use that too! There were a couple of lint checks (composable naming) it was complaining about so that'll be nice. Let me know if we shouldn't do this!

Looking at the Compose libraries, they have set `legacyDisableKotlinStrictApiMode = true` too 🤔 I'll just ignore the visibility warnings in tests for now

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
jossiwolf: Oh, interesting. That makes sense and is probably for the better 😄 

I think there are quite a few issues around bottom sheets and the swipeable APIs and I'm not sure what the plan is, but my understanding was that these might well stay experimental until after 1.0. Looking at the APIs we're using though - I think it might be possible to make them stable. At least `ModalBottomSheetState` can probably be made stable by moving the `SwipeableState` into the class and not inheriting from it anymore (same as with `DrawerState`, https://android-review.googlesource.com/c/platform/frameworks/support/+/1613082/).

If you could talk to the team about this, that'd be awesome! (You can also make a group chat with Matvei, you and me if you want)

The APIs we need are
- `ModalBottomSheetLayout` (though we only need that for our convenience method that's `@ExperimentalMaterialApi` anyways)
- `ModalBottomSheetState` for use in `BottomSheetNavigator` and `SheetContentHost`

The APIs we need from `ModalBottomSheetState` are all related to visibility but that's pretty much at the core of the swipeable issues anyways.

navigation/navigation-compose-material/integration-tests/navigation-demos/src/main/java/androidx/navigation/compose/material/demos/BottomSheetNavDemo.kt
jossiwolf: Yeah, good catch! I think an optional parameter would be nice.

```kotlin
val bottomSheetNavigator = rememberBottomSheetNavigator()
val navController = rememberNavController {
    navigatorProvider += bottomSheetNavigator
}
```

```kotlin
@Composable
fun rememberNavController(
    configureNavController: NavHostController.() -> Unit = { }
): NavHostController {
    val context = LocalContext.current
    return rememberSaveable(configureNavController, saver = NavControllerSaver(context)) {
        createNavController(context).apply(configureNavController)
    }
}
```

I don't think that scoping is a big issue here, and making `configureNavController` `@Composable` isn't great either imo. In case it's an issue, we'll see it pop up in alpha I think :)

If you agree on that approach, I'll be happy to put this up in another PR!

Edit: Mhm, I'm not happy with that - I guess the configuration lambda would have to be composable after all? Let me know what your thoughts are. Just questioning how we'd track changes to anything that's referenced in the lambda.

navigation/navigation-compose-material/integration-tests/navigation-demos/src/main/java/androidx/navigation/compose/material/demos/BottomSheetNavDemo.kt
```diff
@@ -0,0 +1,97 @@
...
+                HomeScreen(
+                    showSheet = {
+                        navController.navigate(Destinations.Sheet + "?arg=From Home Screen")
+                    },
+                    showFeed = { navController.navigate(Destinations.Feed) }
+                )
+            }
+            composable(Destinations.Feed) { Text("Feed!") }
+            bottomSheet(Destinations.Sheet + "?arg={arg}") { backstackEntry ->
+                val arg = backstackEntry.arguments?.get("arg") as String
```
ianhanniballake: Nit: use `getString("arg")` rather than casting.

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
```diff
@@ -0,0 +1,149 @@
...
+ * @param sheetState The [ModalBottomSheetState] that the [BottomSheetNavigator] will use to
+ * drive the sheet state
+ * @param navController The [NavController] used to the pop the back stack when the sheet is
+ * dismissed. Please note that this will be redundant in the next release.
+ */
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+public fun rememberBottomSheetNavigator(
+    sheetState: ModalBottomSheetState =
+        rememberModalBottomSheetState(ModalBottomSheetValue.Hidden),
```
ianhanniballake: These lines brought up an interesting question I had:
- We need to allow hoisting this state to allow for coordinating other things with the bottom sheet, is that correct? Or is there some other reason?
- What happens when someone calls `hide()` or `show()` on this state externally from the `BottomSheetNavigator`? What does that do to the `BottomSheetNavigator`?
- What happens after a configuration change? With an initial state of always `ModalBottomSheetValue.Hidden`, does that mean that when the `NavController` restores its state that the bottom sheet animates in again? I guess we don't know the 'intial state' until the `BottomSheetNavigator` has its `onAttach()` called and we can check the `NavigatorState.backStack` to see if it empty or not? I don't see any 'snap to state' kind of API?

jossiwolf: Great questions!

1. Yes! There might be cases where you'd want to hide or show the sheet (though I imagine most of these use cases would be covered by simply navigating) manually. Theoretically, we do not need to offer this control (it would save us some headache when thinking about the initial bottom sheet state; what if somebody set it to `Expanded` here?) though. I'd be open to removing this param! It would help prevent misuse and the `sheetState` can be accessed through the navigator in any case. I'll play around with that a little bit :)
2. The only way we can react to `show`/`hide` calls is by looking at the `sheetVisiblity` - we currently have a `snapshotFlow` observing the visibility and calling a callback (that will then pop the back stack) when the sheet was visible and changed its state to hidden. So that `LaunchedEffect`/`snapshotFlow` covers both manual and programmatic dismiss (we have tests for that in `SheetContentHostTest` too!).  `show()`ing the sheet is a good question... I thought about it but didn't find any good solution. Calling `show()` is a little ambiguous: 
    1. It implies the sheet is currently hidden and the current nav destination is not a `BottomSheetNavigator.Destination`
    2. When you call `show()`, what is the expected outcome? What should we show?
Do you think we should just ignore `show` calls? Or any other ideas?
3. `rememberModalBottomSheetState` uses `rememberSaveable` under the hood so we will stay in the same state across config changes. But we might not actually want that as it means that the sheet and navigator might be out of sync. I think it'd be better to react to that after the config change though, when we actually know if we want to show the sheet or not.
> snap to state API

Is that in the navigator?

jossiwolf: (also, if you're asking these questions, it makes sense to document that - I'll add explanations around this to the docs later on!)

navigation/navigation-compose-material/integration-tests/navigation-demos/src/main/java/androidx/navigation/compose/material/demos/BottomSheetNavDemo.kt
```diff
@@ -0,0 +1,97 @@
...
+    const val Feed = "FEED"
+    const val Sheet = "SHEET"
+}
+
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun BottomSheetNavDemo() {
+    val navController = rememberNavController()
+    val bottomSheetNavigator = rememberBottomSheetNavigator(navController = navController)
+    SideEffect { navController.navigatorProvider += bottomSheetNavigator }
```
chrisfillmore: Hello, I hope you don't mind a question from a stranger, but I am working on a similar problem and was referred to your PR as an example.

My question is, is the use of `SideEffect` correct here? Since the block doesn't run until after composition, if you call `NavGraphBuilder.bottomSheet()` immediately, you will get an IllegalStateException. That is to say, I believe the following hypothetical code would fail:

```Kotlin
val navController = rememberNavController()
val bottomSheetNavigator = rememberBottomSheetNavigator(navController = navController)
SideEffect { navController.navigatorProvider += bottomSheetNavigator }
// Note: not wrapped in ModalBottomSheetLayout
NavHost(...) {
  bottomSheet(...) {
    ...
  }
}
```
It would fail with this error:
> java.lang.IllegalStateException: Could not find Navigator with name "BottomSheetNavigator". You must call NavController.addNavigator() for each navigation type.

If `SideEffect` were removed, there would be no error. However I am not sure if it is correct to remove `SideEffect`. Perhaps @ianhanniballake can advise?

I speculate that you haven't seen this error so far, because the ModalBottomSheetLayout begins in "Hidden" state and is thus not part of the composition, thus the `SideEffect` is run before `bottomSheet()`. I suspect the tests pass because they don't use `SideEffect`. But I am speculating; I haven't run the code.

ianhanniballake: `SideEffect` is specifically exactly what should be used, as explained in [b/187443146](https://issuetracker.google.com/187443146) and specifically made possible in #176. The navigation graph is applied as part of a `DisposableEffect`, so this all works perfectly in practice 😄

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
jossiwolf: Sooo
1. I just went back and forth on this - IMO we should keep exposing this because folks might want to observe the state to run things etc. Maybe a point can be made that this should be done using effects so I'm not entirely against removing it and adding it back if requested.
2. I'm still not sure what we should do when somebody calls `show` - we could just ignore it. Feels like a sensible thing to do to me. Wouldn't be an issue if we weren't hoisting the state.
3. Is working now 🎉 Working on a test

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
```diff
@@ -0,0 +1,129 @@
...
+
+    override fun createDestination(): Destination = Destination(navigator = this, content = {})
+
+    override fun navigate(
+        entries: List<NavBackStackEntry>,
+        navOptions: NavOptions?,
+        navigatorExtras: Extras?
+    ) {
+        entries.forEach { entry ->
+            state.pushWithTransition(entry)
```
ianhanniballake, line 167: If we're `pushWithTransition` multiple entries (either because multiple entries are passed to this method or `navigate` is called multiple times in a row (which is probably the more likely case - that's how deep linking works), but we only display the latest one, how do those other entries get `onTransitionComplete` called?

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
```diff
@@ -0,0 +1,129 @@
...
+    public val sheetContent: @Composable ColumnScope.() -> Unit = @Composable {
+        val columnScope = this
+        val saveableStateHolder = rememberSaveableStateHolder()
+        val backStackEntries by state.backStack.collectAsState()
+        val transitionsInProgress by state.transitionsInProgress.collectAsState()
+
+        // We always replace the sheet's content instead of overlaying and nesting floating
+        // window destinations. That means that only *one* concurrent destination is supported by
+        // this navigator.
+        val latestEntry = backStackEntries.lastOrNull { entry ->
```
ianhanniballake, line 123: So when the system back button is pressed and `NavController` calls `popBackStack()` on the Navigator, how does this Navigator handle that? The default behavior of `popBackStack()` will call `pop`, which will immediately destroy the state of the `NavBackStackEntry` and remove it from `backStackEntries`. Should you be calling `popWithTransition()` there and looking at `transitionsInProgress`.lastOrNull()` and `backStackEntries` here to know what to be hiding?

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/SheetContentHost.kt
```diff
@@ -0,0 +1,130 @@
...
+                .drop(1)
+                // We want to know when the sheet was visible but is not anymore
+                .filter { isVisible -> !isVisible }
+                // Finally, pop the back stack when the sheet has been hidden
+                .collect { onSheetDismissed(backStackEntry) }
+        }
+
+        // Whenever the composable associated with the latestEntry enters the composition, we
+        // want to show the sheet, and hide it when this composable leaves the composition
+        DisposableEffect(Unit) {
```
ianhanniballake: How does this handle navigating from one bottom sheet destination to another? I would have thought you'd need a `DisposableEffect(backStackEntry` or something similar here? When navigating between bottom sheets, what is the expected behavior here? I would think the first one would want to hide completely, then the second one shows? I'm not sure how to write that logic though.

navigation/navigation-compose-material/integration-tests/navigation-demos/src/main/java/androidx/navigation/compose/material/demos/BottomSheetNavDemo.kt
ianhanniballake: Given the discussion in [b/193555369](https://issuetracker.google.com/issues/193555369), I think I was wrong here and that `SideEffect` is not the right choice. As per the [third comment](https://issuetracker.google.com/issues/193555369#comment3), it is safe to apply the Navigator as part of composition at this point, so let's do that until the better API mentioned in that bug is created.

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * Create and remember a [BottomSheetNavigator]
+ *
+ * @param sheetState The [ModalBottomSheetState] that the [BottomSheetNavigator] will use to
+ * drive the sheet state
+ */
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+public fun rememberBottomSheetNavigator(
+    sheetState: ModalBottomSheetState =
```
ianhanniballake: I'm still not entirely on board with having this an an input given that we specifically don't want them to set a different `initialValue` or `confirmStateChange`, not to mention wanting to avoid the calls to `show()` and `hide()` externally.

Perhaps we can expose the public properties of `ModalBottomSheetState` as part of `BottomSheetNavigator`'s API surface, thus allowing developers to use the Navigator directly where they were using the state while maintaining the ownership properties we need.

ianhanniballake (changes requested):

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
jossiwolf: So when the entry gets removed from `backStackEntries`, the `SheetContentHost` will be recomposed with the new entry or `null`, indicating that the sheet should be empty. `SheetContentHost` has a `DisposableEffect` keyed to its `backStackEntry` to hide the sheet whenever this sheet content leaves the composition - so effectively, the sheet is hidden after the entry has been destroyed. Writing that, yeah sounds like we should use `popWithTransition` for `popBackStack`.

I thought about

```kotlin
override fun popBackStack(popUpTo: NavBackStackEntry, savedState: Boolean) {
    state.popWithTransition(popUpTo, savedState)
}
```

```kotlin
SheetContentHost(
   ...,
   onSheetDismissed = { backStackEntry ->
      transitionsInProgress[backStackEntry]?.onTransitionComplete() ?: state.pop(backStackEntry, saveState = false)
   }
)
```

That'd solve the issue of always popping with a transition and give us the possibility to fall back on `pop`, i.e. when the sheet is dismissed by a user gesture.

Is that what you had in mind too?

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/SheetContentHost.kt
jossiwolf: Yeah you're right about that, the effect should be scoped to the entry. The sheet would hide and then a new one appear - I'm not suuuuper happy with that because if you think about nesting destinations in a sheet you'll most likely want to have a transition along the x-axis. But I think that's out of scope for this PR :D

ianhanniballake: Well, navigating to a new bottom sheet destination is not about replacing the content within the bottom sheet, but hiding the existing bottom sheet destination and then showing a brand new bottom sheet. That's the styling we should shoot for.

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
ianhanniballake: That seems reasonable to me, yeah.

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/SheetContentHost.kt
jossiwolf: Yup. Would be cool to enable something else in the future but we should stick to "hide then show" for now (that's also what it is)

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
jossiwolf: Yeah I agree. Done in d0a03a0, would love to get your thoughts about using an extra class to represent that. Adding these to the navigator felt a bit bloaty, but discoverability is worse this way.

navigation/navigation-compose-material/src/main/java/androidx/navigation/compose/material/BottomSheetNavigator.kt
jossiwolf: Done in e2d181a

navigation/navigation-compose-material/integration-tests/navigation-demos/src/main/java/androidx/navigation/compose/material/demos/BottomSheetNavDemo.kt
```diff
@@ -0,0 +1,97 @@
...
+    const val Feed = "FEED"
+    const val Sheet = "SHEET"
+}
+
+@OptIn(ExperimentalMaterialApi::class)
+@Composable
+fun BottomSheetNavDemo() {
+    val navController = rememberNavController()
+    val bottomSheetNavigator = rememberBottomSheetNavigator()
+    SideEffect { navController.navigatorProvider += bottomSheetNavigator }
```
ianhanniballake: Just a reminder to remove the `SideEffect` here.

navigation/navigation-compose-material/src/androidTest/java/androidx/navigation/compose/material/NavGraphBuilderTest.kt
```diff
@@ -0,0 +1,171 @@
...
+    val composeTestRule = createComposeRule()
+
+    @Test
+    fun testCurrentBackStackEntryNavigate() {
+        lateinit var navController: TestNavHostController
+        val key = "key"
+        val arg = "myarg"
+        composeTestRule.setContent {
+            navController = TestNavHostController(LocalContext.current)
+            navController.navigatorProvider.addNavigator(createBottomSheetNavigator())
```
ianhanniballake: You can use `+=` instead of `addNavigator` here and elsewhere

navigation/navigation-compose-material/src/androidTest/java/androidx/navigation/compose/material/NavGraphBuilderTest.kt
```diff
@@ -0,0 +1,171 @@
...
+            NavHost(navController, startDestination = firstRoute) {
+                navigation(startDestination = secondRoute, route = firstRoute) {
+                    bottomSheet(secondRoute) { }
+                }
+            }
+        }
+
+        composeTestRule.runOnUiThread {
+            assertWithMessage("Destination should be added to the graph")
+                .that(firstRoute in navController.graph)
```
ianhanniballake: `firstRoute` is the route of the navigation graph and not the `bottomSheet` destination itself - don't we want to check on the existence of `secondRoute`? That seems to imply that this check should be something like `secondRoute in (navController.graph.findNode(firstRoute) as NavGraph)`?

navigation/navigation-compose-material/src/androidTest/java/androidx/navigation/compose/material/NavGraphBuilderTest.kt
```diff
@@ -0,0 +1,171 @@
...
+                .isTrue()
+        }
+    }
+
+    @Test
+    fun testNavigationNestedInGraph() {
```
ianhanniballake: TBH, I'm not sure if this adds anything over the previous test, maybe they can be combined? The difference between start destination and not start destination isn't related to the `bottomSheet` extension method.

ianhanniballake (commented):
I'm really happy with how this works.

While we can't approve this due to the usage of experimental APIs across different artifact groups, I'd be absolutely happy with this as part of Accompanist or as a standalone library that anyone using Navigation Compose can use.

navigation/navigation-compose-material/src/androidTest/java/androidx/navigation/compose/material/NavGraphBuilderTest.kt
jossiwolf: Yep good catch!

navigation/navigation-compose-material/src/androidTest/java/androidx/navigation/compose/material/NavGraphBuilderTest.kt
jossiwolf: Ya I agree! I don't think it's needed, removed it :)

jossiwolf:
Ayyyy whohoo! Thank you so much for sticking with this PR :) Gonna move over to Accompanist!

(btw, checks are failing because of Metalava. Hope it'll just magically work over there :D)

PR #175, opened 2021-05-05
[Navigation] Introduce NavBackStackEntry#provideToCompositionLocals

jossiwolf:
…that provides the NavBackStackEntry to the composition locals needed for all Compose-related navigators.

Relnote: Introduced a new NavBackStackEntry#provideToCompositionLocals that provides the NavBackStackEntry to the relevant composition locals.

## Testing

Test: NavBackStackEntryProviderTest

`NavBackStackEntryProviderTest#testSaveableValueInContentIsSaved` is pretty much just copied from [RememberSaveableTest](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/runtime/runtime-saveable/src/androidAndroidTest/kotlin/androidx/compose/runtime/saveable/RememberSaveableTest.kt) and as always, I'll be very happy about suggestions on improving the tests and additional tests! 

## Issues Fixed

Fixes: b/187229439

### DIFF
--- a/navigation/navigation-compose/src/androidTest/java/androidx/navigation/compose/NavBackStackEntryProviderTest.kt
+++ b/navigation/navigation-compose/src/androidTest/java/androidx/navigation/compose/NavBackStackEntryProviderTest.kt
@@ -1,0 +1,159 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.navigation.compose
+
+import android.os.Bundle
+import androidx.compose.runtime.DisposableEffect
+import androidx.compose.runtime.getValue
+import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
+import androidx.compose.runtime.saveable.SaveableStateHolder
+import androidx.compose.runtime.saveable.rememberSaveable
+import androidx.compose.runtime.saveable.rememberSaveableStateHolder
+import androidx.compose.runtime.setValue
+import androidx.compose.ui.platform.LocalLifecycleOwner
+import androidx.compose.ui.platform.LocalSavedStateRegistryOwner
+import androidx.compose.ui.test.junit4.StateRestorationTester
+import androidx.compose.ui.test.junit4.createComposeRule
+import androidx.lifecycle.Lifecycle
+import androidx.lifecycle.LifecycleOwner
+import androidx.lifecycle.ViewModelStoreOwner
+import androidx.lifecycle.viewmodel.compose.LocalViewModelStoreOwner
+import androidx.navigation.NavDestination
+import androidx.navigation.NavigatorState
+import androidx.navigation.testing.TestNavigatorState
+import androidx.savedstate.SavedStateRegistryOwner
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.internal.runner.junit4.statement.UiThreadStatement.runOnUiThread
+import androidx.testutils.TestNavigator
+import com.google.common.truth.Truth.assertThat
+import com.google.common.truth.Truth.assertWithMessage
+import kotlinx.coroutines.runBlocking
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import kotlin.random.Random
+
+@RunWith(AndroidJUnit4::class)
+class NavBackStackEntryProviderTest {
+
+    @get:Rule
+    val composeTestRule = createComposeRule()
+
+    @Test
+    fun testViewModelStoreOwnerProvided() {
+        val testNavigator = TestNavigator()
+        val testNavigatorState = TestNavigatorState()
+        val backStackEntry = testNavigatorState.createActiveBackStackEntry(
+            testNavigator.createDestination(),
+            null
+        )
+        var viewModelStoreOwner: ViewModelStoreOwner? = null
+
+        composeTestRule.setContent {
+            val saveableStateHolder = rememberSaveableStateHolder()
+            backStackEntry.provideToCompositionLocals(saveableStateHolder) {
+                viewModelStoreOwner = LocalViewModelStoreOwner.current
+            }
+        }
+
+        assertWithMessage("ViewModelStoreOwner is provided by $backStackEntry")
+            .that(viewModelStoreOwner).isEqualTo(backStackEntry)
+    }
+
+    @Test
+    fun testLifecycleOwnerProvided() {
+        val testNavigator = TestNavigator()
+        val navigatorState = TestNavigatorState()
+        val backStackEntry =
+            navigatorState.createActiveBackStackEntry(testNavigator.createDestination())
+        var lifecycleOwner: LifecycleOwner? = null
+
+        composeTestRule.setContent {
+            val saveableStateHolder = rememberSaveableStateHolder()
+            backStackEntry.provideToCompositionLocals(saveableStateHolder) {
+                lifecycleOwner = LocalLifecycleOwner.current
+            }
+        }
+
+        assertWithMessage("LifecycleOwner is provided by $backStackEntry")
+            .that(lifecycleOwner).isEqualTo(backStackEntry)
+    }
+
+    @Test
+    fun testLocalSavedStateRegistryOwnerProvided() {
+        val testNavigator = TestNavigator()
+        val navigatorState = TestNavigatorState()
+        val backStackEntry =
+            navigatorState.createActiveBackStackEntry(testNavigator.createDestination())
+        var localSavedStateRegistryOwner: SavedStateRegistryOwner? = null
+
+        composeTestRule.setContent {
+            val saveableStateHolder = rememberSaveableStateHolder()
+            backStackEntry.provideToCompositionLocals(saveableStateHolder) {
+                localSavedStateRegistryOwner = LocalSavedStateRegistryOwner.current
+            }
+        }
+
+        assertWithMessage("LocalSavedStateRegistryOwner is provided by $backStackEntry")
+            .that(localSavedStateRegistryOwner).isEqualTo(backStackEntry)
+    }
+
+    @Test
+    fun testSaveableValueInContentIsSaved() {
+        val restorationTester = StateRestorationTester(composeTestRule)
+        var array: IntArray? = null
+
+        val testNavigator = TestNavigator()
+        val navigatorState = TestNavigatorState()
+        val backStackEntry =
+            navigatorState.createActiveBackStackEntry(testNavigator.createDestination())
+
+        restorationTester.setContent {
+            val saveableStateHolder = rememberSaveableStateHolder()
+            backStackEntry.provideToCompositionLocals(saveableStateHolder) {
+                array = rememberSaveable {
+                    intArrayOf(0)
+                }
+            }
+        }
+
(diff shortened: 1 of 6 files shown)


navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavBackStackEntryProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+ * [LocalSavedStateRegistryOwner] to the [content] and saves the [content]'s saveable states with
+ * the given [saveableStateHolder].
+ *
+ * @param saveableStateHolder The [SaveableStateHolder] that holds the saved states
+ * @param content The content [Composable]
+ */
+@SuppressLint("ComposableNaming")
+@Composable
+public fun NavBackStackEntry.provideToCompositionLocals(
+    saveableStateHolder: SaveableStateHolder,
```
ianhanniballake, line 43: Would this work for the simple case if it had a default value of `rememberSaveableStateHolder()`?

navigation/navigation-compose/src/androidTest/java/androidx/navigation/compose/NavBackStackEntryProviderTest.kt
```diff
@@ -0,0 +1,159 @@
...
+
+        restorationTester.emulateSavedInstanceStateRestore()
+
+        assertThat(array).isEqualTo(intArrayOf(1))
+    }
+
+    // By default, NavBackStackEntrys are in the INITIALIZED state and then get moved to the next
+    // appropriate state by the NavController. In case we aren't testing with a NavController,
+    // this sets the entry's lifecycle state to the passed state so that the entry is active.
+    private fun NavigatorState.createActiveBackStackEntry(
```
ianhanniballake: With [aosp/1697987](https://android-review.googlesource.com/c/platform/frameworks/support/+/1697987), I think you can refactor this to:
1) Create the `TestNavigator`
2) Attach a `TestNavigatorState`
3) Create a `NavBackStackEntry`
4) `navigate()` to that `NavBackStackEntry`
5) Return the created `NavBackStackEntry` (which will now be in the right state).

ianhanniballake (changes requested):

navigation/navigation-compose/src/androidTest/java/androidx/navigation/compose/NavBackStackEntryProviderTest.kt
jossiwolf: omg yes, awesome, that's exactly what we need here! Thanks!

navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavBackStackEntryProvider.kt
jossiwolf: Yeah - I chose not to provide a default value because I felt it might discourage hoisting the `SaveableStateHolder` but I guess that's alright and not a big concern!

navigation/navigation-compose/src/androidTest/java/androidx/navigation/compose/NavBackStackEntryProviderTest.kt
jossiwolf: I rewrote the tests to follow that so we'll just merge after your CL has been merged; they're obvs failing rn

jossiwolf: Alright - tests are passing! We're good to go once you're happy :)

navigation/navigation-compose/src/androidTest/java/androidx/navigation/compose/NavBackStackEntryProviderTest.kt
```diff
@@ -0,0 +1,151 @@
...
+    @Test
+    fun testViewModelStoreOwnerProvided() {
+        val testNavigator = TestNavigator()
+        val testNavigatorState = TestNavigatorState()
+        testNavigator.onAttach(testNavigatorState)
+        val backStackEntry = testNavigatorState.createBackStackEntry(
+            testNavigator.createDestination(),
+            null
+        )
+        testNavigator.navigate(listOf(backStackEntry), null, null)
```
ianhanniballake: I'd suggest moving all of these lines into a `private fun createBackStackEntry(): NavBackStackEntry` method to avoid the code duplication in this test suite, similar to the [`createNavController()` helper](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:navigation/navigation-runtime/src/androidTest/java/androidx/navigation/NavControllerTest.kt;l=1994) from other test suites.

ianhanniballake (changes requested):

ianhanniballake (approved):

navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavBackStackEntryProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+public fun NavBackStackEntry.provideToCompositionLocals(
+    saveableStateHolder: SaveableStateHolder = rememberSaveableStateHolder(),
+    content: @Composable () -> Unit
+) {
+    CompositionLocalProvider(
+        LocalViewModelStoreOwner provides this,
+        LocalLifecycleOwner provides this,
+        LocalSavedStateRegistryOwner provides this
+    ) {
+        saveableStateHolder.SaveableStateProvider {
```
andkulikov: I think at some point we had a lint rule to not create such proxy lambdas as they adds unnecessary overhead for the compose runtime. meaning that you can just write 
saveableStateHolder.SaveableStateProvider(content)
instead

ianhanniballake:
It looks like the `api` files are out of date now with the default parameter. Can you run `updateApi` again?

jossiwolf:
Oh yeah, sorry for that!

navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavBackStackEntryProvider.kt
```diff
@@ -0,0 +1,82 @@
...
+ * Provides [this] [NavBackStackEntry] as [LocalViewModelStoreOwner], [LocalLifecycleOwner] and
+ * [LocalSavedStateRegistryOwner] to the [content] and saves the [content]'s saveable states with
+ * the given [saveableStateHolder].
+ *
+ * @param saveableStateHolder The [SaveableStateHolder] that holds the saved states
+ * @param content The content [Composable]
+ */
+@SuppressLint("ComposableNaming")
+@Composable
+public fun NavBackStackEntry.provideToCompositionLocals(
```
andkulikov: this function name doesn't follow our api guidelines:
the name of the composable function which returns unit should start with a capital letter and be a noun. even if it is an extension/member function on other class. 
similarly to how we have SaveableStateHolder.SaveableStateProvider()
I don't know what is the best name, maybe Ian would have some ideas. something like:
NavBackStackEntry.LocalOwnersProvider()

jossiwolf: Naming is hard :D `LocalOwnersProvider` isn't that descriptive to me, what do you think of `NavBackStackEntry. CompositionLocalProvider`? (What Ian originally suggested, just pascal case) @ianhanniballake

jossiwolf: (btw, AS keeps telling me that it should indeed start with a lowercase letter😂)

ianhanniballake: Yeah, I personally liked the scoping of making this an extension on `NavBackStackEntry` (vs the more obvious top level function of `CompositionLocalProvider(NavBackStackEntry, SaveableStateHolder, @Composable () -> Unit)`) and yet, the capital lettering feels very odd when it comes to using it with extension functions.

jossiwolf: Yeah, I agree - Pascal case extensions are a little weird. I can live with both options but camel case feels more natural to me

andkulikov: I agree that on extension function it looks more strange, but the current guidelines are not having exceptions for extension/member functions. 
with LocalOwners I tried to describe what composition locals would be provided, as with just CompositionLocalProvider name it doesn't explain that. but maybe what exactly we provide is an implementation detail, so ok to me

navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavBackStackEntryProvider.kt
andkulikov: I am actually also not sure have a default value here is reasonable. the whole SaveableStateHolder logic is pretty much useless if it is not hoisted somewhere before iterating through all the entries

navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavBackStackEntryProvider.kt
ianhanniballake: This method would be something that every custom Navigator would use (that's why this needs to be public API).

navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavBackStackEntryProvider.kt
ianhanniballake: I was thinking about the case in #173 where only the latest destination is shown (multiple bottom sheets at the same time doesn't make much sense) where having this default would make this more self contained.

I'd be fine leaving out the default and having developers have to think about the hoisting of their `SaveableStateHolder` if you think that better.

jossiwolf: We could leave out the default param and specify in the KDocs that the `saveableStateHolder` should be hoisted? Without that remark folks probably wouldn't actually hoist this.

navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavBackStackEntryProvider.kt
andkulikov: not blocking this pull-request but just out of my curiosity: what does it mean to have a custom Navigation? if I write my own navigation implementation how will I construct NavBackStackEntry objects?

navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavBackStackEntryProvider.kt
andkulikov: the same SaveableStateHolder should be used for all the backstack entries. with code like this
backStack.forEach {
    it.CompositionLocalProvider(rememberSaveableStateHolder()) {...}
}
we have different holder for each backstack, which is not how it is expected to be used.
so I would vote to remove the default param value just to not add extra confusion (we can explain it with more details in the doc as well)

navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavBackStackEntryProvider.kt
ianhanniballake: No, a custom [`Navigator`](https://developer.android.com/reference/kotlin/androidx/navigation/Navigator), not custom Navigation. `NavController` was built in a such a way that it knows nothing about Fragments, Compose, Views, Dialogs, etc.

It is the individual `Navigator` instances that are added to the `NavController` that provide that functionality - `ComposeNavigator` is just a first party implementation of the public API surface, just like the implementation in #173 is a Navigator specifically designed for modal bottom sheets. They can be mixed and matched and NavController is the one who multiplexes them correctly and controls the creation of NavBackStackEntry instances for each Navigator.

navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavBackStackEntryProvider.kt
jossiwolf: @andkulikov Decided to go with `LocalOwnersProvider` as it did make more sense after thinking about it a bit more :) If anybody has any other ideas we can still rename

Also made the param non-optional and improved docs in a8100a0, let me know what you think :)

PR #178, opened 2021-05-10
[Docs] Update api_guidelines docs

tatocaster:
## Proposed Changes

Removed info about Java 7 after Android Studio 4.2 stable release in api_guideline. It was not relevant anymore

## Testing

Test: no tests


liutikas:
The source of truth for this file is not in git, so i created an equivalent change, once it lands it will update. Thanks for this fix.

PR #181, opened 2021-05-24
[Compose] Add policy as a parameter to observeAsState.

avers2000:
## Proposed Changes
Add SnapshotMutationPolicy parameter to observeAsState() to make it possible to control it from outside.

Explanation.

Using unidirectional data flow requires solving a SingleLiveEvent problem. In accordance to official examples on how to operate with LiveData there is a utility class which makes it possible to issue a single event https://github.com/android/architecture-samples/blob/dev-todo-mvvm-live/todoapp/app/src/main/java/com/example/android/architecture/blueprints/todoapp/SingleLiveEvent.java

Some libraries doesn't support Compose and require using AndroidView as interop.
For example, map library. Usually it holds state inside and provides some methods to operate with it, like ZoomIn() or ZoomOut().
In this case connecting all these two things together will give something like this:

```
@Composable
fun HomeContent(viewModel: HomeViewModel = viewModel(), onNavigate: (Int) -> Unit) {
    val event: MapControlEvent by viewModel.mapControlEvents.observeAsState(MapControlEvent.CENTER_ON_USER)

    AndroidView(
        factory = { createMapView(it) },
        update = { view -> updateMapView(view, event) }
    )
    Hud(viewModel = viewModel, onNavigate = onNavigate)
}
```

This code won't trigger map updates as intended in case events are the same. For example clicking zoom in (emitting `MapControlEvent.ZOOM_IN`) multiple times at row will only trigger update once.
It happens because of comparison mechanism inside `mutableStateOf()` which makes Compose only trigger on changes (and it is cool).
But in case of SingleLiveEvent it works against. Even snackbar or other single events won't trigger if there is the same message.

## Testing

Test: didn't do anything special, as it provides the same default value mutableStateOf() itself.

### DIFF
--- a/compose/runtime/runtime-livedata/src/main/java/androidx/compose/runtime/livedata/LiveDataAdapter.kt
+++ b/compose/runtime/runtime-livedata/src/main/java/androidx/compose/runtime/livedata/LiveDataAdapter.kt
@@ -51,9 +51,12 @@ fun <T> LiveData<T>.observeAsState(): State<T?> = observeAsState(value)
  * @sample androidx.compose.runtime.livedata.samples.LiveDataWithInitialSample
  */
 @Composable
-fun <R, T : R> LiveData<T>.observeAsState(initial: R): State<R> {
+fun <R, T : R> LiveData<T>.observeAsState(
+    initial: R,
+    policy: SnapshotMutationPolicy<R> = structuralEqualityPolicy()
+): State<R> {
     val lifecycleOwner = LocalLifecycleOwner.current
-    val state = remember { mutableStateOf(initial) }
+    val state = remember { mutableStateOf(initial, policy) }
     DisposableEffect(this, lifecycleOwner) {
         val observer = Observer<T> { state.value = it }
         observe(lifecycleOwner, observer)


adamp:
SingleLiveEvent, SingleLiveEvent-like code and using recomposition as an event handler should be avoided. The event stream should be part of a reducer that ends in state; the events should not be treated as states themselves. Recomposition should always be idempotent.

There are a few ways to approach this; you can reduce the state in your Compose code, or if the view is already configured to own that state and perform the reduce itself (i.e. it can already accept change request events such as with the `updateMapView(view, event)` part of the snippet above) then manage the event stream subscription from compose and push the events directly to the view from that subscription.

An example of the first approach, keeping the source of truth about the map state in Compose:

```kotlin
@Composable
fun HomeContent(viewModel: HomeViewModel = viewModel(), onNavigate: (Int) -> Unit) {
    val state: MapState by remember(viewModel.mapControlEvents) {
        viewModel.mapControlEvents
            .scan(MapState.Initial) { state, event -> updateMapState(state, event) }
    }.collectAsState(MapState.Initial)

    AndroidView(
        factory = { createMapView(it) },
        update = { view -> setMapState(view, state) }
    )
    Hud(viewModel = viewModel, onNavigate = onNavigate)
}
```

or alternatively, with the second approach, keeping the source of truth about the map state in the view, something like:

```kotlin
@Composable
fun HomeContent(viewModel: HomeViewModel = viewModel(), onNavigate: (Int) -> Unit) {
    val context = LocalContext.current
    // context is highly unlikely to change in practice, but use key to be correct anyway
    key(context) {
        val view = remember { createMapView(context) }

        LaunchedEffect(viewModel.mapControlEvents) {
            viewModel.mapControlEvents.collect { event ->
                updateMapView(view, event)
            }
        }

        AndroidView(factory = { view })
    }
    Hud(viewModel = viewModel, onNavigate = onNavigate)
}
```

Another approach might perform the event reduce in the `ViewModel` instead, and expose the reduced map state from the `ViewModel` for Compose to access. Still another would be to use a [hoisted state object](https://android.googlesource.com/platform/frameworks/support/+/androidx-main/compose/docs/compose-api-guidelines.md#hoisted-state-types) that might host methods that directly implement state mutation, removing the need for the `MapControlEvent` type and stream entirely.

All of these approaches keep state reduction happening as part of handling the event stream with Compose consuming the reduced state.

We've avoided adding this API because there are few correct usages of it, if any. When it's requested, most commonly the proposed use case involves `neverEqualPolicy`. If a caller is consuming a stream with `neverEqualPolicy` chances are good that either that calling code is attempting to treat an event stream as state and should instead follow one of the above patterns or similar, or that the calling code is repeatedly emitting the same mutable object instance after mutating it to signal changes, which will be unsafe for any sort of parallel recomposition features that may be added to Compose in the future.

For these reasons I'd prefer not to add this API.

PR #183, opened 2021-05-27
[Room][XProcessing] Get all annotations on element

elihart:
## Proposed Changes
Adds `getAllAnnotations` to XAnnotated to support retrieving the complete list of annotations on an element without having to query them specifically by class. This allows us to read information on annotations without prior knowledge of what they are. To represent this information I created a new XAnnotation interface because XAnnotationBox requires the class to already be compiled - however, I do have an extension function to convert XAnnotation to XAnnotationBox when the class is available.

This includes support for repeatable annotations, and non primitive types for classes, enums, and annotations (as well as the list varieties).

I ran into a few KSP problems and linked KSP github issues where appropriate.

## Testing

Test: Added XAnnotationTest, which adopts the same tests from XAnnotationBoxTest

## Issues Fixed

Fixes: https://issuetracker.google.com/issues/188853151

### DIFF
--- a/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XValueArgument.kt
+++ b/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XValueArgument.kt
@@ -16,18 +16,17 @@
 
 package androidx.room.compiler.processing
 
-import kotlin.contracts.contract
-
 /**
- * Type elements that represent Enum declarations.
+ * This wraps information about an argument in an annotation.
  */
-interface XEnumTypeElement : XTypeElement {
-    val entries: Set<XEnumEntry>
-}
+interface XValueArgument {
+    /**
+     * The property name.
+     */
+    val name: String
 
-fun XTypeElement.isEnum(): Boolean {
-    contract {
-        returns(true) implies (this@isEnum is XEnumTypeElement)
-    }
-    return this is XEnumTypeElement
+    /**
+     * The value set on the annotation property, or the default value if it was not explicitly set.
+     */
+    val value: Any?
 }
--- a/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacValueArgument.kt
+++ b/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacValueArgument.kt
@@ -16,12 +16,10 @@
 
 package androidx.room.compiler.processing.javac
 
-import androidx.room.compiler.processing.XAnnotation
 import androidx.room.compiler.processing.XEnumTypeElement
 import androidx.room.compiler.processing.XNullability
 import androidx.room.compiler.processing.XValueArgument
 import com.google.auto.common.MoreTypes
-import java.lang.annotation.Repeatable
 import javax.lang.model.element.AnnotationMirror
 import javax.lang.model.element.AnnotationValue
 import javax.lang.model.element.ElementKind
@@ -46,15 +44,12 @@ internal class JavacValueArgument(
         }
         return when {
             // The List implementation further wraps each value as a AnnotationValue.
-            // We don't use arrays because we don't have reified type to instantiate the array
+            // We don't expose arrays because we don't have a reified type to instantiate the array
             // with, and using "Any" prevents the array from being cast to the correct
             // type later on.
             value is List<*> -> value.map { it.unwrapIfNeeded() }
-            // Class types are represented as DeclaredType
             value is TypeMirror -> env.wrap(value, kotlinType = null, XNullability.NONNULL)
-            value is AnnotationMirror -> {
-                JavacAnnotation(env, value)
-            }
+            value is AnnotationMirror -> JavacAnnotation(env, value)
             // Enums are wrapped in a variable element with kind ENUM_CONSTANT
             value is VariableElement -> {
                 when {
--- a/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacValueArgument.kt
+++ b/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacValueArgument.kt
@@ -1,0 +1,75 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.room.compiler.processing.javac
+
+import androidx.room.compiler.processing.XAnnotation
+import androidx.room.compiler.processing.XEnumTypeElement
+import androidx.room.compiler.processing.XNullability
+import androidx.room.compiler.processing.XValueArgument
+import com.google.auto.common.MoreTypes
+import java.lang.annotation.Repeatable
+import javax.lang.model.element.AnnotationMirror
+import javax.lang.model.element.AnnotationValue
+import javax.lang.model.element.ElementKind
+import javax.lang.model.element.ExecutableElement
+import javax.lang.model.element.VariableElement
+import javax.lang.model.type.TypeMirror
+
+internal class JavacValueArgument(
+    val env: JavacProcessingEnv,
+    val element: ExecutableElement,
+    val annotationValue: AnnotationValue
+) : XValueArgument {
+    override val name: String
+        get() = element.simpleName.toString()
+
+    override val value: Any? by lazy { annotationValue.unwrap() }
+
+    private fun AnnotationValue.unwrap(): Any? {
+        val value = value
+        fun Any?.unwrapIfNeeded(): Any? {
+            return if (this is AnnotationValue) this.unwrap() else this
+        }
+        return when {
+            // The List implementation further wraps each value as a AnnotationValue.
+            // We don't use arrays because we don't have reified type to instantiate the array
+            // with, and using "Any" prevents the array from being cast to the correct
+            // type later on.
+            value is List<*> -> value.map { it.unwrapIfNeeded() }
+            // Class types are represented as DeclaredType
+            value is TypeMirror -> env.wrap(value, kotlinType = null, XNullability.NONNULL)
+            value is AnnotationMirror -> {
+                JavacAnnotation(env, value)
+            }
+            // Enums are wrapped in a variable element with kind ENUM_CONSTANT
+            value is VariableElement -> {
+                when {
+                    value.kind == ElementKind.ENUM_CONSTANT -> {
+                        val enumTypeElement = MoreTypes.asTypeElement(value.asType())
+                        JavacEnumEntry(
+                            env,
+                            value,
+                            JavacTypeElement.create(env, enumTypeElement) as XEnumTypeElement
+                        )
+                    }
+                    else -> error("Unexpected annotation value $value for argument $name")
+                }
+            }
+            else -> value
+        }
+    }
+}
--- a/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacValueArgument.kt
+++ b/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacValueArgument.kt
@@ -18,7 +18,7 @@ package androidx.room.compiler.processing.javac
 
 import androidx.room.compiler.processing.XEnumTypeElement
 import androidx.room.compiler.processing.XNullability
-import androidx.room.compiler.processing.XValueArgument
+import androidx.room.compiler.processing.XAnnotationValue
 import com.google.auto.common.MoreTypes
 import javax.lang.model.element.AnnotationMirror
 import javax.lang.model.element.AnnotationValue
@@ -31,14 +31,14 @@ internal class JavacValueArgument(
     val env: JavacProcessingEnv,
     val element: ExecutableElement,
     val annotationValue: AnnotationValue
-) : XValueArgument {
+) : XAnnotationValue {
     override val name: String
         get() = element.simpleName.toString()
 
     override val value: Any? by lazy { annotationValue.unwrap() }
 
     private fun AnnotationValue.unwrap(): Any? {
-        val value = value
+        val value = this.value
         fun Any?.unwrapIfNeeded(): Any? {
             return if (this is AnnotationValue) this.unwrap() else this
         }
--- a/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspTypeMapper.kt
+++ b/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspTypeMapper.kt
@@ -81,7 +81,7 @@ object KspTypeMapper {
         mapping["java.util.List"] = "kotlin.collections.MutableList"
         mapping["java.util.ListIterator"] = "kotlin.collections.ListIterator"
         mapping["java.util.Map"] = "kotlin.collections.MutableMap"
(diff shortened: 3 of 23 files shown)


room/compiler-processing/src/main/java/androidx/room/compiler/processing/InternalXAnnotation.kt
```diff
@@ -0,0 +1,50 @@
...
+ * nested inside it.
+ */
+internal fun XAnnotation.unwrapRepeatedAnnotationsFromContainer(): List<XAnnotation>? {
+    return try {
+        // The contract of a repeatable annotation requires that the container annotation have a single
+        // "default" method that returns an array typed with the repeatable annotation type.
+        val nestedAnnotations = getAsAnnotationList("value")
+
+        // Ideally we would read the value of the Repeatable annotation to get the container class
+        // type and check that it matches "this" type. However, there seems to be a KSP bug where
```
yigit: can you link the bug here?

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XAnnotated.kt
```diff
@@ -36,6 +36,17 @@ interface XAnnotated {
...
 
+    /**
+     * Returns all annotations on this element represented as [XAnnotation].
+     *
+     * As opposed to other functions like [getAnnotations] this does not require you to have a
+     * reference to each annotation class, and thus it can represent annotations in the module
+     * sources being compiled. However, note that the returned [XAnnotation] cannot provide
+     * an instance of the annotation (like [XAnnotationBox.value] can) and instead all values
+     * must be accessed dynamically.
+     */
```
yigit: lets xref from here to toAnnotationBox and from toAnnotationBox to here.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XAnnotated.kt
```diff
@@ -52,6 +63,7 @@ interface XAnnotated {
      * Returns `true` if this element has an annotation that is declared in the given package.
      */
     // a very sad method but helps avoid abstraction annotation
+    // Possibly deprecate in favor of `getAllAnnotations`?
```
yigit: this was a hacky thing i added because we need this in 1 place in room for auto-value and i didn't want us to parse all annotations just to get that.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XAnnotation.kt
```diff
@@ -0,0 +1,110 @@
...
+ *
+ * In comparison, [XAnnotationBox] is used in situations where the annotation class is already
+ * compiled and can be referenced. This can be converted with [asAnnotationBox] if the annotation
+ * class is already compiled.
+ */
+interface XAnnotation {
+    /**
+     * The simple name of the annotation class.
+     */
+    val simpleName: String
```
yigit: might want to name this `name` because all other elements use `name`.

yigit: do we also need qualifiedName ?

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacAnnotation.kt
```diff
@@ -0,0 +1,50 @@
...
+internal class JavacAnnotation(
+    val env: JavacProcessingEnv,
+    val mirror: AnnotationMirror
+) : InternalXAnnotation {
+
+    override val simpleName: String
+        get() = mirror.annotationType.asElement().simpleName.toString()
+
+    override val type: XType by lazy {
+        JavacDeclaredType(env, mirror.annotationType, XNullability.NONNULL)
```
yigit: nullability is UNKONWN here, no ?

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacEnumEntry.kt
```diff
@@ -0,0 +1,37 @@
...
+internal class JavacEnumEntry(
+    env: JavacProcessingEnv,
+    entryElement: Element,
+    override val enumTypeElement: XEnumTypeElement,
+) : JavacElement(env, entryElement), XEnumEntry {
+
+    override val name: String
+        get() = element.simpleName.toString()
+
+    override val equalityItems: Array<out Any?>
```
yigit: you probably want to cache this.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacValueArgument.kt
```diff
@@ -0,0 +1,70 @@
...
+        fun Any?.unwrapIfNeeded(): Any? {
+            return if (this is AnnotationValue) this.unwrap() else this
+        }
+        return when {
+            // The List implementation further wraps each value as a AnnotationValue.
+            // We don't expose arrays because we don't have a reified type to instantiate the array
+            // with, and using "Any" prevents the array from being cast to the correct
+            // type later on.
+            value is List<*> -> value.map { it.unwrapIfNeeded() }
+            value is TypeMirror -> env.wrap(value, kotlinType = null, XNullability.NONNULL)
```
yigit: ditto, is better to use UNKNOWN nullability here

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspAnnotated.kt
```diff
@@ -72,12 +81,28 @@ internal sealed class KspAnnotated(
...
+        ksAnnotation: KSAnnotation,
+        annotationClass: KClass<out Annotation>
+    ): Boolean {
+        val declaration = ksAnnotation.annotationType.resolve().declaration
+        val qualifiedName = declaration.qualifiedName?.asString() ?: return false
+        if (qualifiedName == annotationClass.qualifiedName) return true
+
+        // In cases where a standard library annotation is looked up the KSP name is the kotlin
+        // version and the KClass is the java version, so we need to map the package names.
+        // TODO: Replace with resolver.mapKotlinNameToJava once
```
yigit: we do have a custom mapper here so can you not use it? (KspTypeMapper)

room/compiler-processing/src/test/java/androidx/room/compiler/processing/XAnnotationTest.kt
```diff
@@ -0,0 +1,709 @@
...
+            """.trimIndent()
+        )
+        runTest(
+            sources = listOf(src)
+        ) { invocation ->
+            val subject = invocation.processingEnv.requireTypeElement("Subject")
+            val annotation = subject.requireAnnotation<JavaAnnotationWithTypeReferences>()
+            val annotationValue = if (invocation.isKsp && !preCompiled) {
+                // TODO: 5/24/21 KSP does not wrap a single item in a list, even though the
+                // return type should be Class<?>[] (only in sources).
```
yigit: could we not handle this by loading the annotation class from resolver and see if the method is declared as a list? 
i assume this might be happening in other types too?

yigit (changes requested):
overall lgtm, mostly small nits.
unfortunately you need to rebase because all room files moved to match androidx-directory guidelines.

thank you!

room/compiler-processing/src/main/java/androidx/room/compiler/processing/InternalXAnnotation.kt
elihart: will change. I think this is the right one https://github.com/google/ksp/issues/358

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XAnnotation.kt
elihart: qualifiedName can be access from `xannotation.type.typeelement. qualifiedName`. I provided simple name because to enable access without resolving the type as an optimization.

I can also expose qualifiedName directly as a property to avoid the deep chain needed to otherwise access it if you think it would be a nice thing to have

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacAnnotation.kt
elihart: maybe I'm not fully understanding what this represents and you can better explain, but I was viewing it as an annotation on an annotated element is never null (ie if you get the list of annotations on an element none of those annotations are nullable). also, annotation parameters cannot be nullable so I don't see where a nullable annotation can come from

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacValueArgument.kt
elihart: Same question here, class parameters in annotations cannot be null so why can't we assume the type is non null?

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspAnnotated.kt
elihart: oh yes, thank you I'll do that until the issue is fixed. I suppose that class needs to be replaced with `resolver.mapKotlinNameToJava` eventually?

elihart: Actually on closer inspection I'm not sure it is valid to use the kotlin and java annotation types interchangeably (I commented again in https://github.com/google/ksp/issues/459).

I think it is safer to instead do
```
it.type.typeElement?.hasAnnotation(Repeatable::class) == true ||
            it.type.typeElement?.hasAnnotation(kotlin.annotation.Repeatable::class) == true

File: room/compiler-processing/src/test/java/androidx/room/compiler/processing/XAnnotationTest.kt (hunk 9 above)
Comment (elihart) [reply room/compiler-processing/src/test/java/androidx/room/compiler/processing/XAnnotationTest.kt]: Good idea, I think I have a good general workaround for this now

**Comment by elihart** [comment]
Thanks for the review, all feedback should now be addressed

File: room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspTypeMapper.kt [hunk 10]
```diff
@@ -81,7 +81,7 @@ object KspTypeMapper {
         mapping["java.util.List"] = "kotlin.collections.MutableList"
         mapping["java.util.ListIterator"] = "kotlin.collections.ListIterator"
         mapping["java.util.Map"] = "kotlin.collections.MutableMap"
-        mapping["java.util.Map.Entry"] = "Map.kotlin.collections.MutableEntry"
```
elihart, line 84: @yigit I noticed this seemed like a preexisting bug that you might want fixed

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XAnnotated.kt
elihart: Makes sense - do you want me to remove this comment or do you imagine there could be a follow up to add a cleaner API for this purpose?

room/compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacAnnotation.kt
yigit: oh then you are right, i wasn't thinking that / aware :/.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XAnnotated.kt
yigit: lets just remove the comment. Can still link to the other methods.

room/compiler-processing/src/main/java/androidx/room/compiler/processing/XAnnotation.kt
yigit: we can. btw, i'm not sure if you can do early filtering by name due to typealiases.

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XValueArgument.kt
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.room.compiler.processing
+
+/**
+ * This wraps information about an argument in an annotation.
+ */
+interface XValueArgument {
```
danysantiago: I think `XAnnotationValue` might be a more fitting name, mimicking https://docs.oracle.com/javase/8/docs/api/javax/lang/model/element/AnnotationValue.html or maybe `XAnnotationProperty` which might describe bettere that is both a method and some default value (a getter?).

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XValueArgument.kt
```diff
@@ -0,0 +1,32 @@
...
+    /**
+     * The property name.
+     */
+    val name: String
+
+    /**
+     * The value set on the annotation property, or the default value if it was not explicitly set.
```
danysantiago: Add the possible types to docs:
* Primitives (Boolean, Byte, Int, Long, Float, Double)
* String
* XEnumEntry
* XAnnotation
* XType
* List of any of the above

Hopefully didn't miss any. =D

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacValueArgument.kt
```diff
@@ -0,0 +1,70 @@
...
+    val element: ExecutableElement,
+    val annotationValue: AnnotationValue
+) : XValueArgument {
+    override val name: String
+        get() = element.simpleName.toString()
+
+    override val value: Any? by lazy { annotationValue.unwrap() }
+
+    private fun AnnotationValue.unwrap(): Any? {
+        val value = value
```
danysantiago: nit: `val value = this.value`

Just for clarity

danysantiago (approved):
LGTM

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspAnnotated.kt
yigit: we can add the mapping in your mapper. Unfortunately, KSP mapper does not support everything we need so we'll need to keep ours. (see commit message at: https://android-review.googlesource.com/c/platform/frameworks/support/+/1714392 )

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspTypeMapper.kt
yigit: :facepalm: thanks!

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XValueArgument.kt
elihart: @danysantiago `ValueArgument` is the term that KSP uses in it's API - would you rather follow KSP or Java API naming?

danysantiago: So far we've have been following Java API when available, TypeElement vs ClassDeclaration, ExecutableElement vs FunctionDeclaration so I was thinking to have the same here, specially when this is only used for annotation values where as KSP's ValueArgument could be used to describe arguments in function calls (doesn't right now, but might...)

room/compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspAnnotated.kt
elihart: I think it's actually a bit dangerous to put in a general mapper since it may not always be correct to use them interchangeable - for this case I just check the kotlin and java annotation versions separately

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XValueArgument.kt
elihart: ok, I'll use `XAnnotationValue`

danysantiago: Thank you!

PR #184, opened 2021-05-27
[Fragment] Fixed IndexOutOfBoundsException on fragment transition issue

sante:
## Description
The issue is caused because transition names for elements that are not `ViewGroup` are not null checked, so elements not intended to be a part of the transition are added to the list

## Testing

Test: Could not test the fix.

## Issues Fixed

Fixes: [IndexOutOfBoundsException on fragment transition](https://issuetracker.google.com/issues/188859853)

### DIFF
--- a/fragment/fragment/src/main/java/androidx/fragment/app/DefaultSpecialEffectsController.java
+++ b/fragment/fragment/src/main/java/androidx/fragment/app/DefaultSpecialEffectsController.java
@@ -713,7 +713,8 @@ class DefaultSpecialEffectsController extends SpecialEffectsController {
                 }
             }
         } else {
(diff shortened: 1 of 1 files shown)


sante:
@googlebot I signed it!

Abhishek-eezy:
When can we expect this ? such a major issue !!

ianhanniballake:
This has already been fixed in [this CL](https://android-review.googlesource.com/1732732), with tests to verify that this no longer occurs.

In the future, note that our [contributing guide](https://github.com/androidx/androidx/blob/androidx-main/CONTRIBUTING.md#making-a-pull-request) requires that all pull requests include a test: "Could not test the fix." is not an acceptable regression test.

PR #186, opened 2021-05-28
Use matrix build to simply the build file

yigit:
Moved module builds to a matrix setup to clean duplicated builds.
Use job outputs instead of custom output variables.

Bug: n/a
Test: https://github.com/androidx/androidx/actions/runs/884479260


tikurahul (approved):
This is amazing !

PR #190, opened 2021-06-04
Implement XTypeElement.getEnclosedTypeElements

elihart:
## Proposed Changes
Implement XTypeElement.getEnclosedTypeElements to be able to access classes declared within another class.

## Testing

Test: Added tests to XTypeElementTest

## Issues Fixed

Fixes: https://issuetracker.google.com/issues/190089681

### DIFF
--- a/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
+++ b/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
@@ -157,4 +157,9 @@ interface XTypeElement : XHasModifiers, XElement, XMemberContainer {
      * List of interfaces implemented by this class
      */
     fun getSuperInterfaceElements(): List<XTypeElement>
+
+    /**
+     * Returns the list of all classes nested directly within this class.
+     */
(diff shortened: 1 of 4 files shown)


room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XTypeElement.kt
```diff
@@ -157,4 +157,9 @@ interface XTypeElement : XHasModifiers, XElement, XMemberContainer {
      * List of interfaces implemented by this class
      */
     fun getSuperInterfaceElements(): List<XTypeElement>
+
+    /**
+     * Returns the list of all classes nested directly within this class.
+     */
+    fun getEnclosedTypeElements(): List<XTypeElement>
```
yigit, line 165: will this include companions?
I'm fine either way but we should document it here and also add a test.

for the most part, we don't include them (e.g. static methods inherited from companion are just reported as static methods on the containing class)

elihart: @yigit Good question. This does include companions; I've added a test and documentation.

yigit (approved):

PR #197, opened 2021-06-15
Avoid building playground projects when they are not affected

yigit:
This PR adds a new Gradle task (`findAffectedModules`) that can report the list of modules that are affected for a given set of files. 

That task is later used in the workflow file to check if the playground project is affected by the changes in the RUN and skip building it if it is not affected.

Bug: n/a
Test: ProjectGraphTest, https://github.com/androidx/androidx/actions/runs/939913011


yigit:
presubmit won't validate this since it includes workflow changes.
i pushed to another branch to also build there for reference output:

https://github.com/androidx/androidx/actions/runs/940385851

.github/workflows/presubmit.yml
```diff
@@ -65,6 +66,12 @@ jobs:
         id: changed-files
         uses: androidx/changed-files-action@main
 
+      - name: "Get affected files in push or pull_request"
```
dlam: Btw, we could also just output both from the action although I think the repo checkout for the action should be cached so no big deal.

.github/workflows/presubmit.yml
```diff
@@ -133,9 +146,29 @@ jobs:
           set -x
           echo "ANDROID_SDK_ROOT=$HOME/Library/Android/sdk" >> $GITHUB_ENV
           echo "DIST_DIR=$HOME/dist" >> $GITHUB_ENV
-
+      - name: "Run gradle to check if any of its projects are affected by changes"
```
dlam: Why not name this after the ./gradlew command it runs like the others?

suggest: `./gradlew findAffectedModules --outputFilePath=${{ github.workspace }}/affected.txt`

dlam (approved):

.github/workflows/presubmit.yml
yigit: yea idk why i didn't think of that 🤦 .

yigit: sent a PR, will update this after that one gets merged.

.github/workflows/presubmit.yml
yigit: just put `./gradlew findAffectedModules` w/o arguments since they might be very long.

PR #200, opened 2021-06-17
trigger webhooks in workflow_run actions

yigit:
This PR creates a duplicate call to androidx.dev hooks about workflows.

It adds an additional parameter, src, so that we can distinguish it in the logs.

Bug: n/a
Test: workflow runs only trigger from main branch so i cannot test this on prod without
merging


tikurahul (approved):

yigit:
sample runs:
started hook:
https://github.com/yigit/github-workflow-playground/actions/runs/948086587

result hook:
https://github.com/yigit/github-workflow-playground/actions/runs/948087651

PR #205, opened 2021-06-24
Fix grammar and wording in documentation

picklebento:
## Proposed Changes

  - Grammar fix: Adding a `}`

## Testing

Test: None, since this is documentation


picklebento:
Originally I included a renaming:

- `repeatOnLifecycle() makes sense for a cold flow. However, for hot `SharedFlow` and `StateFlow`, they are likely to run on a different coroutineScope such as `viewModelScope`, which is different from the collector's launching lifecycleScope.`

On second thought, this is not an issue for `SharedFlow` or `StateFlow` with `SharingStarted.WhileSubscribed()`.

jbw0033 (approved):

PR #208, opened 2021-07-02
Change JavacBasicAnnotationProcessor to wrap each step individually.

bcorso:
## Proposed Changes

This commit changes the behavior of JavacBasicAnnotationProcessor to wrap each XProcessingStep individually in its own Step rather than have a single Step that wraps all XProcessingSteps. The issue with wrapping all XProcessingSteps in a single Step is that the underlying BasicAnnotationProcessor triggers the Step, and thus all wrapped XProcessingSteps, when any annotation is found rather than only triggering the Step that handles the annotation.

In order to share the XProcessingEnv for all XProcessingSteps in a round, the first step to process will initialize the XProcessingEnv for other steps. In addition, the postRound call also now shares the XProcessingEnv too. At the end of each postRound I've set the XProcessingEnv to null to allow GC.

## Testing

Test: I've added a test that adds 2 XProcessingSteps to JavacBasicAnnotationProcessor that each handle a different annotation, and then verify that if only one annotation is present in the sources then only the step that handles that annotation is triggered.

## Issues Fixed

Fixes: https://issuetracker.google.com/192658371

### DIFF
--- a/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacBasicAnnotationProcessor.kt
+++ b/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacBasicAnnotationProcessor.kt
@@ -37,7 +37,7 @@ abstract class JavacBasicAnnotationProcessor :
     // This state is cached here so that it can be shared by all steps in a given processing round.
     // The state is initialized at beginning of each round using the InitializingStep, and
     // the state is cleared at the end of each round in BasicAnnotationProcessor#postRound()
-    private var xEnv: JavacProcessingEnv? = null
+    private var cachedXEnv: JavacProcessingEnv? = null
 
     final override fun steps(): Iterable<Step> {
         val delegatingSteps = processingSteps().map { DelegatingStep(it) }
@@ -52,7 +52,7 @@ abstract class JavacBasicAnnotationProcessor :
         override fun process(
             elementsByAnnotation: ImmutableSetMultimap<String, Element>
         ): Set<Element> {
-            xEnv = JavacProcessingEnv(processingEnv)
+            cachedXEnv = JavacProcessingEnv(processingEnv)
             return emptySet()
         }
     }
@@ -64,28 +64,26 @@ abstract class JavacBasicAnnotationProcessor :
         override fun process(
             elementsByAnnotation: ImmutableSetMultimap<String, Element>
         ): Set<Element> {
+            val xEnv = checkNotNull(cachedXEnv)
             val xElementsByAnnotation = mutableMapOf<String, Set<XElement>>()
             xStep.annotations().forEach { annotation ->
                 xElementsByAnnotation[annotation] =
                     elementsByAnnotation[annotation].mapNotNull { element ->
-                        xEnv!!.wrapAnnotatedElement(element, annotation)
+                        xEnv.wrapAnnotatedElement(element, annotation)
                     }.toSet()
             }
-            return xStep.process(xEnv!!, xElementsByAnnotation).map {
+            return xStep.process(xEnv, xElementsByAnnotation).map {
                 (it as JavacElement).element
             }.toSet()
         }
     }
 
     final override fun postRound(roundEnv: RoundEnvironment) {
-        if (xEnv == null) {
-            // On the last round of processing, postRound() is called without calling any of the
-            // processing steps, so we'll need to initialize xEnv ourselves.
-            assert(roundEnv.processingOver())
-            xEnv = JavacProcessingEnv(processingEnv)
-        }
-        val xRound = XRoundEnv.create(xEnv!!, roundEnv)
-        postRound(xEnv!!, xRound)
-        xEnv = null // Reset after every round to allow GC
+        // The cachedXEnv can be null if none of the steps were processed in the round.
+        // In this case, we just create a new one since there is no cached one to share.
+        val xEnv = cachedXEnv ?: JavacProcessingEnv(processingEnv)
+        val xRound = XRoundEnv.create(xEnv, roundEnv)
+        postRound(xEnv, xRound)
+        cachedXEnv = null // Reset after every round to allow GC
     }
 }
--- a/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacBasicAnnotationProcessor.kt
+++ b/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacBasicAnnotationProcessor.kt
@@ -18,12 +18,10 @@ package androidx.room.compiler.processing.javac
 
 import androidx.room.compiler.processing.XBasicAnnotationProcessor
 import androidx.room.compiler.processing.XElement
-import androidx.room.compiler.processing.XProcessingEnv
 import androidx.room.compiler.processing.XProcessingStep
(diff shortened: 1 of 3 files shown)


room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacBasicAnnotationProcessor.kt
```diff
@@ -31,43 +34,58 @@ import javax.lang.model.element.Element
 abstract class JavacBasicAnnotationProcessor :
```
danysantiago, line 32: You should also update `KspBasicAnnotationProcessor` to not call a `xStep` if there are no annotated elements for it too.

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacBasicAnnotationProcessor.kt
```diff
@@ -31,43 +34,58 @@ import javax.lang.model.element.Element
...
+        override fun annotations() = xStep.annotations()
+
+        override fun process(
+            elementsByAnnotation: ImmutableSetMultimap<String, Element>
+        ): Set<Element> {
+            val xElementsByAnnotation = mutableMapOf<String, Set<XElement>>()
+            xStep.annotations().forEach { annotation ->
+                xElementsByAnnotation[annotation] =
+                    elementsByAnnotation[annotation].mapNotNull { element ->
+                        xEnv!!.wrapAnnotatedElement(element, annotation)
```
danysantiago: Instead of using double bang here with `xEnv` you can create a local value to make sure environment is not null:
`val env = checkNotNull(xEnv)` and then use `env` instead of `xEnv!!`

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacBasicAnnotationProcessor.kt
```diff
@@ -31,43 +34,58 @@ import javax.lang.model.element.Element
...
-        return listOf(parentStep)
     }
 
     final override fun postRound(roundEnv: RoundEnvironment) {
-        // Due to BasicAnnotationProcessor taking over AbstractProcessor#process() we can't
-        // share the same XProcessingEnv from the steps, but that might be ok...
-        val xEnv = JavacProcessingEnv(processingEnv)
-        val xRound = XRoundEnv.create(xEnv, roundEnv)
-        postRound(xEnv, xRound)
+        if (xEnv == null) {
```
danysantiago: It would be nice to also have a non-null local value here, so that if there is a NPE it is in one place, basically something like this:
```
val env = xEnv ?: JavacProcessingEnv(processingEnv).also { 
  check(roundEnv.processingOver()) 
}
xEnv = null
```

bcorso: Thanks! 

Btw, I ended up removing the `roundEnv.processingOver()` check because I think we could actually get a null `xEnv` in any round where our steps don't run.

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacBasicAnnotationProcessor.kt
bcorso: Done.

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacBasicAnnotationProcessor.kt
bcorso: Done.

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/ksp/KspBasicAnnotationProcessor.kt
```diff
@@ -44,16 +44,25 @@ abstract class KspBasicAnnotationProcessor(
...
+                    annotatedElements
+                        .filter { (it as KspElement).declaration.validateExceptLocals() }
+                        .toSet()
+
+                deferredElements.addAll(annotatedElements - validAnnotatedElements)
+                if (!validAnnotatedElements.isEmpty()) {
+                    elementsByAnnotation[annotation] = validAnnotatedElements
+                }
+            }
+            // Only process the step if there are annotated elements found for this step.
```
danysantiago, line 61: I think we can do this without increasing the scope of a mutable deferred elements, we need to change the `associateWith` to a `mapNotNull` of `Pair` where `null` is no annotated elements and then make that a map, it should look like this:
```
val deferredElements = processingSteps().flatMap { step ->
    val invalidElements = mutableSetOf<XElement>()
    val elementsByAnnotation = step.annotations().mapNotNull { annotation ->
        val annotatedElements = round.getElementsAnnotatedWith(annotation)
        val validElements = annotatedElements
            .filter { (it as KspElement).declaration.validateExceptLocals() }
            .also { invalidElements.addAll(annotatedElements - it) }
            .toSet()
        if (validElements.isNotEmpty()) {
            annotation to validElements
        } else {
            null
        }
    }.toMap()
    // Only process the step if there are annotated elements found for this step.
    if (elementsByAnnotation.isNotEmpty()) {
        invalidElements + step.process(processingEnv, elementsByAnnotation)
    } else {
        invalidElements
    }
}
```

bcorso: Done.

danysantiago (approved):
Thanks for making this pull request!

danysantiago (approved):

yigit:
ok i'm approving this to let github run the workflows, i hope that is enough (seems like otherwise we have to manually approve each run until @bcorso has 1 PR merged.

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/javac/JavacBasicAnnotationProcessor.kt
```diff
@@ -31,43 +32,45 @@ import javax.lang.model.element.Element
...
-                    )
-                }
-                return results.map { (it as JavacElement).element }.toSet()
+        override fun process(
+            elementsByAnnotation: ImmutableSetMultimap<String, Element>
+        ): Set<Element> {
+            // The first step in a round initializes the cachedXEnv. Note: the "first" step can
+            // change each round depending on which annotations are present in the current round and
+            // which elements were deferred in the previous round.
+            val xEnv = cachedXEnv ?: JavacProcessingEnv(processingEnv)
```
danysantiago: cachedXEnv is never set, try adding a side effect lambda:
`val xEnv = cachedXEnv ?: JavacProcessingEnv(processingEnv).also { cachedXEnv = it }`

bcorso: Oof, good catch!

I even thought we had a test already for that, but unfortunately it was only testing the cache across rounds rather than across steps in the same round. I'll add a test case for this.

danysantiago (approved):

PR #209, opened 2021-07-15
Add Messager#printMessage() overloads to XMessager

bcorso:
## Proposed Changes

Adds [Messager#printMessage()](https://docs.oracle.com/javase/8/docs/api/javax/annotation/processing/Messager.html) overloads to XMessager.

For KSP, there's not a direct equivalent to these overloads since the KspLogger can accept the element/annotation/annotationValue directly without requiring all preceding information. Thus, for KspMessager, we only look at the most specific, non-null parameter given. For example, if element and annotation are given we report on the annotation.

## Testing

Test: ./gradlew :room:r-c-p:test

## Issues Fixed

Fixes: [https://issuetracker.google.com/193819242](https://issuetracker.google.com/193819242)

### DIFF
--- a/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt
+++ b/room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt
@@ -23,24 +23,48 @@ import javax.tools.Diagnostic
  */
 abstract class XMessager {
     private val watchers = mutableListOf<XMessager>()
+
     /**
-     * Prints the given [msg] to the logs while also associating it with the given [element].
+     * Prints the given [msg] to the logs while also associating it with the given [element],
+     * [annotation], and [annotationValue].
      *
      * @param kind Kind of the message
      * @param msg The actual message to report to the compiler
      * @param element The element with whom the message should be associated with
+     * @param annotation The annotation with whom the msg should be associated with
+     * @param annotationValue The annotation value with whom the msg should be associated with
      */
-    fun printMessage(kind: Diagnostic.Kind, msg: String, element: XElement? = null) {
-        watchers.forEach {
-            it.printMessage(kind, msg, element)
+    @JvmOverloads
+    fun printMessage(
+        kind: Diagnostic.Kind,
+        msg: String,
+        element: XElement? = null,
+        annotation: XAnnotation? = null,
+        annotationValue: XAnnotationValue? = null
+    ) {
+        if (element == null) {
+            assert(annotation == null && annotationValue == null) {
+                "If element is null, annotation and annotationValue must also be null."
+            }
         }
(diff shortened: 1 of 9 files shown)


room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt
```diff
@@ -23,24 +23,47 @@ import javax.tools.Diagnostic
...
+    @JvmOverloads
+    fun printMessage(
+        kind: Diagnostic.Kind,
+        msg: String,
+        element: XElement? = null,
+        annotation: XAnnotation? = null,
+        annotationValue: XAnnotationValue? = null
+    ) {
+        if (element == null) {
+            assert(annotation == null && annotationValue == null) {
```
danysantiago: Use `check` instead of `assert`, assertion  need to be turned ON for them to throw.

room/room-compiler-processing/src/main/java/androidx/room/compiler/processing/XMessager.kt
```diff
@@ -23,24 +23,47 @@ import javax.tools.Diagnostic
...
+     * [annotation], and [annotationValue].
      *
      * @param kind Kind of the message
      * @param msg The actual message to report to the compiler
      * @param element The element with whom the message should be associated with
+     * @param annotation The annotation with whom the msg should be associated with
+     * @param annotationValue The annotation value with whom the msg should be associated with
      */
-    fun printMessage(kind: Diagnostic.Kind, msg: String, element: XElement? = null) {
+    @JvmOverloads
```
danysantiago: In terms of API it might be better to declare the overload manually for the non-null contract of the params:
```
fun printMessage(
    kind: Diagnostic.Kind,
    msg: String,
) {
    printMsg(kind, msg, null, null, null)
}

fun printMessage(
    kind: Diagnostic.Kind,
    msg: String,
    element: XElement,
) {
    printMsg(kind, msg, element, null, null)
}

fun printMessage(
    kind: Diagnostic.Kind,
    msg: String,
    element: XElement,
    annotation: XAnnotation,
) {
    printMsg(kind, msg, element, annotation, null)
}

fun printMessage(
    kind: Diagnostic.Kind,
    msg: String,
    element: XElement,
    annotation: XAnnotation,
    annotationValue: XAnnotationValue
) {
    printMsg(kind, msg, element, annotation, annotationValue)
}

internal fun printMsg(
    kind: Diagnostic.Kind,
    msg: String,
    element: XElement? = null,
    annotation: XAnnotation? = null,
    annotationValue: XAnnotationValue? = null
) {
  // actual print
}
```

danysantiago (approved):

danysantiago (approved):

PR #212, opened 2021-07-24
Find best matched font in TypefaceCompatApi29Impl

RikkaW:
## Proposed Changes

Use best-matched font rather than "hardcoded"

The original part

```
final FontStyle defaultStyle = new FontStyle(
        (style & Typeface.BOLD) != 0 ? FontStyle.FONT_WEIGHT_BOLD
                : FontStyle.FONT_WEIGHT_NORMAL,
        (style & Typeface.ITALIC) != 0 ? FontStyle.FONT_SLANT_ITALIC
                : FontStyle.FONT_SLANT_UPRIGHT
);
return new Typeface.CustomFallbackBuilder(familyBuilder.build())
       .setStyle(defaultStyle)
       .build();
```

This will make the final `Typeface` ALWAYS have either 400 (FONT_WEIGHT_NORMAL) or 700 (FONT_WEIGHT_BOLD) weight. It is incorrect since fonts passed to these methods could have other font-weight. For example, pass only one font with 500 weight, the final `Typeface` will have 400 weight, making the fallback font displayed as 400 weight rather than 500.

This pull request copies the platform's algorithm to find the best-match font from the font family.

## Testing

Test: Observed that the font weight is correct

## Issues Fixed

Fixes: Fix https://issuetracker.google.com/issues/194553426

### DIFF
--- a/core/core/src/main/java/androidx/core/graphics/TypefaceCompatApi29Impl.java
+++ b/core/core/src/main/java/androidx/core/graphics/TypefaceCompatApi29Impl.java
@@ -42,6 +42,27 @@ import java.io.InputStream;
 @RestrictTo(LIBRARY_GROUP)
 @RequiresApi(29)
 public class TypefaceCompatApi29Impl extends TypefaceCompatBaseImpl {
+
+    private static int getMatchScore(@NonNull FontStyle o1, @NonNull FontStyle o2) {
+        return Math.abs((o1.getWeight() - o2.getWeight())) / 100 + (o1.getSlant() == o2.getSlant() ? 0 : 2);
+    }
+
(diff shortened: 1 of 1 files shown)


dlam (approved):

dlam:
Thanks for submitting this and also filing a bug for tracking! Sorry it took awhile to get to this, we're not "officially" taking PRs from Github for text, but spoke with the owner and he seems okay with this change 😃

objcode:
Thank you so much for the change :partying_face: !

It looks like it should merge. I will see this CL through the rest of the way on the gerrit side.

Sorry about the delay in reviewing.

PR #218, opened 2021-08-16
Add missing word in DataStore KDoc & Javadoc

LouisCAD:


rohitsat13 (approved):
Thanks for fixing this Louis!

PR #227, opened 2021-08-23
Workaround for projectDir conflicts

dlam:
Points the automatically generated parent project's projectDir to a
non-existent file to workaround Gradle limitations around conflict
project roots. Normally, the automatically generated parent project
would default to its relative dir, but since it conflicts with the root
project in Playground builds, it reverts to
`frameworks/support/parent/parent`, which conflicts with a real project
in some playgrounds. This causes the AndroidXPlugin to get applied to
the parent project incorrectly, but also causes adverse affects in
studio's code completion and conflicts with the child project if it
exists.

Fixes: 197253160
Test: cd activity && ./gradlew tasks

### DIFF
--- a/activity/settings.gradle
+++ b/activity/settings.gradle
@@ -26,3 +26,8 @@ selectProjectsFromAndroidX({ name ->
     return false
 })
 
+// Workaround for b/197253160; sets the projectDir for automatically generated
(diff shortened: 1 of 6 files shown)


dlam:
Continued in #260

PR #228, opened 2021-08-23
Support playground during CHECKOUT_ROOT detection

dlam:
Currently, we hard-code "../.." to get to the root of git checkout for
declaring a location for build outputs, however this causes issues in
Github checkouts which live directly in ./androidx as opposed
to ./frameworks/support.

This changes the hard-coded "../.." to instead use git to parse the root
checkout directory.

Test: Manually - Gradle sync within Studio from github checkout

### DIFF
--- a/buildSrc/out.gradle
+++ b/buildSrc/out.gradle
@@ -26,7 +26,18 @@ def chooseOutDir(subdir = "") {
         def checkoutRoot = System.getProperty("CHECKOUT_ROOT")
         if (checkoutRoot == null) {
             def workingDir = new File(buildscript.sourceFile.parent)
-            checkoutRoot = new File("git rev-parse --show-toplevel".execute([], workingDir).text)
+            // Note: Using String.execute() causes exit code 148, so we use ProcessBuilder instead.
+            def repoProcess = new ProcessBuilder()
+                .command("repo", "help")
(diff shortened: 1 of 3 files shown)


dlam:
This fixes gradle sync when opening project from playground Studio from github checkout

dlam:
maybe we could ask jeff to review after we mirror into aosp?

buildSrc/out.gradle
```diff
@@ -25,7 +25,8 @@ def chooseOutDir(subdir = "") {
     if (outDir == null) {
         def checkoutRoot = System.getProperty("CHECKOUT_ROOT")
         if (checkoutRoot == null) {
-            checkoutRoot = new File("${buildscript.sourceFile.parent}/../../..")
+            def workingDir = new File(buildscript.sourceFile.parent)
```
yigit, line 28: not sure how this works. git repo is the same in the repo tool checkout as well.
we might consider hardcoding parent folders and cheeck for frameworks/support

dlam: Ah yeah I may have mispoke here, the git repo root is the same, but this returns the repo checkout root as its relative of buildSrc/ but this leads outside the checkout in github

dlam: Done

PR #232, opened 2021-08-25
Allow null result from switchMap extension transform

fargus9:
The existing extension method breaks one of the features of the underlying method - that you aren't forced to return a LiveData on every invocation of the transform.

## Proposed Changes

  -
  -
  -

## Testing

## Issues Fixed


fargus9:
@googlebot I signed it!

dlam:
We are having some issues with our github infra at the moment, but you'll need to run `./gradlew updateApi` if you mean to change public api. If this is already stable, you cannot change it in a binary / source incompatible way.

Please also add the `Test: <test> stanza` as mentioned in the PR template.

fargus9:
This was resolved by https://android-review.googlesource.com/c/platform/frameworks/support/+/1969579

PR #234, opened 2021-08-26
Ensure the coroutine cannot be garbage collected

LouisCAD:
## Proposed Changes

Before this change, the coroutine not being garbage collected relies on the fact that `consumeEach` indirectly injects a strong reference to the `Continuation` inside the `Channel` while the iteration is suspended. This is brittle, because if an API that used `WeakReference`s under the hood was to be called in that coroutine, it could then be garbage collected, creating very subtle bugs that don't always happen since the Garbage Collector doesn't run at predictable times.

These changes keep a strong reference to the `CoroutineScope` which keeps a strong reference to its underlying `Job` instance, which keeps a reference to the coroutines launched in it, regardless of the implementation details of the suspending functions being called.

## Testing

Test: This code is already tested, it should not change the behavior in the current form.

### DIFF
--- a/compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/platform/GlobalSnapshotManager.android.kt
+++ b/compose/ui/ui/src/androidMain/kotlin/androidx/compose/ui/platform/GlobalSnapshotManager.android.kt
@@ -37,11 +37,12 @@ import java.util.concurrent.atomic.AtomicBoolean
  */
 internal object GlobalSnapshotManager {
     private val started = AtomicBoolean(false)
+    private val uiScope = CoroutineScope(AndroidUiDispatcher.Main)
 
     fun ensureStarted() {
         if (started.compareAndSet(false, true)) {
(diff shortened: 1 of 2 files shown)


dlam:
Just closing this thread - I think you and adam talked about this offline and found this change isn't necessary.

Thanks for thinking of us though!

LouisCAD:
It's not the culprit of another issue, but this is still a poor example of how to start a coroutine that can lead to very hard to diagnose issues.

Consequently, it'd be great if you still merge this, for the sake of the examples that future eyes will see when looking at Compose runtime codebase.

Apart for the fact that it won't change anything for the Compose runtime, are there any reasons to not merge this?

dlam:
@adamp is probably best to answer this.

adamp:
I'd prefer not. It expands the scope of access of the `uiScope` `val` beyond the function where it's used to the whole class, which would serve as its own kind of example. Assigning the scope to a `val` doesn't change the structured vs. unstructured properties of this code when the goal is launching a daemon that doesn't terminate until the process does.

PR #240, opened 2021-09-03
[XProcessor] Support Synthetic elements as an aggregating elements

elihart:
## Proposed Changes
I ran into a crash using XProcessing because I pass a KspSyntheticPropertyMethodElement as an aggregating element, which wasn't supported. It's possible to work around this and manage elements in my processor differently, but it is fairly simple to support by expanding the when statement to map XElement to originating element.

## Testing

Test: Added cases to OriginatingElementsTest

## Issues Fixed

Fixes: N/A


yigit (approved):
Thank you

PR #241, opened 2021-09-03
[XProcessing] Support all primitive types in XAnnotationBox

elihart:
## Proposed Changes
It had been left as a TODO to support most primitive types and primitive type arrays (besides Int) in the annotation box, and would result in a runtime crash if they were used. This implements them following the same pattern as the Int implementation.

## Testing

Test: Added tests to XAnnotationBoxTest for all primitive types and type arrays for both kotlin and java sources.

## Issues Fixed

Fixes: b/179081610


yigit (approved):
thank you!

PR #243, opened 2021-09-04
[XProcessing] Fix getAllAnnotations returning empty list

elihart:
## Proposed Changes
There is a bug in `XAnnotated.getAllAnnotations` where it returns empty if the annotation contains a value named `value` and is an array type. This is due to a bug in `unwrapRepeatedAnnotationsFromContainer` where if the array it tries to unwrap is empty it is incorrectly believed to be a repeated annotations array.

The fix is simple, in `unwrapRepeatedAnnotationsFromContainer` we need to check that the array is not empty in addition to the existing `all` check we do.

## Testing

Test: Added a test to XAnnotationTest that reproduces the issue and passes with the fix applied.

## Issues Fixed

Fixes: N/A


danysantiago (approved):
Thanks for fixing this!

PR #244, opened 2021-09-09
Upload test apk's as a separate artifact

yigit:
This PR fixes a problem where we stopped distributing test apks hence
stopped running FTL tests (aosp/1751630).

We will now re-use what aosp does to reduce chances of this breaking
(should probably add some failsafe, this PR does not do it).

For this to take affect, we also need to update androidx-ci-action to
look for the new artifact name.

Bug: n/a
Test: CI

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -16,10 +16,10 @@ jobs:
         run: |
           set -x
           GRADLEW_FLAGS="-Dorg.gradle.internal.http.connectionTimeout=60000 \
-            -Dorg.gradle.internal.http.socketTimeout=60000                   \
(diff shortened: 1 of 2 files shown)


yigit:
just trying on prod :) to see if the artifacts will show up properly

dlam:
Oh, I think you don't need `\` if we have an open `"` already.

.github/workflows/presubmit.yml
```diff
@@ -233,4 +233,5 @@ jobs:
         uses: actions/upload-artifact@v2
         with:
           name: artifacts_${{ env.artifact-id }}
-          path: ~/dist
+          path: |
```
dlam: Can we undo this change?

dlam (approved):

PR #245, opened 2021-09-20
Change playground rootproject name: navigation -> lifecycle

tatocaster:
## Proposed Changes

  - Fix the name of the rootProject. All these time `lifecycle` project has the name `navigation-playground` instead of `lifecycle-playground`, which caused incorrect project title in Android Studio as well.

## Testing
Bug: N/A
Test: local

### DIFF
--- a/lifecycle/settings.gradle
+++ b/lifecycle/settings.gradle
@@ -15,7 +15,7 @@
  */
 
 // see ../playground-common/README.md for details on how this works
-rootProject.name = "navigation-playground"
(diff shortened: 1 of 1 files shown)


yigit (approved):
thanks!

yigit:
btw, you need a bug/test stanza to pass presubmit.
Since it is not needed here, you can add to your message:

```
Bug: N/A
Test: local
```

tatocaster:
Thanks!

PR #246, opened 2021-09-22
Fix equals definition for LayoutWeightImpl

chris-horner:
## Proposed Changes

Fix the equals check in `LayoutWeightImpl`. It currently returns true if its `weight` and `fill` _don't_ equal the other modifier's properties.

## Testing

Bug: N/A
Test: local


mihaipopa12 (approved):
Thanks!

PR #249, opened 2021-10-04
Fix isolalation of nested mutable snapshots

b7woreo:
A nested mutable snapshot should be isolated from all changes made but currently will see changes made to objects between the time the parent was created and when the snapshot was taken.

This fixes that by considiering all snapshots between the parent old id and new nested mutable snapshot's id.

Fixes: b/200575924
Test: ./gradlew :compose:r:r:tDUT


b7woreo:
@googlebot I signed it!

PR #257, opened 2021-10-25
Delayed work requests should not be expedited.

tikurahul:
Fixes: b/203678406
Test: Added unit tests.

### DIFF
--- a/work/work-runtime/src/androidTest/java/androidx/work/WorkRequestTest.kt
+++ b/work/work-runtime/src/androidTest/java/androidx/work/WorkRequestTest.kt
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package androidx.work
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import androidx.test.filters.SmallTest
+import androidx.work.worker.TestWorker
+import org.junit.Assert.assertNotNull
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.util.concurrent.TimeUnit
+
+@RunWith(AndroidJUnit4::class)
+class WorkRequestTest {
+    @Test
+    @SmallTest
+    public fun expeditedRequest_withInitialDelay_throwsException() {
+        var error: Throwable? = null
+        try {
+            OneTimeWorkRequest.Builder(TestWorker::class.java)
+                .setInitialDelay(10, TimeUnit.SECONDS)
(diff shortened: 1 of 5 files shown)


work/work-runtime/src/main/java/androidx/work/impl/background/systemjob/SystemJobInfoConverter.java
```diff
@@ -125,7 +125,8 @@ JobInfo convert(WorkSpec workSpec, int jobId) {
         }
         // Retries cannot be expedited jobs, given they will occur at some point in the future.
         boolean isRetry = workSpec.runAttemptCount > 0;
-        if (BuildCompat.isAtLeastS() && workSpec.expedited && !isRetry) {
+        boolean isDelayed = offset > 0;
+        if (BuildCompat.isAtLeastS() && workSpec.expedited && !isRetry && !isDelayed) {
```
svasilinets, line 129: Could it be better to throw ourselves here an `IllegalStateException`? It seems like we shouldn't allow marking as expedited anything that has delay mechanism with it, so legally it shouldn't never be both and expedited and delayed?

Or are you concerned about already requested workers with incorrect state?

svasilinets (commented):
What about periodic jobs? It seems like they can hit this case as well, right?

work/work-runtime/src/main/java/androidx/work/impl/background/systemjob/SystemJobInfoConverter.java
tikurahul: > Or are you concerned about already requested workers with incorrect state?

Worried about existing workers here. New work requests will throw an ISE at `build` time, where the stack trace will make a lot more sense. Its kind of late to throw an `Exception` here, because this stuff is happening on the `SerialExecutor`.

tikurahul:
> What about periodic jobs? It seems like they can hit this case as well, right?

We never really schedule periodic jobs with `JobScheduler`, because of well known bugs. We always schedule one-time jobs (from the perspective of `JobScheduler`). Not sure if that answers your question.

svasilinets:
> We always schedule one-time jobs (from the perspective of JobScheduler)

What I'm saying that it seems to possible to write:

```
new PeriodicWorkRequest.Builder(...)
    .setExpedited(OutOfQuotaPolicy.DROP_WORK_REQUEST)
    .build()
```

and `SystemJobInfoConverter` will convert `Workspec` for this `PeriodicWorkRequest` to a `JobInfo` that has both `minimumLatency` and `expedited = true`, which is illegal. 

It seems like we shouldn't allow `setExpedited` on periodic work requests.

tikurahul:
We can be more explicit about the error for `PeriodicWorkRequest`s. However, currently `SystemJobConverter` won't actually mark it expedited given that periodic work has a built in offset.

work/work-runtime/src/androidTest/java/androidx/work/WorkRequestTest.kt
```diff
@@ -0,0 +1,59 @@
...
+    @Test
+    @SmallTest
+    public fun expeditedRequest_withInitialDelay_throwsException() {
+        var error: Throwable? = null
+        try {
+            OneTimeWorkRequest.Builder(TestWorker::class.java)
+                .setInitialDelay(10, TimeUnit.SECONDS)
+                .setExpedited(OutOfQuotaPolicy.RUN_AS_NON_EXPEDITED_WORK_REQUEST)
+                .build()
+        } catch (exception: Throwable) {
```
svasilinets, line 38: nit: you probably can be a bit more specific about exception, because now other unrelated errors like NPE can be caught and pass the test

svasilinets (approved):

PR #258, opened 2021-10-27
Detect Flow::launchIn in ComposableCoroutineCreationDetector

DSteve595:
## Proposed Changes

Compose's `ComposableCoroutineCreationDetector` currently detects disallowed usages of `CoroutineScope::launch` and `CoroutineScope::async` inside composable functions. Another common way to launch a coroutine is `Flow::launchIn`, but the detector doesn't catch it. This adds `launchIn` to the detector.

## Testing

Test: The detector's test code now includes usages of `launchIn`. 
I feel that `ComposableCoroutineCreationDetectorTest` is a bit hard to scale up, as line numbers in the expected output will have to be changed after pretty much any change to the test code. I think this test could be improved by breaking it up into smaller tests for each of the cases (plain function, lambda, lambda parameter, etc). I figure this is out of the scope of the PR, but if a maintainer agrees, I'm happy to do that in another PR.

### DIFF
--- a/compose/runtime/runtime-lint/src/main/java/androidx/compose/runtime/lint/ComposableCoroutineCreationDetector.kt
+++ b/compose/runtime/runtime-lint/src/main/java/androidx/compose/runtime/lint/ComposableCoroutineCreationDetector.kt
@@ -39,10 +39,16 @@ import java.util.EnumSet
  * body of a composable function / lambda.
  */
 class ComposableCoroutineCreationDetector : Detector(), SourceCodeScanner {
-    override fun getApplicableMethodNames() = listOf(Async.shortName, Launch.shortName)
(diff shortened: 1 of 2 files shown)


lelandrichardson:
This all looks good to me! but i'm going to let the original author of the lint rule take a look before merging.

Also, i agree with you on the testing front. I wouldn't have any complaints if you wanted to take on updating the tests to be a little bit more independent!

Thanks for the contribution!

DSteve595:
bump

DSteve595:
bump

dlam (approved):
Approving on behalf of lpf :D

Sorry for the delay here.

dlam (approved):

PR #259, opened 2021-10-28
Avoid running internal workflows in forked repos

bigdaz:
## Proposed Changes

The "Register Workflow Run with AndroidX" and "Run Integration Tests" workflows require a secret that is not available when running workflows in a forked repository.
This change disables these workflows on forked repositories allowing them to show as "Skipped".

Note that it is not possible to check for the presence of the required secrets when deciding if a _job_ can be skipped. This is only possible when skipping a _step_ within the job. Skipping individual step causes the entire workflow to show up as successful, rather than skipped, which is misleading in this case.

## Testing

Test: Change was tested manually on my fork, and the appropriate workflows were skipped.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -33,7 +33,7 @@ jobs:
   teardown:
     runs-on: ubuntu-latest
     needs: [run_integration_tests]
-    if: ${{ !github.event.repository.fork && always() }}
+    if: ${{ !github.event.repository.fork && always() }} # always() is required so that this job will run irrespective of status of upstream 'run_integration_test' job.
     steps:
       - name: Parse workflow status
         id: workflow-status
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   run_integration_tests:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -33,7 +33,7 @@ jobs:
   teardown:
     runs-on: ubuntu-latest
     needs: [run_integration_tests]
-    if: always()
+    if: ${{ !github.event.repository.fork && always() }}
```
dlam: Could you add a comment here that the `&& always()` part is needed even though we override the default because `success()` is always implied unless you explicitly check for `always()`?

bigdaz: Sure thing. I could also replace `always()` with `success() || failure()` and then we wouldn't need the repository check, since the upstream job would be neither of these on a fork.

So we'd have `if: success() || failure()` instead.

dlam: Keeping it as `if: ${{ !github.event.repository.fork && always() }}` with a comment seems better since it's a bit more explicit in case we end up reworking these hooks in the future

bigdaz: I realize that we also need to run this job when `run_integration_tests` is `cancelled()`, so the forked-repo check is still required.

Leaving as `always()`, and added the the comment as requested.

dlam (approved):

PR #260, opened 2021-11-03
Avoid duplicate project dirs in playground

yigit:
This CL fixes a bug in playground where we might end up with duplicate
project directories when Gradle automatically creates a parent project
when it does not exist. When that case happens, Gradle will pick a
project directory relative to the playground root project. Instead, we
provide a project directory relative to the child project to ensure
consistency.

With this change, we can also enable to the project path validation in
the AndroidXPlugin.

Bug: 197253160
Test: cd activity && ./gradlew tasks // fails without this change

### DIFF
--- a/buildSrc/private/src/main/kotlin/androidx/build/AndroidXImplPlugin.kt
+++ b/buildSrc/private/src/main/kotlin/androidx/build/AndroidXImplPlugin.kt
@@ -874,10 +874,7 @@ private const val GROUP_PREFIX = "androidx."
  * Validates the project structure against Jetpack guidelines.
  */
 fun Project.validateProjectStructure(groupId: String) {
-    // TODO(b/197253160): Re-enable this check for playground. For unknown reasons in playground
-    //  builds, automatically generated parent projects such as :activity are incorrectly
-    //  inheriting their children's build file which causes the AndroidXPlugin to get applied.
-    if (studioType() == StudioType.PLAYGROUND || !project.isValidateProjectStructureEnabled()) {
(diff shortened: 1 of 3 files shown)


dlam:
Are we guaranteed to move the project dir for the generated projects and not actual projects here? It's important that we only move generated projects to a fake dir because the projectDir property is used for real builds (such as default location for build.gradle)

yigit:
> Are we guaranteed to move the project dir for the generated projects and not actual projects here? It's important that we only move generated projects to a fake dir because the projectDir property is used for real builds (such as default location for build.gradle)

So this actually already failed by including more than we want. Not sure why it works this way but it seems like calling `include(:a)` then `include(:a:b)` is different than calling `include(:a:b)`. Looking at the source code, i can't see why: https://github.com/gradle/gradle/blob/7c58f21fd7fc7d1accb7c204ce92d7fa2f610284/subprojects/core/src/main/java/org/gradle/initialization/DefaultSettings.java#L156
But it did fail in navigation where it tried to include `:compose:integration-tests:demos` whereas it was not doing it before. I can work around that by setting the gradle file to an ignored one.

**UPDATE** I figured it was working before because we never set the `projectDir` for a project that we didn't include explicitly, hence gradle was unable to find the build gradle file. So it kind of worked by chance :facepalm: 
I can also change this to automatically ignore all parent projects of the newly created project (change their project directory). They'll simply never work since their project dirs are not set. Technically, gradle expects them to be proper but i think, given the use case in navigation, we really want those projects to not work.

dlam (approved):

PR #261, opened 2021-11-03
Set up Gradle enterprise for playground projects

liutikas:
- Apply gradle enterprise settings plugin in every playground
settings.gradle file
- Configure to upload build scans when GRADLE_ENTERPRISE_ACCESS_KEY
  is set
- Obfuscate username, hostname, and IP

Test: GRADLE_ENTERPRISE_ACCESS_KEY=SOMEKEYHERE ./gradlew tasks

### DIFF
--- a/activity/settings.gradle
+++ b/activity/settings.gradle
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+plugins {
+    id "com.gradle.enterprise" version "3.7.1"
+}
+
 // see ../playground-common/README.md for details on how this works
 rootProject.name = "activity-playground"
 apply from: "../playground-common/playground-include-settings.gradle"
--- a/playground-common/playground-include-settings.gradle
+++ b/playground-common/playground-include-settings.gradle
@@ -72,6 +72,22 @@ def setupPlayground(settings, String relativePathToRoot) {
             new File(supportRoot, "testutils/testutils-common"))
     settings.includeProject(":internal-testutils-gradle-plugin",
             new File(supportRoot, "testutils/testutils-gradle-plugin"))
+
+    settings.gradleEnterprise {
+        server = "https://ge.androidx.dev"
+        buildScan {
+            publishAlways()
+            publishIfAuthenticated()
+            capture {
+                taskInputFiles = true
+            }
+            obfuscation {
+                username { name -> "unset" }
+                hostname { host -> "unset" }
+                ipAddresses { addresses -> addresses.collect { address -> "0.0.0.0"} }
+            }
+        }
+    }
 }
 
 /**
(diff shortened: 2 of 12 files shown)


dlam (approved):

playground-common/playground-include-settings.gradle
```diff
@@ -72,6 +72,21 @@ def setupPlayground(settings, String relativePathToRoot) {
             new File(supportRoot, "testutils/testutils-common"))
     settings.includeProject(":internal-testutils-gradle-plugin",
             new File(supportRoot, "testutils/testutils-gradle-plugin"))
+
+    settings.gradleEnterprise {
+        server = "https://ge.androidx.dev"
+        buildScan {
+            publishAlwaysIf(System.getenv("GRADLE_ENTERPRISE_ACCESS_KEY") != null)
```
bigdaz: You can accomplish the same thing with:
```
publishAlways()
publishIfAuthenticated()
```
Then this will work for other forms of authentication, such as https://docs.gradle.com/enterprise/gradle-plugin/#automated_access_key_provisioning

liutikas: I don't see documentation for this neither or https://docs.gradle.com/enterprise/gradle-plugin/#automated_access_key_provisioning nor on https://docs.gradle.com/enterprise/gradle-plugin/api/com/gradle/scan/plugin/BuildScanExtension.html . Can you point me to the docs?

activity/settings.gradle
```diff
@@ -14,6 +14,10 @@
  * limitations under the License.
  */
 
+plugins {
+    id "com.gradle.enterprise" version "3.7.1"
+}
```
bigdaz, line 19: It would be great to add
```
id 'com.gradle.common-custom-user-data-gradle-plugin" version "1.4.2"
```
as well, since this will enhance the builds scans with a bunch of tags and custom values. For example:
- link to all build scans with the same commit id
- tag for CI vs LOCAL (to make these easy to filter)
- link to all build scans from the same workflow or run

playground-common/playground-include-settings.gradle
bigdaz: Unfortunately `publishIfAuthenticated()` is currently not documented (a hidden feature), since we weren't overly happy with the API. You can see it in use on [JUnit](https://github.com/junit-team/junit5/blob/main/settings.gradle.kts#L52-L56) and [Spring](https://github.com/spring-io/gradle-enterprise-conventions/blob/7d25b99352bdaa2e8b818fd97edf18e7b8df30b3/src/main/java/io/spring/ge/conventions/gradle/BuildScanConventions.java#L84).

Fortunately groovy lets you call the method without any reflection hoops.

PR #267, opened 2021-11-05
Pass GE secrets to bOS

dlam:
Test: Merge change on separate branch and manually check scans show up
Change-Id: I190b1142bd33f66e20a76faf560dfd9d8014cdfb

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -141,6 +141,8 @@ jobs:
     env:
       artifact-id: ${{matrix.project}}
       project-root: ${{matrix.project-root || matrix.project}}
+      GRADLE_BUILD_CACHE_PASSWORD: ${{ secrets.GRADLE_BUILD_CACHE_PASSWORD }}
+      GRADLE_ENTERPRISE_ACCESS_KEY: ${{ secrets.GRADLE_ENTERPRISE_ACCESS_KEY }}
     steps:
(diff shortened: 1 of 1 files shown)


yigit (approved):

liutikas:
Something is not right, scans are not uploading
```
A build scan was not published as you have not authenticated with server 'ge.androidx.dev'.
```

and it is uploading to build cache
```
> Task :annotation:annotation-experimental:mergeReleaseJniLibFolders
Could not store entry ee9ad8e683f476a4f501bfb237a57b86 in remote build cache
org.gradle.api.UncheckedIOException: Storing entry at 'https://ge.androidx.dev/cache/ee9ad8e683f476a4f501bfb237a57b86' response status 401: Unauthorized
```

dlam:
It won't run in this check because it needs to be a push event on a branch where this change is already merged. I'm testing on `develop` but looks like the test change I pushed didn't actually change any source so all the builds were skipped 🤦 

Hopefully this change will run room playground: https://github.com/androidx/androidx/actions/runs/1424448508

The other idea I have is that I'm unsure about scope of setting the env variable for the secret, we may want to keep it to just the interested steps themselves

dlam:
Hopefully this change will run room playground: https://github.com/androidx/androidx/actions/runs/1424448508

liutikas:
It works!

liutikas (approved):

PR #269, opened 2021-11-05
Enable rich gradle console

liutikas:
This should also enable github actions to show
colored output.

Test: ./gradlew tasks -> now colored


liutikas:
Turns out rich console makes it all messy

PR #271, opened 2021-11-06
Reorder repositories to optimize download speed

liutikas:
Prefer using external maven repos over androidx.dev to
speed up dependency download speed.

Test: It still builds locally

### DIFF
--- a/buildSrc/private/src/main/kotlin/androidx/build/AndroidXPlaygroundRootImplPlugin.kt
+++ b/buildSrc/private/src/main/kotlin/androidx/build/AndroidXPlaygroundRootImplPlugin.kt
@@ -174,21 +174,31 @@ class AndroidXPlaygroundRootImplPlugin : Plugin<Project> {
     }
 
     private fun RepositoryHandler.addPlaygroundRepositories() {
-        repos.all.forEach { playgroundRepository ->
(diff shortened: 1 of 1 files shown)


buildSrc/private/src/main/kotlin/androidx/build/AndroidXPlaygroundRootImplPlugin.kt
```diff
@@ -174,21 +174,30 @@ class AndroidXPlaygroundRootImplPlugin : Plugin<Project> {
     }
 
     private fun RepositoryHandler.addPlaygroundRepositories() {
-        repos.all.forEach { playgroundRepository ->
+        fun RepositoryHandler.addCustomRepo(customRepository: PlaygroundRepository) {
```
dlam: Odd naming in terms of customRepo vs PlaygroundRepo, maybe lets just align the names?

dlam (approved):

buildSrc/private/src/main/kotlin/androidx/build/AndroidXPlaygroundRootImplPlugin.kt
liutikas: Updated

PR #272, opened 2021-11-07
Fix ProcessCompat.isApplicationUid always returns true

yujincheng08:
## Proposed Changes

  - Fix `ProcessCompat.isApplicationUid` always returns true when 17 <= SDK_INT <= 23

## Testing

Test: Test `ProcessCompat.isApplicationUid(1000)` on older AVD which should return false.

Co-authored-by: canyie <a1364259@163.com>

### DIFF
--- a/core/core/src/main/java/androidx/core/os/ProcessCompat.java
+++ b/core/core/src/main/java/androidx/core/os/ProcessCompat.java
@@ -106,6 +106,7 @@ public final class ProcessCompat {
                         // This should never happen, as the method returns a boolean primitive.
                         throw new NullPointerException();
                     }
+                    return result;
                 }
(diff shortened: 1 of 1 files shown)


yigit:
Thanks for the PR. Can you add a regression test so this won't fail again?

yujincheng08:
@yigit I don't think we could deploy a unit test since this function can `return true` as a fallback, and we cannot ensure it returns false when `uid=1000`.

dlam:
You'll also need to accept the CLA for us to consider this change - although the general rule is that every change needs a test as well.

yujincheng08:
I have signed. But it needs maintainer to recheck.

PR #277, opened 2021-11-12
Enable host side tests in GitHub actions

liutikas:
- Removing setting ANDROID_SDK_ROOT, it is hardcode to a non existent path

Test: tested using https://github.com/androidx/androidx/actions/runs/1451151278

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -62,7 +62,7 @@ jobs:
         shell: bash
         run: |
           set -x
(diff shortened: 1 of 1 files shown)


dlam (approved):

dlam:
I think both ANDROID_SDK_ROOT and ANDROID_HOME are set by the runner so we may be able to just remove that line entirely?

liutikas:
> I think both ANDROID_SDK_ROOT and ANDROID_HOME are set by the runner so we may be able to just remove that line entirely?

Done

PR #280, opened 2021-11-15
Add `compose-ui` playground.

tikurahul:
* Added support for NDK builds.

Test: `./gradlew test`


.github/workflows/presubmit.yml
```diff
@@ -62,6 +62,7 @@ jobs:
         shell: bash
         run: |
           set -x
+          echo "ANDROID_NDK_HOME=$ANDROID_NDK_LATEST_HOME" >> $GITHUB_ENV
```
liutikas, line 66: Please add a comment that we need NDK 23 and that default ANDROID_NDK_HOME is 21.

https://github.com/actions/virtual-environments/blob/4f8049546ee84b94b4fe947bab7fa9558b34b195/images/linux/Ubuntu2004-README.md#environment-variables-3

liutikas (commented):
Should we use project() for all the dependencies that we know are within compose playground as discussed in chat?

.github/workflows/presubmit.yml
tikurahul: Done.

tikurahul:
Used `project()` consistently.

PR #287, opened 2021-11-24
Re-enable compose-compiler playground

dlam:
Test: cd compose/compiler && ./gradlew bOS
Change-Id: Ic9f98028d47a74d0a4a1653f16ddcdb0f1530177


liutikas:
It looks like the tests are failing

PR #290, opened 2021-11-25
Add invalid Dao method lint check

mzgreen:
Proposed Changes
This change adds room-lint module containing a lint check which detects an invalid Dao method. In this case an invalid Dao method is a method which is suspending and returns one of RxJava2 or RxJava3 types at the same time.

Testing
Test: Added InvalidDaoMethodTest

Issues Fixed
Fixes: https://issuetracker.google.com/issues/151029087

### DIFF
--- a/room/room-lint/build.gradle
+++ b/room/room-lint/build.gradle
@@ -1,0 +1,41 @@
+/*
+ * Copyright 2021 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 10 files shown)


yigit:
Thanks for the suggestion.
There is definitely good value in detecting this (and maybe more) but I'm not sure if we should do this in the compiler rather than a lint.
Lint has better response time(ide) but its API is not very stable and it might be trickier to stay consistent with the compiler.

mzgreen:
As we discussed, it would be better to improve the compiler error messages rather than relying on lint. I've created another PR: https://github.com/androidx/androidx/pull/299 so I'm closing this one.

PR #293, opened 2021-12-01
Fix cache misses, warnings and deprecations caused by SdkResourceGenerator

bigdaz:
#### Ignore generated `sdk.prop` file when composing a cache key with `runtimeClasspath` normalization. 

This generated file contains absolute path references, resulting in cache misses for tasks that have this resource as an input. Since differences in this file are not significant for caching purposes the file is ignored for runtimeClasspath normalization.

#### Improve the way generated resources are registered
By registering the generated resources as a separate SourceSetOutput and including a `builtBy` reference, Gradle is able to automatically maintain the required task dependencies for any consumers of these resources. This mechanism is recommended at https://docs.gradle.org/current/dsl/org.gradle.api.tasks.SourceSetOutput.html.

This fix removes a Gradle warning about "execution optimizations are disabled" due to undeclared task dependencies with the `SdkResourceGenerator` task, as well as the associated deprecation warning.

## Testing

Test: This is a build refactor only - no new tests added

### DIFF
--- a/buildSrc/public/src/main/kotlin/androidx/build/SdkResourceGenerator.kt
+++ b/buildSrc/public/src/main/kotlin/androidx/build/SdkResourceGenerator.kt
@@ -137,6 +137,13 @@ abstract class SdkResourceGenerator : DefaultTask() {
             val extension = project.extensions.getByType<JavaPluginExtension>()
             val resources = extension.sourceSets.getByName("test").resources
             resources.srcDirs(setOf(resources.srcDirs, generatedDirectory))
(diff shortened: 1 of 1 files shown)


dlam (approved):

dlam:
Forwarding feedback from @liutikas:

>  Aurimas Liutikas
I don't think this is a safe change. We should instead be using relative paths. If we ever make a change where one of our dependencies references sdk.prop, these tests will continue to be cached even though the values have changed.

PR #294, opened 2021-12-01
Increase memory limit of gradle daemon for Room tests

dlam:
Test: GH workflows
Change-Id: I98c1bba88b8ed8c7f728d89f0895953397cfe073


dlam:
This seems to improve the rate at which gradle daemon survives, although I have no real concrete proof and not sure I have time to really look into this..

liutikas (approved):

PR #297, opened 2021-12-05
Fix ProcessCompat.isApplicationUid always returns true 

yujincheng08:
## Proposed Changes

  - Fix `ProcessCompat.isApplicationUid` always returns true when 17 <= SDK_INT <= 23

## Testing

Test: Test `ProcessCompat.isApplicationUid(1000)` on older AVD which should return false.

Co-authored-by: canyie <a1364259@163.com>


yujincheng08:
Merged here: https://android-review.googlesource.com/#/c/1908997

PR #299, opened 2021-12-07
Improve invalid Dao method errors

mzgreen:
Proposed Changes
This change improves invalid Dao method error messages. In this case an invalid Dao method is a method which is suspending and returns one of the following types:

- any of RxJava2 types
- any of RxJava3 types
- LiveData
- ComputableLiveData
- ListenableFuture
- ReactiveStreams Publisher
- Kotlin Flow

Testing
Test: Added unit tests for Update, Insertion, Deletion and Query methods

Issues Fixed
Fixes: https://issuetracker.google.com/issues/151029087

### DIFF
--- a/room/room-compiler/src/main/kotlin/androidx/room/processor/DeletionMethodProcessor.kt
+++ b/room/room-compiler/src/main/kotlin/androidx/room/processor/DeletionMethodProcessor.kt
@@ -34,12 +34,6 @@ class DeletionMethodProcessor(
 
         val returnType = delegate.extractReturnType()
 
-        context.checker.check(
-            !delegate.isUsingMultipleConcurrencyPatterns(),
-            executableElement,
-            ProcessorErrors.USING_MULTIPLE_CONCURRENCY_PATTERNS
-        )
-
         val methodBinder = delegate.findDeleteOrUpdateMethodBinder(returnType)
 
         context.checker.check(
--- a/room/room-compiler/src/main/kotlin/androidx/room/processor/DeletionMethodProcessor.kt
+++ b/room/room-compiler/src/main/kotlin/androidx/room/processor/DeletionMethodProcessor.kt
@@ -34,6 +34,12 @@ class DeletionMethodProcessor(
 
         val returnType = delegate.extractReturnType()
 
+        context.checker.check(
+            !delegate.isUsingMultipleConcurrencyPatterns(),
+            executableElement,
+            ProcessorErrors.USING_MULTIPLE_CONCURRENCY_PATTERNS
+        )
+
         val methodBinder = delegate.findDeleteOrUpdateMethodBinder(returnType)
 
         context.checker.check(
--- a/room/room-compiler/src/main/kotlin/androidx/room/processor/MethodProcessorDelegate.kt
+++ b/room/room-compiler/src/main/kotlin/androidx/room/processor/MethodProcessorDelegate.kt
@@ -22,15 +22,11 @@ import androidx.room.compiler.processing.XSuspendMethodType
 import androidx.room.compiler.processing.XType
 import androidx.room.compiler.processing.XVariableElement
 import androidx.room.compiler.processing.isSuspendFunction
-import androidx.room.ext.GuavaUtilConcurrentTypeNames
+import androidx.room.ext.DEFERRED_TYPES
 import androidx.room.ext.KotlinTypeNames
 import androidx.room.ext.L
-import androidx.room.ext.LifecyclesTypeNames
 import androidx.room.ext.N
-import androidx.room.ext.ReactiveStreamsTypeNames
 import androidx.room.ext.RoomCoroutinesTypeNames
-import androidx.room.ext.RxJava2TypeNames
-import androidx.room.ext.RxJava3TypeNames
 import androidx.room.ext.T
 import androidx.room.parser.ParsedQuery
 import androidx.room.solver.TypeAdapterExtras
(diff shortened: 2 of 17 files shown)


mzgreen:
Not sure why `buildOnServer zipTestConfigsWithApks` failed. I ran `buildOnServer` locally before pushing and it passed. I don't see an option to retry the build either.

room/room-compiler/src/main/kotlin/androidx/room/processor/MethodProcessorDelegate.kt
```diff
@@ -71,6 +76,31 @@ abstract class MethodProcessorDelegate(
             ).process()
         }
     }
+    fun isUsingMultipleConcurrencyPatterns(): Boolean {
+        val concurrencyTypes = listOf(
```
danysantiago: We have a list just like this in [TransactionMethodProcessor](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:room/room-compiler/src/main/kotlin/androidx/room/processor/TransactionMethodProcessor.kt;l=86-99), lets try and use it, likely moving it to a more accessible location, maybe in `javapoet_ext.kt`

room/room-compiler/src/main/kotlin/androidx/room/processor/MethodProcessorDelegate.kt
```diff
@@ -71,6 +76,31 @@ abstract class MethodProcessorDelegate(
             ).process()
         }
     }
+    fun isUsingMultipleConcurrencyPatterns(): Boolean {
```
danysantiago: I would move this to a top-level extension function on `MethodProcessorDelegate`, the first check should be for suspend function and return early and then go through the list doing `isAssignableFrom`. If we can return early we should because finding types and checking for assignability is expensive.

room/room-compiler/src/main/kotlin/androidx/room/processor/DeletionMethodProcessor.kt
```diff
@@ -34,6 +34,12 @@ class DeletionMethodProcessor(
 
         val returnType = delegate.extractReturnType()
 
+        context.checker.check(
+            !delegate.isUsingMultipleConcurrencyPatterns(),
+            executableElement,
+            ProcessorErrors.USING_MULTIPLE_CONCURRENCY_PATTERNS
+        )
```
danysantiago: You can move this and all the other `check` calls into the `ShortcutMethodProcessor` since they all need it.
Before returning from `extractReturnType()`.

```
// In ShortcutMethodProcessor
fun extractReturnType(): XType {
    context.checker.check(
        ...
    )
    return delegate.extractReturnType()
}
```

room/room-compiler/src/main/kotlin/androidx/room/processor/ProcessorErrors.kt
```diff
@@ -175,6 +175,10 @@ object ProcessorErrors {
     val CANNOT_FIND_UPDATE_RESULT_ADAPTER = "Not sure how to handle update method's " +
         "return type. Currently the supported return types are void, int or Int."
 
+    val USING_MULTIPLE_CONCURRENCY_PATTERNS = "Dao methods that have a suspend modifier " +
```
danysantiago: It would be nice, similar to `transactionMethodAsync` to include the conflicting return type. It is especially useful for KAPT since the stub will have a weird signature (`public Object mySuspendQuery(Continuation<? extends LiveData<MyEntity>> cont)`) and might be hard to parse for the user, but by proving the return type and the 'suspend' part of the message they can link things. We also use the word 'deferred' and 'async' through the documentation and other errors, so would be good to include that too for consistency.

danysantiago: `methods` -> `functions`

danysantiago (commented):
Thanks for sending this contribution!

room/room-compiler/src/main/kotlin/androidx/room/processor/MethodProcessorDelegate.kt
mzgreen: Didn't know about it! This list is not complete though. It's missing `LifecyclesTypeNames.COMPUTABLE_LIVE_DATA` and `ReactiveStreamsTypeNames.PUBLISHER`. Can I add them or they aren't there for a reason and I should ignore them?

danysantiago: Yeah, you can totally add them, seems like we just missed them.

mzgreen: Done

room/room-compiler/src/main/kotlin/androidx/room/processor/MethodProcessorDelegate.kt
mzgreen: Done

room/room-compiler/src/main/kotlin/androidx/room/processor/DeletionMethodProcessor.kt
mzgreen: Done

room/room-compiler/src/main/kotlin/androidx/room/processor/ProcessorErrors.kt
mzgreen: Done

mzgreen: Done

room/room-compiler/src/main/kotlin/androidx/room/processor/ProcessorErrors.kt
```diff
@@ -175,9 +175,9 @@ object ProcessorErrors {
     val CANNOT_FIND_UPDATE_RESULT_ADAPTER = "Not sure how to handle update method's " +
         "return type. Currently the supported return types are void, int or Int."
 
-    val USING_MULTIPLE_CONCURRENCY_PATTERNS = "Dao methods that have a suspend modifier " +
-        "shouldn't return another async type. Most probably this is an error. Consider " +
-        "changing the return type or removing the suspend modifier."
+    fun suspendReturnsDeferredType(returnTypeName: String) = "Dao functions that have a suspend " +
+        "modifier must not return deferred/async return type $returnTypeName. Most probably this " +
```
danysantiago: nit: extra 'return' here

Maybe change to `must not contain a deferred/async return type`
or `must not return a deferred/async type`

and then move the input name to a parenthesis `($returnTypeName)`

danysantiago (commented):
Looking good, just one last small comment!

room/room-compiler/src/main/kotlin/androidx/room/processor/ProcessorErrors.kt
mzgreen: Done, thank you for the review Dany!

danysantiago (approved):

PR #308, opened 2022-01-14
Add lifecycle to activity

yigit:
Activity and lifecycle has close relationship so
it is best to keep lifecycle as a project for activity
playground setup.

Bug: n/a
Test: fixes activity build


dlam:
iirc @ianhanniballake preferred to keep them separate

liutikas (approved):

yigit:
given that lifecycle comes with other stuff. may not be a bad idea for local . 
Still i think it is worth checking what we need to make this pass, then we can decide whether to do it or not.

PR #310, opened 2022-01-17
Extended integrity error message

nkpodolyan:
Extended integrity error message with db identity hashes.

## Proposed Changes

  - Currently it is hard to tell by the data integrity exception message what database versions are being compared. With mentioned hashes it would become easy for code consumers to understand what have changed and what db versions are conflicting.

## Testing

### DIFF
--- a/room/room-runtime/src/main/java/androidx/room/RoomOpenHelper.java
+++ b/room/room-runtime/src/main/java/androidx/room/RoomOpenHelper.java
@@ -153,7 +153,8 @@ public class RoomOpenHelper extends SupportSQLiteOpenHelper.Callback {
             if (!mIdentityHash.equals(identityHash) && !mLegacyHash.equals(identityHash)) {
                 throw new IllegalStateException("Room cannot verify the data integrity. Looks like"
                         + " you've changed schema but forgot to update the version number. You can"
-                        + " simply fix this by increasing the version number.");
+                        + " simply fix this by increasing the version number. Expected identity"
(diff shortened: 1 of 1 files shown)


room/room-runtime/src/main/java/androidx/room/RoomOpenHelper.java
```diff
@@ -153,7 +153,8 @@ private void checkIdentity(SupportSQLiteDatabase db) {
             if (!mIdentityHash.equals(identityHash) && !mLegacyHash.equals(identityHash)) {
                 throw new IllegalStateException("Room cannot verify the data integrity. Looks like"
                         + " you've changed schema but forgot to update the version number. You can"
-                        + " simply fix this by increasing the version number.");
+                        + " simply fix this by increasing the version number. Current identity: "
+                        + identityHash + ", old: " + mIdentityHash + ", legacy: " + mLegacyHash);
```
danysantiago: I would remove the `legacy` hash since it is not helpful for those in Room 2.x+ and its flawed, so might add to the confusion.

Moreover, the words 'current' and 'old' are a bit weird here depending on how it's interpreted, for me the hashes are more like 'expected' and 'found'. Where the expected hash is based on your current analyzed schema, while the found one is the one in the database.

nkpodolyan: Great point. Addressed.

danysantiago (approved):
LGTM

PR #318, opened 2022-01-28
Update `presubmit` workflow to `gradle-build-action@v2`

bigdaz:
## Proposed Changes

  - Replace use of `gradle-command-action@v1` with `gradle-build-action@v2`.
  - Only write to the cache from the `androidx-main` branch. See https://github.com/gradle/gradle-build-action#only-write-to-the-cache-from-the-default-branch.
  - Enable build-cache and build-scans for `lint` workflow job
  - Disable Gradle configuration-cache due to errors reusing persisted state.
  - Remove use of `findAffectedModules` for build avoidance in `build-modules` job. This was causing failures to be hidden in subsequent workflow runs.
  - Add a single link to all build scans for a workflow run.

## Testing

Test: via GitHub Actions run

### DIFF
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -45,6 +45,8 @@ jobs:
           fi
           echo "::set-output name=repository::$REPOSITORY"
       - name: Publish build scans link
+        # No scans are produced for PRs from forked repos, so omit this notice for forked PRs.
+        if: ${{ !(github.event.pull_request && github.event.pull_request.head.repo.fork) }}
         run: echo "::notice title=All build scans for workflow::https://ge.androidx.dev/scans?search.names=CI%20run&search.values=$GITHUB_RUN_ID"
 
   lint:
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -108,7 +108,10 @@ jobs:
       - name: "Setup Gradle"
         uses: gradle/gradle-build-action@v2
         with:
-          # Don't share cache entries with other jobs          
+          # Only save Gradle User Home state for builds on the 'androidx-main' branch.
+          # Builds on other branches will only read existing entries from the cache.
+          cache-read-only: ${{ github.ref != 'refs/heads/androidx-main' }}
+          # Don't share cache entries with other jobs
           gradle-home-cache-strict-match: true
           # Limit the size of the cache entry
           gradle-home-cache-excludes: | 
@@ -175,7 +178,10 @@ jobs:
       - name: "Setup Gradle"
         uses: gradle/gradle-build-action@v2
         with:
-          # Don't share cache entries with other jobs          
+          # Only save Gradle User Home state for builds on the 'androidx-main' branch.
+          # Builds on other branches will only read existing entries from the cache.
+          cache-read-only: ${{ github.ref != 'refs/heads/androidx-main' }}
+          # Don't share cache entries with other jobs
           gradle-home-cache-strict-match: true
           # Limit the size of the cache entry
           gradle-home-cache-excludes: | 
--- a/.github/workflows/presubmit.yml
+++ b/.github/workflows/presubmit.yml
@@ -3,6 +3,9 @@ on:
   push:
   pull_request:
 
+env:
+  GRADLE_BUILD_ACTION_CACHE_DEBUG_ENABLED: true
+
 jobs:
   setup:
     runs-on: ubuntu-latest
@@ -19,6 +22,7 @@ jobs:
             -Dorg.gradle.internal.http.socketTimeout=60000                  \
             -Dorg.gradle.internal.repository.max.retries=20                 \
             -Dorg.gradle.internal.repository.initial.backoff=500            \
+            --no-configuration-cache                                        \
             --stacktrace"
           echo "::set-output name=gradlew_flags::$GRADLEW_FLAGS"
       - name: "Compute actions/checkout arguments"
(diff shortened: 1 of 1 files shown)


.github/workflows/presubmit.yml
```diff
@@ -3,6 +3,9 @@ on:
   push:
   pull_request:
 
+env:
+  GRADLE_BUILD_ACTION_CACHE_DEBUG_ENABLED: true
```
liutikas, line 9: Please add a comment on what this does with a link to docs https://github.com/marketplace/actions/gradle-build-action#cache-debugging-and-analysis

.github/workflows/presubmit.yml
```diff
@@ -37,13 +41,18 @@ jobs:
             REPOSITORY=${{ github.repository }}
           fi
           echo "::set-output name=repository::$REPOSITORY"
+      - name: Publish build scans link
+        run: echo "::notice title=All build scans for workflow::https://ge.androidx.dev/scans?search.names=CI%20run&search.values=$GITHUB_RUN_ID"
```
liutikas, line 50: Oh this is nice!

.github/workflows/presubmit.yml
```diff
@@ -94,35 +103,26 @@ jobs:
...
-          wrapper-directory: activity/gradle/wrapper
-          distributions-cache-enabled: true
+          # Only save Gradle User Home state for builds on the 'androidx-main' branch.
+          # Builds on other branches will only read existing entries from the cache.
+          cache-read-only: ${{ github.ref != 'refs/heads/androidx-main' }}
+          # Don't share cache entries with other jobs
+          gradle-home-cache-strict-match: true
+          # Limit the size of the cache entry
+          gradle-home-cache-excludes: | 
+            caches/jars-9
```
liutikas, line 126: Please add comments on what is in these directories so we know what we are dropping. Ideally link to Gradle docs.

.github/workflows/presubmit.yml
```diff
@@ -19,6 +22,7 @@ jobs:
             -Dorg.gradle.internal.http.socketTimeout=60000                  \
             -Dorg.gradle.internal.repository.max.retries=20                 \
             -Dorg.gradle.internal.repository.initial.backoff=500            \
+            --no-configuration-cache                                        \
```
liutikas, line 28: Are we disabling configuration to speed up CI?

liutikas (approved):

liutikas:
You will need to rebase this PR since I removed AffectedModuleDetector.

.github/workflows/presubmit.yml
bigdaz: There are 2 reasons this needs to be disabled:
1. Currently, the `buildOnServer` task fails when restored from the configuration cache. This happens when I run the build locally: https://ge.androidx.dev/s/zist7pgwj3qfy/console-log#L303
2. The configuration-cache state contains links directly into the `caches/jars-9` and `caches/transforms-3` directories that are currently being excluded from the Gradle User Home save/restore.

We can fix the first issue, but until we are sure we can remove the cache excludes we cannot use configuration caching. I'll add a comment to this effect.

Note that there isn't yet a way to tell `gradle-build-action` to avoid save/restore of the configuration-cache state: if the configuration-cache is written in a build invocation, the action will save it and restore on the next run. So `--no-configuration-cache` is the best way to avoid this problem.

.github/workflows/presubmit.yml
bigdaz: Done

.github/workflows/presubmit.yml
bigdaz: Done. Unfortunately these directories are undocumented.

.github/workflows/presubmit.yml
bigdaz: Comment added to workflow file.

liutikas (approved):

.github/workflows/presubmit.yml
```diff
@@ -171,42 +179,33 @@ jobs:
...
+          # Builds on other branches will only read existing entries from the cache.
+          cache-read-only: ${{ github.ref != 'refs/heads/androidx-main' }}
+
+          # Don't reuse cache entries from any other Job.
+          gradle-home-cache-strict-match: true
+
+          # Limit the size of the cache entry.
+          # These directories contain instrumented/transformed dependency jars which can be reconstructed relatively quickly.
+          gradle-home-cache-excludes: | 
+            caches/jars-9
```
yigit, line 200: Do we need something like jars-* instead of 9? (Idk if * is supported but 9 seems very random and could break).

yigit:
Overall LGTM, just not sure about the numbers in directories

.github/workflows/presubmit.yml
bigdaz: Unfortunately this parameter doesn't (yet) support wildcards. I'll add an issue to `gradle-build-action` for this.

The `jars-9` is tied to a set of Gradle versions, and will only be incremented when the instrumentation algorithm changes in Gradle. If you update to a new Gradle version with a new `jars-` version, then the cache entry will start to contain these files. Things will keep working, but we might start to see cache entry eviction based on hitting the 10Gb limit.

In the end, I'm not certain that these exclusions are necessary, but in my testing they allowed a bit of headroom for growth in entry size without leading to premature eviction. I'd like to get this in and then start to monitor the real-world behaviour, tweaking as necessary.

bigdaz: The `gradle-build-action` makes it easy to monitor the size of entries being uploaded: eg https://github.com/bigdaz/androidx/runs/4974841719?check_suite_focus=true#step:17:221. So this will allow us to monitor how well the Gradle User Home caching is performing.
I'm sure further tweaks will be required.

liutikas (approved):

PR #321, opened 2022-02-01
Use defaults to checkout source in GitHub workflow

bigdaz:
By default, the `actions/checkout@v2` action will checkout the code matching
the SHA of the event that triggered the workflow. By overriding this default
the current workflow can run against a commit different than expected, or
even against different commits for different jobs in the same workflow execution.

This change simplifies the workflow so that `actions/checkout` uses the default
behaviour, which will result in the workflow running builds against the expected commit.

## Testing

Test: Tested via GH Actions workflow


dlam (approved):

525411:
Merge pull request

PR #322, opened 2022-02-01
Use sub action to build individual projects

yigit:
This CL refactors the workflow to use a sub action to build individual
projects. This helps cleanup the build and also have an easy way to
disable certain builds.

I've also added a new ci-config.json file to control which projects are
included / excluded along with a companion script to decide based on the
current branch.

For now, we have two setups, 1 for compose and 1 for the rest. Compose
branch only enables compose projects meanwhile the main branch enables
everything by default.

Bug: n/a
Test: CI


.github/ci-control/ci-config.json
```diff
@@ -0,0 +1,21 @@
+{
+    "__comment": [
+        "Compose and Main are two configurations we have in CI.",
+        "Each can have include and exclude filters where these filters run a",
+        "  'contains' match on the project name. e.g. if the project name is compose-runtime",
```
liutikas: Can we go with exact matches and validate it matches a real project? I am a bit worried this will lead to subtle errors.

.github/actions/build-single-project/action.yml
```diff
@@ -0,0 +1,106 @@
...
+    description: "Name of the project (used for artifacts)"
+    required: true
+  project-root:
+    description: "Root folder of the project (used for building)"
+    required: true
+  gradle-cache-password:
+    description: "password for gradle cache"
+  gradle-enterprise-access-key:
+    description: "access key for gradle enterprise"
+  gradle-flags:
```
liutikas, line 14: Why is this one not required? We seem to always set it.

.github/ci-control/should_run_project.sh
```diff
@@ -0,0 +1,79 @@
...
+    echo "Missing projectName argument"
+    usage
+fi
+
+if  [[ $BRANCH == "" ]]; then
+    echo "Missing branch argument"
+    usage
+fi
+
+if [[ $BRANCH == *compose* ]]; then
```
liutikas: Can we only match `compose-compiler-*`? I prefer we are as explicit as possible

.github/ci-control/should_run_project.sh
```diff
@@ -0,0 +1,79 @@
+# Decides whether we should build a project in the current branch based on the ci-config.json file.
```
liutikas: Not a blocker, but could we pick python or something else thats better than bash? :)

.github/workflows/presubmit.yml
```diff
@@ -117,6 +116,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
+        # if you would like to remove some projects temporarily, use ci-config.json instead.
```
liutikas: Put pull path .github/ci-control/ci-config.json so people can find it.

liutikas (approved):
Could we add a README.md explaining how our CI is structured? It might be nice to make it easy to figure out how it works.

.github/ci-control/ci-config.json
yigit: yea i flip-flopped here a lot on what to do. Another option is to allow matchers here (e.g. you can write `*compose*`).

.github/actions/build-single-project/action.yml
yigit: they may not be available in the caller (actually, the secret is only available in post-submit builds)

.github/ci-control/should_run_project.sh
yigit: i'm afraid to admit that i'm more familiar with bash than python :) And i don't know bash at all :) so maybe i should try to learn python and convert this :)

525411:
Merge full

.github/ci-control/ci-config.json
yigit: moving to exact string match for now.

.github/ci-control/should_run_project.sh
yigit: moved to "*compose-compiler*"

.github/ci-control/should_run_project.sh
yigit: moved to python. Should not review on full stomach.

.github/ci-control/should_run_project.py
```diff
@@ -0,0 +1,65 @@
...
+import sys
+import getopt
+import json
+import os
+
+
+def usage(exitCode):
+    print("""
+        Usage:
+        ./should_run_project.sh --project <project-name> --branch <current branch name>\n\
```
dlam:
```suggestion
        ./should_run_project.py --project <project-name> --branch <current branch name>\n\
```

liutikas (approved):

525411:
Idonate all up to u.ok thanks...

yigit:
added a README for the ci-config folder.

PR #323, opened 2022-02-02
Use the correct argument name for FTL action

dlam:
Unclear why this started breaking now, maybe `github.event.workflow_run.id` was returning empty before so we were able to default the correct value?

Either way, it seems the intended arg name is target-run-id from: https://github.com/androidx/androidx-ci-action/blob/main/action.yml

Test: GH workflow
Change-Id: I8718e1efa1647133f9856712ef5735e02ac53e90


525411:
[]()

liutikas (approved):

PR #328, opened 2022-02-09
Use marker object instead of null for unset derived state results

DSteve595:
## Proposed Changes

Derived states currently use a `null` result to indicate that the calculation has not yet been run. However, since states can have nullable types, this means an actual result of `null` is treated as if no result were present, effectively breaking caching for the derived state. This changes the result to use a specific marker object when unset, rather than `null`. This is the approach taken by Kotlin's `Lazy`.

## Testing

Test: Added a test that ensures a calculation returning `null` is only run  once, similar to the existing `theCalculationIsCached()` test. (Should I move it next to that one?) The test fails without this change.

## Issues Fixed

Fixes: https://issuetracker.google.com/issues/217643833


compose/runtime/runtime/src/commonMain/kotlin/androidx/compose/runtime/DerivedState.kt
```diff
@@ -64,8 +64,10 @@ private class DerivedSnapshotState<T>(
     private var first: ResultRecord<T> = ResultRecord()
 
     private class ResultRecord<T> : StateRecord() {
+        private object Unset
```
JakeWharton: You can save on APK size / class-loaded memory by changing this to just an arbitrary instance instead.
```suggestion
        companion object {
            private val unset = Any()
        }
```
(also needs additional refactors of the code below due to casing change)

I'm not sure why the `Lazy` implementation feels the need to use an `object` which incurs the penalty of a whole class in the binary, a class being loaded at runtime, and a static field to forever retain the object instance.

Now you might say "but the `companion object` does the very same thing!". R8 will, however, eliminate the companion and hoist the reference to a static on the bytecode class for `ResultRecord` whereas it cannot do this for `Unset` as you might be relying on the fact that `Unset::class != Any::class` somewhere.

It seems small, but this kind of stuff can really be death by a thousand cuts...

Otherwise 👍

DSteve595: Makes sense, thanks!
I kept the casing as `Unset` to match other Compose static properties.

tikurahul (approved):
Approving on behalf of @chuckjaz.

PR #331, opened 2022-02-22
Explicitly specify MaxMetaspaceSize for playground

liutikas:
## Proposed Changes

Due to https://github.com/gradle/gradle/issues/19750 Gradle doesn't set a default
value for MaxMetaspaceSize when org.gradle.jvmargs is set and the JVM 8 default is
unbounded. This sets it explicitly to the default value defined in https://github.com/gradle/gradle/blob/6ccf8a2cc74c8060992bd36189dec929143caed7/subprojects/launcher/src/main/java/org/gradle/launcher/daemon/configuration/DaemonParameters.java#L39

## Testing

Test: None

## Issues Fixed

Daemon disappearing in CI

### DIFF
--- a/playground-common/androidx-shared.properties
+++ b/playground-common/androidx-shared.properties
@@ -24,7 +24,7 @@
 # This separation is necessary to ensure gradle can read certain properties
 # at configuration time.
 
-org.gradle.jvmargs=-Xmx4g -XX:+HeapDumpOnOutOfMemoryError -XX:+UseParallelGC -Dkotlin.daemon.jvm.options=-XX:MaxMetaspaceSize=1g -Dlint.nullness.ignore-deprecated=true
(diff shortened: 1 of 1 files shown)


playground-common/androidx-shared.properties
```diff
@@ -24,7 +24,7 @@
 # This separation is necessary to ensure gradle can read certain properties
 # at configuration time.
 
-org.gradle.jvmargs=-Xmx4g -XX:+HeapDumpOnOutOfMemoryError -XX:+UseParallelGC -Dkotlin.daemon.jvm.options=-XX:MaxMetaspaceSize=1g -Dlint.nullness.ignore-deprecated=true
+org.gradle.jvmargs=-Xmx4g -XX:+HeapDumpOnOutOfMemoryError -XX:+UseParallelGC -XX:MaxMetaspaceSize=256m -Dkotlin.daemon.jvm.options=-XX:MaxMetaspaceSize=1g -Dlint.nullness.ignore-deprecated=true
```
dlam: I think we need this to be 512m as I had bumped this  in the past due to seeing oom: metaspace error

liutikas: Done

dlam (approved):

PR #336, opened 2022-03-14
Update PageFetcherSnapshotTests using deprecated test API

claraf3:
Behavior of coroutine dispatchers changed when switching to
kotlinx-coroutines-test 1.6 API, requiring manual adjustments.

Bug: 220884819

### DIFF
--- a/paging/paging-common/src/test/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
+++ b/paging/paging-common/src/test/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
@@ -29,6 +29,12 @@ import androidx.paging.PagingSource.LoadResult.Page
 import androidx.paging.RemoteMediatorMock.LoadEvent
 import androidx.paging.TestPagingSource.Companion.LOAD_ERROR
 import com.google.common.truth.Truth.assertThat
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertFalse
+import kotlin.test.assertNotNull
+import kotlin.test.assertTrue
+import kotlin.test.fail
 import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.CoroutineStart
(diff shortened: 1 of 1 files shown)


paging/paging-common/src/test/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
```diff
@@ -2713,9 +2704,8 @@ class PageFetcherSnapshotTest {
         }
     }
 
-    @Suppress("DEPRECATION") // b/220884819
     @Test
-    fun remoteMediator_endOfPaginationNotReachedLoadStateAppend() = runBlockingTest {
+    fun remoteMediator_endOfPaginationNotReachedLoadStateAppend() = testScope.runTest {
```
dlam: Can we revert the change for this test for now, since we're missing an event and it seems like it'll take some work to make this test a bit more consistent?

paging/paging-common/src/test/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
```diff
@@ -2505,9 +2501,8 @@ class PageFetcherSnapshotTest {
         state.job.cancel()
     }
 
-    @Suppress("DEPRECATION") // b/220884819
     @Test
-    fun remoteMediator_endOfPaginationNotReachedLoadStatePrepend() = runBlockingTest {
+    fun remoteMediator_endOfPaginationNotReachedLoadStatePrepend() = testScope.runTest {
```
dlam: Do we need to use testScope here? Let's try to get rid of it if we don't need it since it relies on UnconfinedTestDispatcher.

claraf3: This test is also missing an event like remoteMediator_endOfPaginationNotReachedLoadStateAppend(). I'll revert this one as well.

And it does seem to require UnconfinedTestDispatcher. If we use the StandardTestDispatcher, coroutines are not getting launched properly but the test setup doesn't allow more refined control of launches. I'll leave a comment on the bug regarding this.

paging/paging-common/src/test/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
claraf3: Reverted along with its prepend counterpart remoteMediator_endOfPaginationNotReachedLoadStatePrepend

paging/paging-common/src/test/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
claraf3: reverted along with its Append counterpart and noted in bug report the need for new test setups to guarantee launch order

dlam (approved):

PR #337, opened 2022-03-14
Add room-paging-guava module to Room

claraf3:
Create module for integrating Guava into Room Paging

Test: n/a
Bug: 203666906


room/room-paging-guava/src/main/androidx/room/androidx-room-room-paging-guava-documentation.md
```diff
@@ -0,0 +1,7 @@
+# Module root
```
dlam: Empty package docs? Maybe just remove this if we're gonna leave it blank and we can add it back later if we actually have something to write here.

claraf3: Got it -- removed. I wasn't sure what to do with it.

droid-wan-kenobi (approved):
Looks good to me - welcome back Clara!

dlam (approved):

PR #340, opened 2022-03-15
Yigit/fix playground include regex

yigit:
A change in AOSP broke the project regex parser for settings.gradle.

This CL updates the regex and also adds tests for expected cases.

Bug: n/a
Test: local + CI


liutikas (approved):

playground-common/playground-plugin/src/main/kotlin/androidx/playground/SettingsParser.kt
```diff
@@ -0,0 +1,81 @@
...
+ * projects.
+ */
+internal object SettingsParser {
+    /**
+     * Match lines that start with includeProject, followed by a require argument for project gradle
+     * path and an optional argument for project file path.
+     */
+    /* ktlint-disable max-line-length */
+    private val includeProjectPattern = Regex(
+        """^[\n\r\s]*includeProject\("(?<name>[a-z0-9-:]*)"(,[\n\r\s]*"(?<path>[a-z0-9-/]+))?.*\).*$""",
```
dlam, line 32: I think \s includes \n and \r already. Also instead of IGNORE_CASE since you want `includeProject` to match exactly, you could use [a-zA-Z0-9] or if you don't like typing both a-z and A-Z you can use the POSIX character class `\p{alnum}`.

PR #341, opened 2022-03-19
Use type nullability instead of declaration nullability in EmbeddedField

mzgreen:
## Proposed Changes

Previously `EmbeddedField` was set to `null` if all of its columns were `null` unless it was annotated with `androidx.annotation.NonNull`. The issue is that for non nullable properties in data classes, the Kotlin compiler automatically adds `org.jetbrains.annotations.NotNull` annotation which isn't recognized by Room. This change fixes the issue by relying on type nullability instead of declaration nullability. Type nullability in Room compiler is inferred from type declaration for Kotlin types and from annotations for Java types.

## Testing

## Issues Fixed

Fixes: b/110463870

### DIFF
--- a/room/integration-tests/kotlintestapp/src/androidTest/java/androidx/room/integration/kotlintestapp/test/UsersDaoTest.kt
+++ b/room/integration-tests/kotlintestapp/src/androidTest/java/androidx/room/integration/kotlintestapp/test/UsersDaoTest.kt
@@ -20,8 +20,7 @@ import androidx.room.androidx.room.integration.kotlintestapp.vo.Email
 import androidx.room.androidx.room.integration.kotlintestapp.vo.User
 import androidx.room.integration.kotlintestapp.test.TestDatabaseTest
 import androidx.test.filters.MediumTest
-import org.hamcrest.CoreMatchers
-import org.hamcrest.MatcherAssert
+import com.google.common.truth.Truth.assertThat
 import org.junit.Test
 
 @MediumTest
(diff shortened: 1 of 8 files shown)


mzgreen:
Not sure why `buildOnServer zipTestConfigsWithApks` failed:
```
* What went wrong:
Gradle build daemon disappeared unexpectedly (it may have been killed or may have crashed)
```
I ran `buildOnServer` locally before pushing and it passed. I don't see an option to retry the build either.

dlam:
> I ran buildOnServer locally before pushing and it passed. I don't see an option to retry the build either.

Don't worry about this, is a known flake when running on github runners which seems to keep coming back to haunt us :/

room/integration-tests/kotlintestapp/src/androidTest/java/androidx/room/integration/kotlintestapp/test/UsersDaoTest.kt
```diff
@@ -0,0 +1,52 @@
...
+        usersDao.insertUser(USER_4)
+
+        val expectedList = ArrayList<User>()
+        expectedList.add(USER_1)
+        expectedList.add(USER_2)
+        expectedList.add(User(USER_3.userId, USER_3.email, null))
+        expectedList.add(User(USER_4.userId, USER_4.email, null))
+
+        MatcherAssert.assertThat(
+            database.usersDao().getUsers(), CoreMatchers.`is`<List<User>>(expectedList)
```
danysantiago: Can you please use the `Truth` library here, we are trying to migrate from assert test library. Should be something like: `Truth.assertThat(actual).isEqualTo(expected)`

danysantiago (approved):
Thanks, looks like we missed that nullability check during the migration to XProcessing. I took a look to see if we had more like this and it [seems this is the only one](https://cs.android.com/search?q=androidx.annotation&sq=&ss=androidx%2Fplatform%2Fframeworks%2Fsupport:room%2Froom-compiler%2Fsrc%2Fmain%2F), so thats good.

room/integration-tests/kotlintestapp/src/androidTest/java/androidx/room/integration/kotlintestapp/test/UsersDaoTest.kt
mzgreen: Done

PR #343, opened 2022-03-22
Introduce "NONE" conflict resolution strategy

mzgreen:
## Proposed Changes

Currently Room uses "ABORT" resolution strategy as a default. It means that for `@Insert` and `@Update` annotated methods, Room compiler generates SQL code containing on conflict statements. This prevents INSERT/UPDATE triggers from specifying their own resolution strategy.

This change adds "NONE" conflict resolution strategy. If "NONE" strategy is set, then the Room compiler won't generate on conflict statements for @Insert and @Update annotated methods.

## Testing

## Issues Fixed

Fixes: b/154196564

### DIFF
--- a/room/room-common/src/main/java/androidx/room/OnConflictStrategy.kt
+++ b/room/room-common/src/main/java/androidx/room/OnConflictStrategy.kt
@@ -36,7 +36,6 @@ public annotation class OnConflictStrategy {
          * OnConflict strategy constant used by default when no other strategy is set. The
          * runtime behavior is the same as when [ABORT] strategy is applied. *The transaction
          * is rolled back.*
-         *
          */
         public const val NONE: Int = 0
         /**
--- a/room/room-common/src/main/java/androidx/room/OnConflictStrategy.kt
+++ b/room/room-common/src/main/java/androidx/room/OnConflictStrategy.kt
@@ -33,9 +33,10 @@ import androidx.annotation.IntDef
 public annotation class OnConflictStrategy {
     public companion object {
         /**
-         * OnConflict strategy constant used by default when no other strategy is set. The
-         * runtime behavior is the same as when [ABORT] strategy is applied. *The transaction
-         * is rolled back.*
+         * OnConflict strategy constant used by default when no other strategy is set. Using it
+         * prevents Room from generating ON CONFLICT clause. It may be useful when there is a need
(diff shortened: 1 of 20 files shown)


mzgreen:
I wasn't able to find a way to do this without API changes. The change of the default `onConflict` value inside `@Insert` and `@Update` annotations seems to be a breaking change itself so I'm not sure if it's even possible to do it in a non breaking way.

yigit:
I think adding NONE is fine but we cannot change the default value since it will be a behavior change for existing code.

mzgreen:
> I think adding NONE is fine but we cannot change the default value since it will be a behavior change for existing code.

According to https://database.guide/how-on-conflict-works-in-sqlite/ SQLite is using ABORT by default if no on conflict clause is specified. So these two `INSERT INTO Products` and `INSERT OR ABORT INTO Products` should be equivalent, right?

danysantiago:
I believe the behaviour change is not that `ABORT` is the default conflict when no `ON CONFLICT` is specified, but that by removing it from `INSERT`, `UPDATE` and `DELETE` statements, those triggers with `ON CONFLICT` will behave differently, i.e. users might sadly be relying on the current behaviour.

Besides that the introduction of `NONE` and the PR looks good, and indeed it would enable those who care to use `NONE` and have their resolution in triggers.

mzgreen:
I see, it makes sense now. I'll revert the default values change 👍

mzgreen:
I've reverted the default values change.

room/room-common/src/main/java/androidx/room/OnConflictStrategy.kt
```diff
@@ -31,6 +32,13 @@ import androidx.annotation.IntDef
 )
 public annotation class OnConflictStrategy {
     public companion object {
+        /**
+         * OnConflict strategy constant used by default when no other strategy is set. The
+         * runtime behavior is the same as when [ABORT] strategy is applied. *The transaction
+         * is rolled back.*
+         *
```
danysantiago: nit: extra empty line here

danysantiago (approved):
Looks good, thank you!

room/room-common/src/main/java/androidx/room/OnConflictStrategy.kt
mzgreen: Fixed :)

PR #352, opened 2022-04-04
Add -Pandroidx.summarizeStderr to github ktlint

dlam:
Fixes: 203458619
Test: GH workflow log output
Change-Id: I63a7a6b0d8500db04bec2b7d73e0c27190486761


dlam:
So unfotunately this requires build_log_simplifier which depends on OUT_DIR being set. For playground we set CHEKCOUT_ROOT from gradle plugin extension :/

PR #361, opened 2022-04-06
Add workflow to create update prebuilts PR

yigit:
This CL adds a new workflow that can be manually triggered to create a prebuilt update PR.

Bug: n/a
Test: CI


.github/workflows/update_prebuilts.yml
```diff
@@ -0,0 +1,56 @@
...
+        required: false
+        default: ''
+        type: string
+jobs:
+  update-prebuilts:
+    runs-on: ubuntu-latest
+    steps:
+      - name: "Checkout androidx repo"
+        uses: actions/checkout@v2
+      - name: "Set environment variables"
```
dlam: Is this step really setting environment variables :thinking:

.github/workflows/update_prebuilts.yml
```diff
@@ -0,0 +1,56 @@
...
+        shell: bash
```
dlam, line 25: I know we do this elsewhere, but now we can use https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrun so we don't have to keep copying this around.

.github/workflows/update_prebuilts.yml
```diff
@@ -0,0 +1,56 @@
...
+        run: development/update_playground.sh
+      - name: "Check if anything changed"
+        shell: bash
+        id: "check-for-changes"
+        run: |
+          set -x
+          git status
```
dlam, line 32: is this just for logging? with set -x i don't think we'll need this to see the output of git status since we also call it below

.github/workflows/update_prebuilts.yml
```diff
@@ -0,0 +1,56 @@
...
+        id: "create-pr"
+        if: ${{ steps.check-for-changes.outputs.has_changes >= 1 }}
+        run: |
+          set -x
+          BRANCH_NAME="ci-workflow-bot/update-prebuilts-$(date "+%Y-%m-%d-%H-%M")/${{ github.run_id }}-${{ github.run_number }}"
+          git config user.email "androidx-github-bot@google.com"
+          git config user.name "AndroidX Github Bot"
+          git checkout -b $BRANCH_NAME
+          git add playground-common/playground.properties
+          git commit -m "Update Prebuilts" # PR body overrides this commit message
```
dlam:
```suggestion
          git commit -m "Update playground snapshot ids" # PR body overrides this commit message
```

Let's also change this to be accurate for git log history

.github/workflows/update_prebuilts.yml
```diff
@@ -0,0 +1,56 @@
...
+        shell: bash
+        id: "create-pr"
+        if: ${{ steps.check-for-changes.outputs.has_changes >= 1 }}
+        run: |
+          set -x
+          BRANCH_NAME="ci-workflow-bot/update-prebuilts-$(date "+%Y-%m-%d-%H-%M")/${{ github.run_id }}-${{ github.run_number }}"
+          git config user.email "androidx-github-bot@google.com"
+          git config user.name "AndroidX Github Bot"
+          git checkout -b $BRANCH_NAME
+          git add playground-common/playground.properties
```
dlam, line 45: Can we Just use `git add .` so we don't have to update this in case `update_playground.sh` changes or those properties jmove for whatever reason?

.github/workflows/update_prebuilts.yml
yigit: 🗡️ copy pasta. fixed

.github/workflows/update_prebuilts.yml
yigit: yea just to see it but also below we pipe it, idk if set x will show every intermediate step's output.

.github/workflows/update_prebuilts.yml
yigit: we don't need it, PR body overrides the message

.github/workflows/update_prebuilts.yml
yigit: i wanted to keep this specific to the file for now. we can consider -am later

dlam (approved):

PR #365, opened 2022-04-15
Further unify repositories used by playground

yigit:
This CL removes the common external repositories from the playground
root plugin since they are added by repos.gradle.

Furthermore, it also removes the repository list in the shared
playground-build file and instead delegates to repos.gradle.

Bug: n/a
Test: cd room && ./gradlew projects

### DIFF
--- a/buildSrc/repos.gradle
+++ b/buildSrc/repos.gradle
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+// if you make changes no this file, make sure to update ReposProxy class in buildSrc
 def supportRootFolder = ext.supportRootFolder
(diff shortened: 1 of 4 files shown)


tikurahul (approved):

buildSrc/repos.gradle
```diff
@@ -14,6 +14,7 @@
  * limitations under the License.
  */
 
+// if you make changes no this file, make sure to update ReposProxy class in buildSrc
```
mathjeff: Should "no this file" be "in this file"?

yigit: whoops. fixed.

PR #366, opened 2022-04-15
Download cmake 3.22.1 onto build machines

yigit:
AndroidX uses this Cmake version which is only available in the canary
channel.

Also re-enabled datastore-mutliprocess on playground and updated
contributing.md doc to include the cmake download instruction.

Fixes: 229400584
Test: Github CI

### DIFF
--- a/.github/actions/build-single-project/action.yml
+++ b/.github/actions/build-single-project/action.yml
@@ -33,27 +33,6 @@ runs:
       uses: actions/setup-java@v1
       with:
         java-version: "11"
-    # Required for CMake when building workmanager.
-    - name: "Install Ninja"
-      shell: bash
-      run: |
-        set -x
-        if [ "$RUNNER_OS" == "Linux" ]; then
-          # Should use version 1.10.2 but it is not available on apt yet.
-          sudo apt-get install ninja-build=1.10.0-1build1
-        elif [ "$RUNNER_OS" == "macOS" ]; then
-          # Should use version 1.10.2, but it is not well supported by Homebrew.
-          brew install ninja
-        elif [ "$RUNNER_OS" == "Windows" ]; then
-          choco install ninja --version=1.10.2
-        else
-          echo "Failed to install ninja due to unsupport OS: $RUNNER_OS"
-          exit 1
-        fi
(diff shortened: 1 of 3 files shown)


.github/actions/build-single-project/action.yml
```diff
@@ -54,7 +54,9 @@ runs:
         if [ ! -e "/usr/local/bin/ninja" ]; then
           ln -s "/usr/bin/ninja" "/usr/local/bin/ninja"
         fi
-
+    - name: "Install Cmake"
+      shell: bash
+      run: echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "cmake;3.22.1" --channel=3
```
dlam, line 38: `yes | sdkmanager --licenses` instead of `echo "yes"` if this is for licenses.

yigit: so when i did that, it failed with "broken pipe". idk why i thought it sends too many "yes"s so i changed it to echo.

.github/actions/build-single-project/action.yml
```diff
@@ -54,7 +54,9 @@ runs:
         if [ ! -e "/usr/local/bin/ninja" ]; then
           ln -s "/usr/bin/ninja" "/usr/local/bin/ninja"
         fi
-
+    - name: "Install Cmake"
```
dlam, line 36: Are we able to remove the ninja step above as well? It should be included in cmake.

dlam:
```
$ ~/Android/Sdk/cmake/3.22.1/bin/ninja --version
1.10.2
```

yigit: let me try, i was going to ask you about it.

yigit: hmm it passed but idk if it really tried to build or jus used the cache. i guess i need to change on of the WM elevant files and tr in another branch.

yigit: i couldn't figure out what in workmanager requires ninja. mightve been inspection which is not included anymore.

dlam (approved):

PR #367, opened 2022-04-15
Centralize playground-plugin repository management

liutikas:
It is all in one spot under playground-common/playground-plugin/settings.gradle

Test: ./gradlew tasks in room/


yigit:
is this in addition to mine or replaces mine? if it replaces, shouldn't we be removing the other lists? (the playground androidx root plugin and the common build file)

yigit (approved):

PR #372, opened 2022-04-22
Explicitly specify repositories for plugin resolution

liutikas:
Explicitly specify repositories for plugin resolution as the default is
gradlePluginPortal which has unsigned artifacts for kotlinx.serialization

Test: ./gradlew tasks locally


playground-common/playground-plugin/src/main/kotlin/androidx/playground/PlaygroundExtension.kt
```diff
@@ -87,6 +87,10 @@ open class PlaygroundExtension @Inject constructor(
      * @param relativePathToRoot The relative path of the project to the root AndroidX project
      */
     fun setupPlayground(relativePathToRoot: String) {
+        settings.pluginManagement.repositories {
```
dlam, line 92: Can we add a comment on why this is needed and how has helped with dependency verification errors?

yigit (approved):

PR #374, opened 2022-04-22
re-enable lifecyle

yigit:
Bug: n/a
Test: CI


dlam:
Continued here:  https://github.com/androidx/androidx/pull/430

PR #377, opened 2022-04-22
Remove explicit loading of buildSrc dependencies for the classpath

liutikas:
This seems to be an old workaround that is not needed anymore.

Test: ./gradlew tasks in room/


yigit (approved):
nicee!

PR #378, opened 2022-04-25
Paulseawa/datastore kmp

paulseawa:
## Proposed Changes

Moved SingleProcessDatastoreTest to common.   
- A few tests commented out to be resolved.  
- All tests passing in jvm
- 4 tests failing on macosArm64

Expanded kruth library.
Expect/actual corruption handlers.
Various warning fixed.

### DIFF
--- a/datastore/datastore-core/src/commonTest/kotlin/androidx/kruth/TruthForKotlinTest.kt
+++ b/datastore/datastore-core/src/commonTest/kotlin/androidx/kruth/TruthForKotlinTest.kt
@@ -39,18 +39,16 @@ open class KruthAssertion<T>(
 open class KruthStringAssertion(
     actual: String?
 ) : KruthAssertion<String>(actual) {
-    fun startsWith(prefix:String) {
-        assertTrue(actual?.startsWith(prefix) ?:false)
+    fun startsWith(prefix: String) {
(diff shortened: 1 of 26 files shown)


datastore/datastore-core/src/commonTest/kotlin/androidx/kruth/TruthForKotlinTest.kt
```diff
@@ -33,6 +34,59 @@ class KruthAssertion<T>(
...
 }
+
+open class KruthStringAssertion(
+    actual: String?
+) : KruthAssertion<String>(actual) {
+    fun startsWith(prefix:String):Boolean {
+        return actual?.startsWith(prefix) ?:false
+    }
+    fun endsWith(suffix:String):Boolean {
+        return actual?.endsWith(suffix) ?:false
```
yigit: doesn't these need to throw to fail the test?

yigit (approved):

datastore/datastore-core/src/commonTest/kotlin/androidx/kruth/TruthForKotlinTest.kt
paulseawa: Heh yes, nice catch.  Luckily the (jvm) tests are all still passing.

PR #379, opened 2022-04-25
Add native tests to bOS

yigit:
This PR updates AndroidXPlugin to add native tests to the build on server task.

Bug: n/a
Test: cd collection && ./gradlew bOS --dry-run

### DIFF
--- a/buildSrc/private/src/main/kotlin/androidx/build/AndroidXImplPlugin.kt
+++ b/buildSrc/private/src/main/kotlin/androidx/build/AndroidXImplPlugin.kt
@@ -84,6 +84,8 @@ import java.util.Locale
 import java.util.concurrent.ConcurrentHashMap
 import org.gradle.kotlin.dsl.KotlinClosure1
 import org.gradle.kotlin.dsl.register
+import org.jetbrains.kotlin.gradle.targets.native.KotlinNativeHostTestRun
+import org.jetbrains.kotlin.gradle.testing.KotlinTaskTestRun
 
 /**
  * A plugin which enables all of the Gradle customizations for AndroidX.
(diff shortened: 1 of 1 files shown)


buildSrc/private/src/main/kotlin/androidx/build/AndroidXImplPlugin.kt
```diff
@@ -653,6 +656,28 @@ class AndroidXImplPlugin : Plugin<Project> {
...
+            Project ${project.path} applies kotlin multiplatform plugin but we cannot find the
+            KotlinMultiplatformExtension.
+            """.trimIndent()
+        }
+        // Add all "configured" native tests to the buildOnServer task.
+        // Note that we don't check if platform is enabled in flags because it wouldn't be
+        // configured in the first place if the target is not enabled
+        kmpExtension.testableTargets.all { kotlinTarget ->
+            kotlinTarget.testRuns.all { kotlinTestRun ->
+                if (kotlinTestRun is KotlinNativeHostTestRun &&
```
dsaff, line 675: Can you add a quick comment about why both "is" checks are needed?  Is there not a subclass relationship between them?

dsaff (approved):

buildSrc/private/src/main/kotlin/androidx/build/AndroidXImplPlugin.kt
yigit: yea unfortunately i need to cast to `KotlinNativeHostTestRun` to limit it tests that run on host (e..g not ios) but also need to cast to `KotlinTaskTestRun` to get the `executionTask`. I'll take another look and if not, add comments.

yigit: added comment w/ the reason for each class.

PR #382, opened 2022-04-27
Update Compose compiler plugin for K/Native iOS

Zzzode:
- Updates function copy mechanisms used in iOS to match descriptors for KLIB compilation.
- Use decoys on K/N iOS target to fix deserialization error in link phase.

Test: `cd frameworks/support/ && ./gradlew :compose:integration-tests:demos:installDebug`

### DIFF
--- a/compose/compiler/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/ComposeIrGenerationExtension.kt
+++ b/compose/compiler/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/ComposeIrGenerationExtension.kt
@@ -35,6 +35,8 @@ import androidx.compose.compiler.plugins.kotlin.lower.decoys.SubstituteDecoyCall
 import org.jetbrains.kotlin.backend.common.extensions.IrGenerationExtension
 import org.jetbrains.kotlin.backend.common.extensions.IrPluginContext
 import org.jetbrains.kotlin.backend.common.serialization.DeclarationTable
+import org.jetbrains.kotlin.backend.common.serialization.GlobalDeclarationTable
+import org.jetbrains.kotlin.backend.common.serialization.mangle.ir.IrBasedKotlinManglerImpl
 import org.jetbrains.kotlin.backend.common.serialization.signature.IdSignatureSerializer
 import org.jetbrains.kotlin.backend.common.serialization.signature.PublicIdSignatureComputer
 import org.jetbrains.kotlin.ir.ObsoleteDescriptorBasedAPI
@@ -43,7 +45,9 @@ import org.jetbrains.kotlin.ir.backend.js.lower.serialization.ir.JsManglerIr
 import org.jetbrains.kotlin.ir.declarations.IrModuleFragment
 import org.jetbrains.kotlin.platform.js.isJs
 import org.jetbrains.kotlin.platform.jvm.isJvm
+import org.jetbrains.kotlin.platform.konan.isNative
 import org.jetbrains.kotlin.resolve.DelegatingBindingTrace
+import kotlin.reflect.full.primaryConstructor
 
 class ComposeIrGenerationExtension(
     @Suppress("unused") private val liveLiteralsEnabled: Boolean = false,
@@ -56,6 +60,7 @@ class ComposeIrGenerationExtension(
     private val reportsDestination: String? = null
 ) : IrGenerationExtension {
     var metrics: ModuleMetrics = EmptyModuleMetrics
+
     @OptIn(ObsoleteDescriptorBasedAPI::class)
     override fun generate(
         moduleFragment: IrModuleFragment,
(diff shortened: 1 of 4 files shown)


Zzzode:
Hello, I'm trying to get Compose Compiler plugin to work on iOS. I use Decoys on K/N iOS to fix deserialization error in K/N link phase.

jimgoog:
cc @igordmn

compose/compiler/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/lower/ComposerParamTransformer.kt
```diff
@@ -485,7 +495,9 @@ class ComposerParamTransformer(
         if (this !is IrSimpleFunction) return false
         if (valueParameters[index].defaultValue != null) return true
 
-        if (context.platform.isJs() && this.isDecoyImplementation()) {
+        // support decoy on native target
+        if (context.platform.isJs() || context.platform.isNative() &&
```
eymar, line 499: shouldn't this condition look like this?
```kotlin 
if ((context.platform.isJs() || context.platform.isNative()) && this.isDecoyImplementation()) { ... }
```

Otherwise it will be always true for k/js targets.

compose/compiler/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/lower/decoys/CreateDecoysTransformer.kt
```diff
@@ -178,14 +347,21 @@ class CreateDecoysTransformer(
...
         val newFunction = factory {
             updateFrom(original)
             name = newName
             returnType = original.returnType
-            isPrimary = (original as? IrConstructor)?.isPrimary ?: false
+            // In konan backend for K/N, only a primary constructor can has
+            // IrInstanceInitializerCall inside, others MUST have IrDelegatingConstructorCall.
+            // So if we building a decoy implementation constructor, we should check and transfer
+            // "isPrimary" flag if need.
+            isPrimary =
```
eymar, line 362: I try to understand this change:
previous version: `isPrimary = (original as? IrConstructor)?.isPrimary ?: false`
new version `if (original is IrConstructor && context.platform.isNative()) original.isPrimary else false`

If I understand correctly, the previous version did the same but for all targets (k/js, k/n), right? And the new version will set `true` only for k/n. So can we keep the old version for setting `isPrimary` here?

compose/compiler/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/ComposeIrGenerationExtension.kt
```diff
@@ -121,6 +126,18 @@ class ComposeIrGenerationExtension(
 
         val mangler = when {
             pluginContext.platform.isJs() -> JsManglerIr
+            // Use Reflection to get KonanManglerIr from konan
+            // TODO: Use `import` to get org.jetbrains.kotlin.backend.konan.* modules
+            pluginContext.platform.isNative() -> {
```
eymar, line 131: I guess we need to access KonanManglerIr and KonanGlobalDeclarationTable for k/n only with decoys enabled. JB's builds of Compose for K/N don't use decoys now. So maybe adding one more condition `decoysEnabled` would be helpful. 

Another thought: ideally, it would be accessible without reflection.  Maybe it's worth submitting a ticket at https://youtrack.jetbrains.com/ describing the use case and linking it here in a comment, so we don't forget to update/improve this part later.

compose/compiler/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/lower/decoys/CreateDecoysTransformer.kt
```diff
@@ -162,10 +229,112 @@ class CreateDecoysTransformer(
...
         }
+
+        // We want to make origin constructor non-primary, cause ONLY primary constructor can have
+        // instance initializer call. We create a constructor's copy to transfer to non-primary,
+        // but after that, we should fix references to the origin one to make IR tree correct,
+        // modify by reflecting is the more convenient way :).
+        // If you want to copy from origin, should add @Decoy annotation on it also. References to
+        // the origin constructor haven't been changed, pass "SubstituteDecoyCallsTransformer"
+        // will check @Decoy annotation and change reference to decoy implementation.
+        if (context.platform.isNative() && original.isPrimary) {
```
eymar, line 244: At what moment does it get broken if we keep `isPrimary` as is (not changing it here at all)?

compose/compiler/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/ComposeIrGenerationExtension.kt
eymar: Maybe, since, both `mangler` and `idSignatureBuilder` needed only when decoys enabled, we can have two functions (somewhere in decoys sub-package) to create mangler and idSignatureBuilder instances. This would make this class a bit lighter, I guess.

compose/compiler/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/lower/decoys/CreateDecoysTransformer.kt
```diff
@@ -150,6 +159,64 @@ class CreateDecoysTransformer(
         }
     }
 
+    override fun visitCall(expression: IrCall): IrExpression {
+        val function = (expression.symbol.owner as? IrSimpleFunction)
+            ?: return super.visitCall(expression)
+        val property = (function.correspondingPropertySymbol?.owner)
+            ?: return super.visitCall(expression)
+
+        if (function.isDecoy() || !function.shouldBeRemapped()) {
```
eymar, line 168: Should we also skip this part for k/js (here and in other cases below)? 

`if (function.isDecoy() || !function.shouldBeRemapped() || context.platform.isJs()) {` ?

jimgoog:
Ping @Zzzode

liutikas:
This is obsolete as this project has moved to Jetbrains/kotlin repository.

PR #383, opened 2022-04-28
Fix collection KMP by duplicating Lock

yigit:
This is just a temporary fix

Bug: n/a
Test: gw bOS


dlam:
How did core fail here?

dlam (approved):

dlam:
oh it only failed on push event, it seems unrelated anyhow, but I am confused on why there are no logs in the build step.

dlam:
Also, should we file a bug to de-duplicate?

yigit:
> Also, should we file a bug to de-duplicate?

Yesi should do that. Was quickly sending this to try before leaving home to commute.

collection/collection/src/macosArm64Main/kotlin/androidx/collection/internal/Lock.macosArm64.kt
```diff
@@ -0,0 +1,69 @@
...
+    }
+
+    private class Resources {
+        private val arena = Arena()
+        private val attr: pthread_mutexattr_t = arena.alloc()
+        val mutex: pthread_mutex_t = arena.alloc()
+
+        init {
+            pthread_mutexattr_init(attr.ptr)
+            pthread_mutexattr_settype(attr.ptr, PTHREAD_MUTEX_RECURSIVE.convert())
```
arkivanov, line 59: Most likely `convert` is not need here, if duplicated per each source set.

PR #384, opened 2022-04-28
Paulseawa/datastore preferences core kmp

paulseawa:
datastore-preferences-core converted using Json serialization for non-jvm.

All JVM tests passing.  A few native test failing mostly because of collection mutability.
Also added datastore-internal-test-utils to unify testing utilities between modules.
TODO: wire implementation


datastore/datastore-core-okio/build.gradle
```diff
@@ -29,7 +29,7 @@ kotlin {
         withJava()
     }
     macosX64()
-    linuxArm64()
+//    linuxArm64()
```
yigit, line 32: is this related to the Lock breakage or something else preventing it ?

datastore/datastore-core/src/nativeMain/kotlin/androidx/datastore/core/OkioStorage.kt
```diff
@@ -23,7 +23,7 @@ import okio.FileNotFoundException
 import okio.FileSystem
 import okio.Path
 
-internal class OkioStorage<T>(
+class OkioStorage<T>(
```
yigit, line 26: some of these internal/actual changes; (OkioStorage, FileStorage, StorageImpl); are they "just to keep progress moving" or are they real required changes? I'm just curious to see if it will have an impact on how we decide between option 1/3

datastore/datastore-internal-test-utils/build.gradle
```diff
@@ -0,0 +1,102 @@
...
+
+        }
+        nativeMain {
+            dependsOn(commonMain)
+        }
+        nativeTest {
+            dependsOn(commonTest)
+        }
+    }
+    targets.forEach {target ->
```
yigit, line 71: i think we should do this in the main file too

datastore/datastore-preferences-core/build.gradle
```diff
@@ -16,29 +16,98 @@
...
+            /* from = */ "com.google.protobuf",
+            /* to =   */ "androidx.datastore.preferences.protobuf"
+    )
+    sourceSets {
+        commonMain {
+            dependencies {
+                api(libs.kotlinStdlib)
+                api(libs.kotlinCoroutinesCore)
+                api(project(":datastore:datastore-core"))
+                api(project(":datastore:datastore-core-okio"))
```
yigit, line 49: hmm looks like this is full okio based, is that a relevant thing wrt option 1/3 or just it is easier to do for now ?

datastore/datastore-preferences-core/src/commonMain/kotlin/androidx/datastore/preferences/core/PreferenceDataStoreFactory.kt
```diff
@@ -0,0 +1,76 @@
...
+    DataStore<Preferences> by delegate {
+    override suspend fun updateData(transform: suspend (t: Preferences) -> Preferences):
+        Preferences {
+        return delegate.updateData {
+            val transformed = transform(it)
+            // Freeze the preferences since any future mutations will break DataStore. If a user
+            // tunnels the value out of DataStore and mutates it, this could be problematic.
+            // This is a safe cast, since MutablePreferences is the only implementation of
+            // Preferences.
+            (transformed as MutablePreferences).freeze()
```
yigit, line 72: why do we freeze? it should be no-op in the new memory model

datastore/datastore-preferences-core/src/commonMain/kotlin/androidx/datastore/preferences/core/PreferenceDataStoreFactory.kt
```diff
@@ -0,0 +1,76 @@
...
+     * @param produceFile Function which returns the file that the new DataStore will act on.
+     * The function must return the same path every time. No two instances of PreferenceDataStore
+     * should act on the same file at the same time. The file must have the extension
+     * preferences_pb.
+     *
+     * @return a new DataStore instance with the provided configuration
+     */
+    @JvmOverloads
+    public fun create(
+        storage: Storage<Preferences>,
```
yigit, line 55: i'm a bit confused on how this API works, more specifically, how do they create the Storage?

yigit (commented):
i'm a bit confused w/ the api part so maybe i need to re-read it or check it out locally.
i though the preferences json/protobuf will all be hidden from public API such that we do have a PreferencesSerializer that is backed by one of them and developer doesn't know about it.
but even in test we seem to have getSerializer so i'm a bit confused.

datastore/datastore-preferences-core/src/commonMain/kotlin/androidx/datastore/preferences/core/PreferenceDataStoreFactory.kt
paulseawa: This is a datastore preferences freeze.  To ensure people don't change preferences that aren't being modified within a write block.

datastore/datastore-preferences-core/src/commonMain/kotlin/androidx/datastore/preferences/core/PreferenceDataStoreFactory.kt
paulseawa: The actuals take files and paths. (well still have to do the native one).  This is a similar to the factory in datstore-core.

But the real answer is, I got it to work with the tests and I didn't want to over-think it until we decide on 1 vs. 3 which will really impact the factories.

datastore/datastore-core/src/nativeMain/kotlin/androidx/datastore/core/OkioStorage.kt
paulseawa: This is required to allow the new test-utils to create OkIoStorage which are in a different module

datastore/datastore-core-okio/build.gradle
paulseawa: Need to clean up.  I had to comment these out earlier, but it cleared up. Missed this one.

datastore/datastore-internal-test-utils/build.gradle
paulseawa: Don't quite follow.

PR #387, opened 2022-04-30
Move LongSparseArray to common

yigit:
This CL moves the LongSparseArray to common.

Unfortunately, LongSparseArray implements Cloneable (jvm only) and Kotlin doesn't support having actual implementations in expect classes. (See: https://youtrack.jetbrains.com/issue/KT-20427

We have two options:
* fake cloneable (this CL)
* commonize and duplicate code: https://android-review.googlesource.com/c/platform/frameworks/support/+/2079822/5

In this implementation, we create two interfaces:

```
internal expect interface CloneableKmp
public expect abstract class JvmCloneableAny {
   protected constructor()
   public open fun clone(): Any?
}
```

We need the interface to mark LongSparseArray as `Cloneable`.
We need the class so that `LongSparseArray` can call `super.clone`.

The interface gets typealiased to `java.lang.Cloneable` whereas the class gets typealised to `java.lang.Object`.

This way, we have 0 API impact for jvm/android.

For native, there are real implementations of these where the native `JvmCloneableAny#clone` method returns `null`.
To handle both cases, `LongSparseArray#clone` handles the case where `super.clone` returns `null`.

Note that, we don't have true jvm like cloning on the non-jvm platforms so it will break if it is sub-classed.
I think it is a fair compromise to avoid duplicating the class. 

Bug: b/228347315
Test: LongSparseArrayTest


Osaigbovo (approved):

dlam:
Merged this version instead: https://android-review.googlesource.com/c/platform/frameworks/support/+/2079822/6

PR #388, opened 2022-05-02
Cleanup Playground Setup

yigit:
This PR updates the playground setup to re-sym-link into `playground-common/gradle` where we can have symlinks to the main project for select files.

In the new setup:
`playground-common/gradle` has files that are symlinks to `<root>/gradle`'s files that we want (libs, verification metadata etc)
Each playground project's `gradle` folder symlinks to `playground-common/gradle`.
.idea, gradlew and gradlew.bat files are still symlinked to playground-common/* (same as before).
This should ensure that, in the future if `<root>/gradle` contents change, we can easily re-link all playground projects by linking the new file from `playground-common`.

I updated the playground plugin to no longer load the libs.version.toml file since it is now symlinked to each project (this is what will fix the tests).

I've also added a script to update all existing playground setups so that if we make future structural changes, we can run `playground-common/update-playground-setups.sh`

Bug: n/a
Test: CI


yigit:
ready for review.
files that should be reviewed: 
* changes in playground-common
* changes in PlaygroundPlugin.
rest is just auto-generated from re-running the scripts.

playground-common/setup-playground.sh
```diff
@@ -40,11 +47,7 @@ for IDEA_CONFIG_FILE in "${TRACKED_IDEA_FILES[@]}"
 do
     # path to the actual .idea config file
     ORIGINAL_FILE="$PLAYGROUND_REL_PATH/../$IDEA_CONFIG_FILE"
-    TARGET_DIR=$(dirname $IDEA_CONFIG_FILE)
-    # relativize it wrt to the file we'll create
-    REL_PATH=$(relativize $ORIGINAL_FILE $TARGET_DIR )
-    # symlink to the original idea file
-    ln -s $REL_PATH $IDEA_CONFIG_FILE
-    # forse add the file to git
+    # force add the file to git
```
dlam: nit: comment is on the wrong line

playground-common/update-playground-setups.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# Helper script to update all playground setups when we make changes
+# on how playground works (in other words, when we update setup-playground.sh).
```
dlam, line 3: Can you add a comment for sample invocation?

playground-common/update-playground-setups.sh
```diff
@@ -0,0 +1,28 @@
...
+
+WORKING_DIR=$(pwd)
+PLAYGROUND_REL_PATH=$(dirname $0)
+SUPPORT_ROOT_ABS_PATH=$(absPath "$PLAYGROUND_REL_PATH/..")
+PLAYGROUND_ABS_PATH=$(absPath $PLAYGROUND_REL_PATH)
+
+# re-runs the playground setup script on the given folder
+function setupPlayground {
+    echo "setting up playground in $1"
+    cd $1
```
dlam: nit: `cd $1 && $PLAYGROUND_ABS_PATH/setup-playground.sh`

is slightly safer

playground-common/update-playground-setups.sh
```diff
@@ -0,0 +1,28 @@
...
+    $PLAYGROUND_ABS_PATH/setup-playground.sh
+    echo "finished setting up playground in $1"
+}
+
+# find all playground settings files
+PLAYGROUND_SETTINGS_FILES=$(egrep -lir --include=settings.gradle "setupPlayground" $SUPPORT_ROOT_ABS_PATH)
```
dlam: just `egrep -lr` since i think you don't want case insensitive?

playground-common/update-playground-setups.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+# Helper script to update all playground setups when we make changes
+# on how playground works (in other words, when we update setup-playground.sh).
+function absPath {
+    python3 -c "import os.path; print(os.path.abspath('$1'))"
```
dlam, line 6: don't need to change, but just fyi i think `realpath` is on glinux by default and does the same thing

playground-common/update-playground-setups.sh
dlam: Also this only works if you run from `frameworks/support` right? Can we add validation for `$WORKING_DIR` ?

playground-common/update-playground-setups.sh
```diff
@@ -0,0 +1,28 @@
...
+}
+
+# find all playground settings files
+PLAYGROUND_SETTINGS_FILES=$(egrep -lir --include=settings.gradle "setupPlayground" $SUPPORT_ROOT_ABS_PATH)
+for SETTINGS_FILE in $PLAYGROUND_SETTINGS_FILES
+do
+    PROJECT_DIR=$(absPath $(dirname $SETTINGS_FILE))
+    setupPlayground $PROJECT_DIR
+done
+cd $WORKING_DIR
```
dlam: If you do `cd && blah` above i think you won't need this anymore

playground-common/setup-playground.sh
```diff
@@ -4,23 +4,30 @@
...
-rm -rf buildSrc
-ln -s "${PLAYGROUND_REL_PATH}/../buildSrc" buildSrc
+# helper symlink function that will also normalize the paths.
+function symlink() {
+    SRC=$1
+    TARGET=$2
+    TARGET_PARENT_DIR=$(dirname $TARGET)
+    REL_PATH_TO_TARGET_PARENT=$(relativize $SRC $WORKING_DIR/$TARGET_PARENT_DIR)
+    rm -rf $TARGET
+    ln -s $REL_PATH_TO_TARGET_PARENT $TARGET
```
dlam, line 20: Just curious, for relative path, does `ln -sr` not work? I think it will automatically relativize the path based on original file.

dlam (approved):

playground-common/update-playground-setups.sh
yigit: resolved offline.

playground-common/update-playground-setups.sh
yigit: it should work anywhere.

playground-common/update-playground-setups.sh
yigit: yea it doesn't exist on macs.

playground-common/update-playground-setups.sh
yigit: i think so but i might need to do `()` for that to work. wasn't sure so kept it like this

playground-common/setup-playground.sh
yigit: `ln: illegal option --r` so i guess it is not a thing on mac? will keep it as is.

playground-common/update-playground-setups.sh
yigit: fixed

PR #394, opened 2022-05-11
Paulseawa/datastore io split

paulseawa:
Early draft of splitting datastore-io into two jvm targets.  

Code is compiling but still need to move tests over.


datastore/datastore-core/build.gradle
```diff
@@ -15,22 +15,89 @@
...
-    api("androidx.annotation:annotation:1.1.0")
+def datastoreType = Attribute.of('androidx.datastore.type', String)
 
-    testImplementation(libs.junit)
-    testImplementation(libs.kotlinCoroutinesTest)
-    testImplementation(libs.truth)
-    testImplementation(project(":internal-testutils-truth"))
+kotlin {
+    jvm {
+        attributes.attribute(datastoreType, 'okio')
```
yigit, line 31: discussed offline but putting it here for reference so we can followup.
my understanding was that, if we declare an attirbute for the "Default", this will require anyone downstream to declare that (and make gradle fail with a confusing error because it cannot find matching artifact).

datastore/datastore-io/src/googMain/kotlin/androidx/datastore/io/IO.kt
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package androidx.datastore.io
+
+import java.io.File
+import java.io.InputStream
+import java.io.OutputStream
+
+actual class DatastoreOutput(val outputStream:OutputStream) {
```
yigit, line 23: i thought we were going to typealias these to InputStream / BufferedSink.

Would be good to see how you'll write 1 Serializer in our codebase that compiles with both of these, it feels impossible to write that in common right now unless i'm missing something. The DatastoreOutput expect has no methods so how would you use it from common? (feels like option 3)

PR #407, opened 2022-05-28
Fix ViewportHints sent to wrong generation

claraf3:
Currently, the new generation's receiver is set in PagingDataDiffer
before new generation is actually presented. This leads to current
generation's hints (which are calcualted based on current gen's
placeholder states) being sent to new generation.

Refactor the differ to set new receiver after new generation is
presented.

Fixes: 234030464
Change-Id: Iced47b9366805cb75d5f7a666cb8ae07f038c7f1

### DIFF
--- a/paging/paging-common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt
+++ b/paging/paging-common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt
@@ -43,7 +43,8 @@ public abstract class PagingDataDiffer<T : Any>(
     private val mainContext: CoroutineContext = Dispatchers.Main
 ) {
     private var presenter: PagePresenter<T> = PagePresenter.initial()
-    private var receiver: UiReceiver? = null
+    private var hintReceiver: HintReceiver? = null
+    private var uiReceiver: UiReceiver? = null
     private val combinedLoadStatesCollection = MutableCombinedLoadStateCollection()
     private val onPagesUpdatedListeners = CopyOnWriteArrayList<() -> Unit>()
 
@@ -139,10 +140,24 @@ public abstract class PagingDataDiffer<T : Any>(
 
     public suspend fun collectFrom(pagingData: PagingData<T>) {
         collectFromRunner.runInIsolation {
-            receiver = pagingData.receiver
-
+            uiReceiver = pagingData.uiReceiver
             pagingData.flow.collect { event ->
                 withContext(mainContext) {
+                    /**
+                     * The hint receiver of a new generation is set only after it has been
+                     * presented. This ensures that:
+                     *
+                     * 1. while new generation is still loading, access hints (and jump hints) will
+                     * be sent to current generation.
+                     *
+                     * 2. the access hint sent from presentNewList will have the correct
+                     * placeholders and indexInPage adjusted according to new presenter's most
+                     * recent state
+                     *
+                     * Ensuring that viewport hints are sent to the correct generation helps
+                     * synchronize fetcher/presenter in the correct calculation of the
+                     * next anchorPosition.
+                     */
                     if (event is Insert && event.loadType == REFRESH) {
                         presentNewList(
                             pages = event.pages,
@@ -151,6 +166,7 @@ public abstract class PagingDataDiffer<T : Any>(
                             dispatchLoadStates = true,
                             sourceLoadStates = event.sourceLoadStates,
                             mediatorLoadStates = event.mediatorLoadStates,
+                            newHintReceiver = pagingData.hintReceiver
                         )
                     } else if (event is StaticList) {
                         presentNewList(
@@ -166,6 +182,7 @@ public abstract class PagingDataDiffer<T : Any>(
                                 event.mediatorLoadStates != null,
                             sourceLoadStates = event.sourceLoadStates,
                             mediatorLoadStates = event.mediatorLoadStates,
+                            newHintReceiver = pagingData.hintReceiver
                         )
                     } else {
                         if (postEvents()) {
@@ -212,7 +229,7 @@ public abstract class PagingDataDiffer<T : Any>(
                                     presenter.storageCount
 
                                 if (shouldResendHint) {
-                                    receiver?.accessHint(
+                                    hintReceiver?.accessHint(
                                         presenter.accessHintForPresenterIndex(lastAccessedIndex)
                                     )
                                 } else {
@@ -247,7 +264,7 @@ public abstract class PagingDataDiffer<T : Any>(
         lastAccessedIndexUnfulfilled = true
         lastAccessedIndex = index
 
-        receiver?.accessHint(presenter.accessHintForPresenterIndex(index))
+        hintReceiver?.accessHint(presenter.accessHintForPresenterIndex(index))
         return presenter.get(index)
     }
 
@@ -280,7 +297,7 @@ public abstract class PagingDataDiffer<T : Any>(
      *  * [RemoteMediator.load] returning [RemoteMediator.MediatorResult.Error]
      */
     public fun retry() {
-        receiver?.retry()
+        uiReceiver?.retry()
     }
 
     /**
@@ -300,7 +317,7 @@ public abstract class PagingDataDiffer<T : Any>(
      * @sample androidx.paging.samples.refreshSample
      */
     public fun refresh() {
-        receiver?.refresh()
+        uiReceiver?.refresh()
     }
 
     /**
@@ -420,6 +437,7 @@ public abstract class PagingDataDiffer<T : Any>(
         dispatchLoadStates: Boolean,
         sourceLoadStates: LoadStates?,
         mediatorLoadStates: LoadStates?,
+        newHintReceiver: HintReceiver,
     ) {
         require(!dispatchLoadStates || sourceLoadStates != null) {
             "Cannot dispatch LoadStates in PagingDataDiffer without source LoadStates set."
@@ -440,6 +458,7 @@ public abstract class PagingDataDiffer<T : Any>(
             onListPresentable = {
                 presenter = newPresenter
                 onListPresentableCalled = true
+                hintReceiver = newHintReceiver
             }
         )
         check(onListPresentableCalled) {
@@ -461,7 +480,7 @@ public abstract class PagingDataDiffer<T : Any>(
             // Send an initialize hint in case the new list is empty, which would
             // prevent a ViewportHint.Access from ever getting sent since there are
             // no items to bind from initial load.
-            receiver?.accessHint(newPresenter.initializeHint())
+            hintReceiver?.accessHint(newPresenter.initializeHint())
         } else {
             // Transform the last loadAround index from the old list to the new list
             // by passing it through the DiffResult, and pass it forward as a
@@ -471,7 +490,7 @@ public abstract class PagingDataDiffer<T : Any>(
             // the prepend / append load that would have fulfilled it in the old
             // list.
             lastAccessedIndex = transformedLastAccessedIndex
-            receiver?.accessHint(
+            hintReceiver?.accessHint(
                 newPresenter.accessHintForPresenterIndex(
                     transformedLastAccessedIndex
                 )
--- a/paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
+++ b/paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
@@ -43,6 +43,7 @@ import kotlinx.coroutines.flow.consumeAsFlow
 import kotlinx.coroutines.flow.first
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.flow.flowOf
+import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
 import kotlinx.coroutines.test.TestScope
@@ -141,7 +142,6 @@ class PagingDataDifferTest(
         }
 
         differ.retry()
-
         assertEquals(1, receiver.retryEvents.size)
 
         job.cancel()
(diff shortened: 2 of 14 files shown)


paging/paging-common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt
```diff
@@ -139,7 +139,16 @@ public abstract class PagingDataDiffer<T : Any>(
...
+            /**
+             * Set receiver right away for first generation because we need a receiver to
+             * process refresh/retry calls.
+             *
+             * Subsequent generations will only register receiver after initial load is presented.
+             * If refresh/retry calls happened before the new gen is presented, the calls will
+             * be sent to current(old) receiver and fetcher would be able to trigger next generation
+             * based on cached hint
+             */
+            receiver = receiver ?: pagingData.receiver
```
dlam: This seems a bit brittle - I think we might be better served by splitting it into two separate interfaces, one for stateful single generation signals which needs to wait for real data and one for connecting to PageFetcher for multi-generational signals (that can be sent right away without waiting for real data).

For example, what happens if you mix multiple `Pager`s or display a static list first using `PagingData.from()` ? It seems like the correct thing to do here is to actually always set the multi-generational one immediately, but for the stateful hint signals, we can wait to set that until refresh succeeds. This also brings up the problem of mixing the static list with a regular pager causing us to drop refresh signals, but we can look into that separately.

paging/paging-common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt
```diff
@@ -440,6 +452,7 @@ public abstract class PagingDataDiffer<T : Any>(
             onListPresentable = {
                 presenter = newPresenter
                 onListPresentableCalled = true
+                receiver = newReceiver
```
dlam: Can we add a comment for how important it is to set this here and not earlier so we don't hit the same bug in future refactor?

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -163,6 +164,314 @@ class PagingDataDifferTest(
...
+        // double check that the pagingdata's receiver was registered and had received refresh call
+        // before any PageEvent is collected/presented
+        assertThat(receiver1.refreshEvents.size).isEqualTo(1)
+
+        job1.cancel()
+    }
+
+    @Test
+    fun receiverSetAfterNewListPresented() = testScope.runTest {
+        withContext(coroutineContext) {
```
dlam: Why is `withContext` here necessary?

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -163,6 +164,314 @@ class PagingDataDifferTest(
...
+            val differ = SimpleDiffer(differCallback = dummyDifferCallback)
+            val pageEventCh1 = Channel<PageEvent<Int>>(Channel.UNLIMITED)
```
dlam: Instead of using a Channel here and converting it to a Flow, can we just make a Flow directly? There is a `flow {  }` builder that should work

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -163,6 +164,314 @@ class PagingDataDifferTest(
...
+            pageEventCh2.trySend(
+                localLoadStateUpdate(refreshLocal = Loading)
+            )
+            assertThat(differ.loadStateFlow.first()).isEqualTo(
+                localLoadStatesOf(refreshLocal = Loading)
+            )
+
+            // try sending a hint, should be sent to first receiver
+            differ[4]
+            assertThat(receiver1.hints.size).isEqualTo(1)
```
dlam: nit: I think Truth has a `.hasSize(int)` assert that would give a clearer error

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -1611,7 +1924,9 @@ private fun infinitelySuspendingPagingData(receiver: UiReceiver = dummyReceiver)
     receiver
 )
 
-private class UiReceiverFake : UiReceiver {
+private class UiReceiverFake(
```
dlam: nit: If we're actually passing through to a real UiReceiver, I don't think this can be considered a fake anymore.

suggest: TrackableUiReceiverWrapper

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -1577,10 +1886,14 @@ class PagingDataDifferTest(
                     ).also { pagingSources.add(it) }
                 }
             ),
+        pagingDataMapping: (PagingData<Int>) -> PagingData<Int> = { it },
```
dlam: Since we never pass anything differently and it's a no-op anyway, let's just simplify and always do the wrapping for tracking here

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -1611,7 +1924,9 @@ private fun infinitelySuspendingPagingData(receiver: UiReceiver = dummyReceiver)
     receiver
 )
 
-private class UiReceiverFake : UiReceiver {
+private class UiReceiverFake(
+    private val operationalUiReceiver: UiReceiver? = null,
```
dlam: nit: just call this `uiReceiver`

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: Removed.

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: Done.

paging/paging-common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt
claraf3: Done.

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: Done.

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: done

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: Done.

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: Done.

paging/paging-common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt
claraf3: The differ now holds two receivers: a statelessReceiver for multi-gen, and a hintReceiver for access hints. Both would store the same pagingData.receiver but calls to refresh and retry would strictly use statelessReceiver which is set right away.

paging/paging-common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt
```diff
@@ -139,10 +140,24 @@ public abstract class PagingDataDiffer<T : Any>(
 
     public suspend fun collectFrom(pagingData: PagingData<T>) {
         collectFromRunner.runInIsolation {
-            receiver = pagingData.receiver
-
+            statelessReceiver = pagingData.receiver
```
dlam: Since we still need to set these independently, let's just split them into two interfaces right away. Let's stick to `UiReceiver` and `HintReceiver` - those are good names. I'd prefer to avoid merging the half-way solution if we are already touching this part.

paging/paging-common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt
```diff
@@ -300,7 +317,7 @@ public abstract class PagingDataDiffer<T : Any>(
      * @sample androidx.paging.samples.refreshSample
      */
     public fun refresh() {
-        receiver?.refresh()
+        statelessReceiver?.refresh()
```
dlam: For when we drop these signals on the floor, I wonder if we should throw an exception. Can we file a bug to follow up on that as a new FR?

claraf3: zzz had to file it under my personal email, b/234749375

paging/paging-common/src/main/kotlin/androidx/paging/PagingDataDiffer.kt
claraf3: done

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -163,6 +163,339 @@ class PagingDataDifferTest(
...
+        // second generation
+        val pageEventCh = Channel<PageEvent<Int>>(Channel.UNLIMITED)
+        val hintReceiver2 = HintReceiverFake()
+        val job2 = launch {
+            differ.collectFrom(
+                PagingData(pageEventCh.consumeAsFlow(), dummyUiReceiver, hintReceiver2)
+            )
+        }
+
+        // we send the initial load state. this should NOT cause second gen receiver to register
```
ianhanniballake: This should be `second gen hint receiver to register`, correct?

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -163,6 +163,339 @@ class PagingDataDifferTest(
...
+        pageEventCh.trySend(
+            localLoadStateUpdate(refreshLocal = Loading)
+        )
+        assertThat(differ.loadStateFlow.first()).isEqualTo(
+            localLoadStatesOf(refreshLocal = Loading)
+        )
+
```
ianhanniballake, line 222: Can we add another check of the preconditions here before we call `differ[4]`?

```
assertThat(hintReceiver1.hints).hasSize(0)
assertThat(hintReceiver2.hints).hasSize(0)
```

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -163,6 +163,339 @@ class PagingDataDifferTest(
...
+        assertThat(differ.snapshot()).containsAtLeastElementsIn(listOf(1, 2, 3))
+
+        val uiReceiver = UiReceiverFake()
+        val pagingData2 = infinitelySuspendingPagingData(uiReceiver = uiReceiver)
+        val job2 = launch { differ.collectFrom(pagingData2) }
+        assertTrue(job2.isActive)
+
+        // even though the second paging data never presented, it should be receiver of the refresh
+        differ.refresh()
+        assertThat(pagingData1.uiReceiver).isEqualTo(PagingData.NOOP_UI_RECEIVER)
```
ianhanniballake: This is always going to be true, right? This seems to be testing the internals of `PagingData.from` and doesn't actually verify that refresh was never called on `pagingData1.uiReceiver1`. I think our check for `hasSize(1)` below this line is enough since we know that a `refresh()` can only be received at most once.

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -163,6 +163,339 @@ class PagingDataDifferTest(
...
+        assertThat(differ.snapshot()).containsAtLeastElementsIn(listOf(1, 2, 3))
+
+        val uiReceiver = UiReceiverFake()
+        val pagingData2 = infinitelySuspendingPagingData(uiReceiver = uiReceiver)
+        val job2 = launch { differ.collectFrom(pagingData2) }
+        assertTrue(job2.isActive)
+
+        // even though the second paging data never presented, it should be receiver of the retry
+        differ.retry()
+        assertThat(pagingData1.uiReceiver).isEqualTo(PagingData.NOOP_UI_RECEIVER)
```
ianhanniballake: Ditto here for removing this assert that is always going to be true.

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -1577,10 +1922,28 @@ class PagingDataDifferTest(
                     ).also { pagingSources.add(it) }
                 }
             ),
-        block: (SimpleDiffer, TestDispatcher, MutableList<TestPagingSource>) -> Unit
+        block: (
+            SimpleDiffer,
+            TestDispatcher,
+            MutableList<TestPagingSource>,
+            MutableList<TrackableUiReceiverWrapper>,
+            MutableList<TrackableHintReceiverWrapper>
```
ianhanniballake: This could be `hintReceivers: List<TrackableHintReceiverWrapper>` to give autocomplete to Android Studio on what these variables should be named.

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -1577,10 +1922,28 @@ class PagingDataDifferTest(
                     ).also { pagingSources.add(it) }
                 }
             ),
-        block: (SimpleDiffer, TestDispatcher, MutableList<TestPagingSource>) -> Unit
+        block: (
+            SimpleDiffer,
+            TestDispatcher,
+            MutableList<TestPagingSource>,
```
ianhanniballake: This and the other `MutableList` parameters can be exposed as a `List`. I don't think we need any `block` to be able to modify these lists :)

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: done

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: done

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: done

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: done

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: done

ianhanniballake (approved):

PR #418, opened 2022-06-20
Update Gradle Enterprise plugins

jprinet:


tikurahul (approved):

tikurahul (commented):
You might need to import these libraries as prebuilts using `development/importMaven`.

tikurahul (changes requested):
You need to add prebuilts.

jprinet:
> You might need to import these libraries as prebuilts using `development/importMaven`.

Ok, it's not only the verification-metadata, just read about the prebuilt follow up to your comment.
I don't have time to handle that at the moment.

Let's put this on hold then

PR #419, opened 2022-06-20
Have RxPagedListBuilder emit errors in RxJava chain

veyndan:
## Proposed Changes

From https://issuetracker.google.com/issues/233703110:

> In Paging 2.1.2, when an error is thrown from the `DataSource.Factory`, the error is propagated through the rx chain, so we could use things like `onErrorResumeNext(…)` when there's an error so we can fail safely. In Paging 3, this error is no longer propagated through the rx chain, so our callback in `onErrorResumeNext(…)` is never invoked.

I've just wrapped the invalidate job contents with a try-catch. This is a greatly simplified version of what [`rxObservable` does](https://github.com/Kotlin/kotlinx.coroutines/blob/2ee99e269ec37d703f44f2598b4634637d8354d6/reactive/kotlinx-coroutines-rx3/src/RxObservable.kt#L125-L139).

Note: the diff in GitHub looks like many things changed in `RxPagedListBuilder`, but checking out the diff in Android Studio shows that I did in fact just wrap it all in a try-catch.

## Testing

Test: ./gradlew paging:paging-rxjava2:test paging:paging-rxjava2:test

## Issues Fixed

Fixes: 233703110

### DIFF
--- a/paging/paging-rxjava2/src/main/java/androidx/paging/RxPagedListBuilder.kt
+++ b/paging/paging-rxjava2/src/main/java/androidx/paging/RxPagedListBuilder.kt
@@ -388,49 +388,53 @@ class RxPagedListBuilder<Key : Any, Value : Any> {
 
             currentJob?.cancel()
             currentJob = GlobalScope.launch(fetchDispatcher) {
-                currentData.pagingSource.unregisterInvalidatedCallback(callback)
-                val pagingSource = pagingSourceFactory()
-                pagingSource.registerInvalidatedCallback(callback)
-                if (pagingSource is LegacyPagingSource) {
-                    pagingSource.setPageSize(config.pageSize)
-                }
+                try {
+                    currentData.pagingSource.unregisterInvalidatedCallback(callback)
+                    val pagingSource = pagingSourceFactory()
+                    pagingSource.registerInvalidatedCallback(callback)
+                    if (pagingSource is LegacyPagingSource) {
+                        pagingSource.setPageSize(config.pageSize)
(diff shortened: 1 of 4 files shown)


claraf3:
Hi, thanks for the PR. Would you be able to update the original bug `https://issuetracker.google.com/issues/233703110` with a sample project so we can repro your use case? It would help us evaluate whether it is a bug and how we might resolve it. 

Closing this PR in the meantime but we'll keep an eye on the bug for when you update it. Thank you.

veyndan:
Hi @claraf3,

In https://issuetracker.google.com/issues/233703110#comment1 I'd already linked a failing test which showcases the problem. The link is to this [file](https://github.com/veyndan/androidx-paging-2-to-3-regressions/blob/master/src/test/kotlin/com/veyndan/Reports.kt), mainly because the only substantive changes in the repo is that file, but you should be able to clone https://github.com/veyndan/androidx-paging-2-to-3-regressions and run the tests locally. By toggling between Paging 2 (e.g., v2.1.2) and Paging 3 (e.g., v3.1) in the [`build.gradle.kts`](https://github.com/veyndan/androidx-paging-2-to-3-regressions/blob/ab54887ce19a199b66143631d9915ee73351ecbf/build.gradle.kts#L21-L22) file, you'll be able to see that the tests pass when using Paging 2, but fail when using Paging 3.

Please let me know if there were troubles recreating the issue on your end.

veyndan:
Hey @claraf3, is there anything else that I can do to move this along?

PR #424, opened 2022-06-23
invalidate cache

yigit:
Invalidate cache key to workaround the serialization issue in gradle.

Bug: n/a
Test: CI


tikurahul (approved):

liutikas:
Do we want to report to Gradle with some scans of what the broken configuration cache looks like here?

yigit:
> Do we want to report to Gradle with some scans of what the broken configuration cache looks like here?

Yep, working on it in a tab

PR #428, opened 2022-07-10
Support for queries with > 999 host parameters

mzgreen:
## Proposed Changes

This change adds support for large SQLite queries. Large queries are queries with more than 999 host parameters. 

Prior to this change, if someone created a query like:
```
@Query("DELETE * FROM user WHERE id IN (:ids)")
fun delete(ids: List<Int>)
```
and invoked this function passing a list larger than 999 items, the library would throw a `SQLiteException`.

The implementation details and codegen changes are documented in the design doc: https://docs.google.com/document/d/1tQPjWLmBS55Cu-1lxa3Nn-eUaOCT6HCnftsmOo_NEfo/edit?usp=sharing

### Limitations

The solution supports everything except:
 - `@RawQuery` - it's not possible to support raw queries, because the implemented solution relies on query parsing which happens during compilation while raw queries are executed at runtime.
 - `Cursor` returning dao methods - The implemented solution uses temp tables which have to be closed as soon as the dao method finishes execution. Cursor is lazy, so the underlying query is executed when the Cursor is accessed for the first time at which point the temp table probably no longer exists.
 - Paging - it may be possible to support it, but I don't have enough experience with Paging to tell and the change got quite big already so I decided to skip this part for now.

## Testing

Test: Added new large query unit tests tests and a benchmark test

## Issues Fixed

Fixes: b/73634057

RelNote: Room now supports queries with more than 999 host parameters.

### DIFF
--- a/room/benchmark/src/androidTest/java/androidx/room/benchmark/LargeQueryBenchmark.kt
+++ b/room/benchmark/src/androidTest/java/androidx/room/benchmark/LargeQueryBenchmark.kt
@@ -1,0 +1,131 @@
+/*
+ * Copyright 2022 The Android Open Source Project
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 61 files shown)


mzgreen:
Not sure why FTL tests are failing on CI, `connectedCheck` and `buildOnServer` are passing for me locally.

mzgreen:
Pinging @yigit @danysantiago @droid-wan-kenobi

danysantiago:
Thanks for sending this, looks promising! I am a bit busy this week but will try to take a look as soon as I can.

mzgreen:
Looks like Room is getting a huge Kotlin rewrite right now so it may not be the best time for such contribution?

PR #435, opened 2022-07-24
Fix shouldShowRequestPermissionRationale causing memory leak in Andro…

getActivity:
## Proposed Changes

Hello, there will be a memory leak when calling `activity.shouldShowRequestPermissionRationale` in Android 12. The problem is fixed in Android 12L, but it is still a legacy in Android 12. It is recommended to reflect the `application.shouldShowRequestPermissionRationale` method to avoid this problem.

## Testing

Test: manually ran with LeakCanary

```java
ActivityCompat.shouldShowRequestPermissionRationale(MainActivity.this, Manifest.permission.CAMERA);
finish();
```

* Use leakcanary to view memory leaks

* Or use profiler to view memory leaks

Please see

[IssueTracker 212113500](https://issuetracker.google.com/issues/212113500)

[IssueTracker 207946473](https://issuetracker.google.com/issues/207946473)

[IssueTracker 204539732](https://issuetracker.google.com/issues/204539732)

## Issues Fixed

Fixes: b/212113500

Fixes: b/207946473

Fixes: b/204539732


core/core/src/main/java/androidx/core/app/ActivityCompat.java
```diff
@@ -752,6 +756,19 @@ private Api31Impl() {
         static boolean isLaunchedFromBubble(@NonNull final Activity activity)  {
             return activity.isLaunchedFromBubble();
         }
+
+        @DoNotInline
+        static boolean shouldShowRequestPermissionRationale(Activity activity, String permission) {
+            try {
+                // Fix shouldShowRequestPermissionRationale causing memory leak in Android 12
```
liutikas: Add a link to the leak in question.

Which version was it fixed in?

core/core/src/main/java/androidx/core/app/ActivityCompat.java
```diff
@@ -553,7 +555,9 @@ public void run() {
      */
     public static boolean shouldShowRequestPermissionRationale(@NonNull Activity activity,
             @NonNull String permission) {
-        if (Build.VERSION.SDK_INT >= 23) {
+        if (Build.VERSION.SDK_INT == 31) {
```
liutikas: I assume we dont want this workaround in perpetuity (SDK_INT > 31). Should this first check for SDK_INT >= 32 can call in without a workaround?

core/core/src/main/java/androidx/core/app/ActivityCompat.java
liutikas: Also add a comment why calling this via reflection  fixes the issue.

core/core/src/main/java/androidx/core/app/ActivityCompat.java
getActivity: > I assume we dont want this workaround in perpetuity (SDK_INT > 31). Should this first check for SDK_INT >= 32 can call in without a workaround?

SDK_INT >= 32 does not have this problem, because this problem has been fixed in Android 12L

core/core/src/main/java/androidx/core/app/ActivityCompat.java
getActivity: > Add a link to the leak in question.
> 
> Which version was it fixed in?
> 
> Also add a comment why calling this via reflection fixes the issue.

Sorry, my English is not good, would like to confirm, should I add the answer to the issue comment or the code comment?

core/core/src/main/java/androidx/core/app/ActivityCompat.java
liutikas: Does this mean this code should look like 

```
if Build.VERSION.SDK_INT >= 32 {
             return Api32Impl.shouldShowRequestPermissionRationale(activity, permission);
} else if if (Build.VERSION.SDK_INT == 31) {
            return Api31Impl.shouldShowRequestPermissionRationale(activity, permission);
} else if (Build.VERSION.SDK_INT >= 23) {
            return Api23Impl.shouldShowRequestPermissionRationale(activity, permission);
}
```
?

core/core/src/main/java/androidx/core/app/ActivityCompat.java
liutikas: Yes, please leave a comment with links to the bugs that we are trying to work around.

getActivity: > Add a link to the leak in question.
> 
> Which version was it fixed in?
>
> Also add a comment why calling this via reflection fixes the issue.

 1. Background of the problem：Fix shouldShowRequestPermissionRationale causing memory leak in Android 12，
    this problem has been fixed on the Android 12L system, but it is still a historical problem for the Android 12 system
 2. The reason for the problem: The culprit is that the PermissionUsageHelper holds the Context object as a field,
    and calls AppOpsManager.startWatchingStarted in the constructor to start the monitoring,
    so that the PermissionUsageHelper object will be added to the AppOpsManager#mStartedWatchers collection,
    which will lead to active calls in the Activity When finishing, stopWatchingStarted is not used to remove the watch,
    which causes the Activity object to be held in the AppOpsManager#mStartedWatchers collection,
    which indirectly causes the Activity object to not be recycled by the system.
 3. The solution to the problem: The handling of this problem is also very simple and rude, that is,
    to replace the Context parameter passed in the outer layer from the Activity object to the Application object,
    because the Application life cycle is relatively long, but there is only the shouldShowRequestPermissionRationale method in the Activity,
    and What if there is no such method in Application? Looking at the implementation of this method, in fact,
    that method will eventually call the PackageManager.shouldShowRequestPermissionRationale method (hidden API, but not in the blacklist),
    so as long as the PackageManager object can be obtained, and finally use reflection to execute this method , which avoids memory leaks.

getActivity: @liutikas Hello, issue comments and code comments for this piece have been added.

core/core/src/main/java/androidx/core/app/ActivityCompat.java
getActivity: @liutikas Hello, the code has been modified as required.

alanv (approved):

alanv:
Presubmit is unhappy because there's no `Test:` stanza in the commit message.

Please update your PR message to use `Bug: <bug-id>` or `Fixes: <bug-id>` for the bug numbers addressed and `Test: <class-name>` for the test added, as in https://github.com/androidx/androidx/pull/428

getActivity:
@alanv Thanks for the reminder, the revision has been completed

alanv:
Still needs a `Test:` stanza. Given that we're not adding LeakCanary to the CI tests, I think `Test: manually ran with LeakCanary` is appropriate.

getActivity:
> Still needs a `Test:` stanza. Given that we're not adding LeakCanary to the CI tests, I think `Test: manually ran with LeakCanary` is appropriate.

@alanv Hello, I have made the modification here as required.

getActivity:
@liutikas @alanv Hello, what is the problem with the merge code?

getActivity:
@tikurahul @alanv @liutikas Hello, this problem seems to be difficult to solve, do I need to initiate a new pull request?

alanv:
If @dlam and @tikurahul can't get this working, our GitHub workflow is busted. Let's keep pushing on this PR.

tikurahul:
I added some additional logging in the `check-pr-format-action`. @getActivity Did you actually change the commit message to include the `Test` stanza as well ? GitHub lets you change it _after_ you make the PR, but that does not change the underlying commit message. One thing that might help is to `git commit --amend` with the `Test` stanza and do a `force push` so the action is forced to pick up all the new changes.

getActivity:
@tikurahul Hello, I have modified the commit message as required, can you take a look again?

tikurahul:
Kicking off another workflow run ! It moved past the point of last failure. So its promising 🤞🏽

getActivity:
image

getActivity:
@tikurahul Hello, I found a code check failure here, will this affect code merging?

dlam:
> @tikurahul Hello, I found a code check failure here, will this affect code merging?

```
> Task :core:core:lintDebug FAILED
/home/runner/work/androidx/androidx/core/core/src/main/java/androidx/core/app/ActivityCompat.java:830: Error: Calling Method.invoke without an SDK check [BanUncheckedReflection from androidx.build]
                return (boolean) method.invoke(packageManager, permission);
                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

   Explanation for issues of type "BanUncheckedReflection":
   Jetpack policy discourages reflection. In cases where reflection is used on
   platform SDK classes, it must be used within an SDK_INT check that
   delegates to an equivalent public API on the latest version of the
   platform. If no equivalent public API exists, reflection must not be used.
```

This is a real error that you need to fix. @alanv can probably advise on best way to go about it.

getActivity:
@alanv Hello, do you have any good advice on this issue?

getActivity:
@tikurahul @dlam @liutikas Do you guys have any good solutions?

getActivity:
@alanv @tikurahul @dlam @liutikas Hi, I see `@SuppressLint("BanUncheckedReflection")` is used in AndroidX code to suppress reflection warnings, can I add `@SuppressLint("BanUncheckedReflection")` to avoid this issue?

alanv:
> @alanv @tikurahul @dlam @liutikas Hi, I see `@SuppressLint("BanUncheckedReflection")` is used in AndroidX code to suppress reflection warnings, can I add `@SuppressLint("BanUncheckedReflection")` to avoid this issue?

Yes, this can be suppressed since the call site is wrapped with an SDK_INT check.

getActivity:
@alanv Hello, I have submitted the modification of this part of the code

getActivity:
@alanv Hello, please check whether there are any problems with the changes, if there are any, please point out, if not, please approve them

PR #437, opened 2022-08-08
Bump playground to gradle 7.5 stable

dlam:
Test: ./gradlew wrapper --gradle-version=7.5 --distribution-type=bin --gradle-distribution-sha256-sum=cb87f222c5585bd46838ad4db78463a5c5f3d336e5e2b98dc7c0c586527351c2
Change-Id: I9cc0f278a2679f6e82b7513a3319c486abc48927


liutikas (approved):
It would be nice if we also reenabled the check for version match.

dlam:
> It would be nice if we also reenabled the check for version match.

I think the fix for the blocking bug landed in 7.6-rc1 so seems like we need to wait for that?

PR #448, opened 2022-09-10
Reformat the source to use Unix line breaks always

adaext:
- Source file had line breaks mixed in Windows format and Unix format.
  - Reformat the file using dos2unix command.
 

alanv:
I'll revert the original change from Gerrit, then re-apply it with correct line breaks. Otherwise the file history gets all messed up.

alanv:
Given that you were the only author, this should be fine to take instead of a revert.

alanv (approved):

adaext:
> Given that you were the only author, this should be fine to take instead of a revert.

Ok, thank you!

PR #460, opened 2022-11-04
Fix flaky testDialogFragmentDismiss

sanura-njaka:
Fix flaky on testDialogFragmentDismiss
    
DialogFragmentDismissTest.testDialogFragmentDismiss
is flaking with a leak found in b/157955883.
Since this is flaky on APIs lower than 31,
want to early return to prevent flakiness.
    
Bug: 256188823
Test: ./gradlew bOS
Change-Id: I1f15b75f8a25c5622fe21656b587928d47bff9c8


dlam:
Closing since this fix was merged separately.

PR #462, opened 2022-11-09
Treat kotlinx.collections.immutable.ImmutableCollection as stable

mcpiroman:
This will make parameters that have more general type of `ImmutableCollection` rather than List or Set stable.


jimgoog:
👍 Seems reasonable to me.  CLA required, see instructions from CLA bot.  Also, lint is complaining that there is no unit test stanza, you can amend your commit message to say `Test: unit tests` that will make lint happy.

mcpiroman:
Done.

PR #468, opened 2022-11-30
Use projectOrArtifact for annotation:annotation in core playground

dlam:
Test: cd core && ./gradlew bOS
Change-Id: Iff00ec1c8d65937f7e449561bd2771e24001d50b


liutikas:
I think you also need to bump the snapshot build id. Alternatively, we can include annotation into core playground.

liutikas (approved):

dlam:
Hmm so it is failing because it is expecting kmp artifacts for annotations in snapshots :|

PR #470, opened 2022-12-05
Fix up core dependencies on annotation

dlam:
We are inconsistently including parts of annotation as project or snapshot dependencies. There also seems to be a bug with configuration caching and task dependency, which is causing annotations to not get built. This change aligns all annotation deps to be projectOrArtifact, except for annotation-sampled, which is not published.

Test: cd core && ./gradlew bOS
Change-Id: Ie606e42d48a0f88fb8f32fa38357fea28045b74d


yigit (approved):
we really need a repro for this as it can happen in any other project.
+2ing to unblock.

PR #479, opened 2022-12-29
Remove mismatched androidTest dependency already provided by implementation

dlam:
Fixes a "resolution ambiguity" error with lifecycle-runtime-ktx which  is broken on playground as androidTest resolves a different version of the artifact than the snapshot, but on AOSP both resolve to the project successfully, which works around the problem.

Test: cd appcompat && ./gradlew appcompat:appcompat:compileDebugAndroidTestKotlin
Change-Id: I3c4b8cc388d8a6c38c38f30ae5bbd955f613a755

### DIFF
--- a/appcompat/appcompat/build.gradle
+++ b/appcompat/appcompat/build.gradle
@@ -48,7 +48,14 @@ dependencies {
     androidTestImplementation(libs.espressoCore, excludes.espresso)
     androidTestImplementation(libs.mockitoCore, excludes.bytebuddy) // DexMaker has it's own MockMaker
     androidTestImplementation(libs.dexmakerMockito, excludes.bytebuddy) // DexMaker has it's own MockMaker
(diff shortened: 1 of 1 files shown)


appcompat/appcompat/build.gradle
```diff
@@ -48,7 +48,9 @@ dependencies {
     androidTestImplementation(libs.espressoCore, excludes.espresso)
     androidTestImplementation(libs.mockitoCore, excludes.bytebuddy) // DexMaker has it's own MockMaker
     androidTestImplementation(libs.dexmakerMockito, excludes.bytebuddy) // DexMaker has it's own MockMaker
-    androidTestImplementation("androidx.lifecycle:lifecycle-runtime-testing:2.5.1" )
+    androidTestImplementation("androidx.lifecycle:lifecycle-runtime-testing:2.5.1", {
```
yigit, line 51: please add a clarifying comment so someone doesn't remove this.

yigit (approved):

PR #485, opened 2023-02-15
Update repeatable annotation test to test repeatable kotlin annotations

oas004:
## Proposed Changes

  - Since https://youtrack.jetbrains.com/issue/KT-12794 is fixed, we can now update the repeatable annotation tests to check that getAllAnnotations method works for repeatable Kotlin annotations as well.
 - Added unit test to check that non repeatable annotation also work with the created repeatable Kotlin annotation.

## Testing

Test: Used 
- ./gradlew :room:room-compiler-processing:test --tests "androidx.room.compiler.processing.XAnnotationTest.repeatableAnnotation[preCompiled_false]"
- ./gradlew :room:room-compiler-processing:test --tests "androidx.room.compiler.processing.XAnnotationTest.kotlinRepeatableAnnotation_notRepeated[preCompiled_false]"

### DIFF
--- a/room/room-compiler-processing/src/test/java/androidx/room/compiler/processing/XAnnotationTest.kt
+++ b/room/room-compiler-processing/src/test/java/androidx/room/compiler/processing/XAnnotationTest.kt
@@ -28,6 +28,7 @@ import androidx.room.compiler.processing.testcode.JavaEnum
 import androidx.room.compiler.processing.testcode.MainAnnotation
 import androidx.room.compiler.processing.testcode.OtherAnnotation
 import androidx.room.compiler.processing.testcode.RepeatableJavaAnnotation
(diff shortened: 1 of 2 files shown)


oas004:
@liutikas or @danysantiago Is it normal that these pipelines fails like this? Some of them say that I have a Kotlin version that is not compatible with the Compose version. Do I need to take any action on this? :)

dlam:
> @liutikas or @danysantiago Is it normal that these pipelines fails like this? Some of them say that I have a Kotlin version that is not compatible with the Compose version. Do I need to take any action on this? :) 

Our build on GitHub is broken ATM, so you may ignore for now :/

danysantiago (approved):
Looks good, thanks!

PR #487, opened 2023-02-21
Contract TypedArray.use invocation exactly once

Goooler:
## Proposed Changes

  - If I want to initialize final fields in `TypedArray.use` in a View's constructor, there will be a compiler error: `Captured member values initialization is forbidden due to possible reassignment`, in this case, we should specify the lambda is invoked [EXACTLY_ONCE](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.contracts/-invocation-kind/-e-x-a-c-t-l-y_-o-n-c-e.html).
  - Repro: https://github.com/Goooler/DemoApp/pull/343
  - Ref: https://youtrack.jetbrains.com/issue/KT-42044

## Testing

Test: cd core && ./gradlew core:core-ktx:connectedCheck

### DIFF
--- a/core/core-ktx/src/androidTest/java/androidx/core/content/res/TypedArrayTest.kt
+++ b/core/core-ktx/src/androidTest/java/androidx/core/content/res/TypedArrayTest.kt
@@ -16,7 +16,11 @@
 
 package androidx.core.content.res
 
+import android.content.Context
+import android.content.res.TypedArray
 import android.graphics.Color
+import android.util.AttributeSet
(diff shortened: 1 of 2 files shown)


JakeWharton:
Note that contracts were intentionally omitted from all ktx because of their experimental nature and it was unclear as to what (if any) compatibility guarantees there were. 

https://issuetracker.google.com/issues/122663367

Goooler:
Thanks for your explanation! I was guessing this behavior is intended.

PR #489, opened 2023-02-25
Do not expose kotlin stdlib to the consumers

yukukotani:
## Proposed Changes

This PR proposes to declare Kotlin stdlib with `implementation` instead of `api` in the Health Connect SDK's build configuration.

Currently, the Kotlin stdlib that Health Connect SDK depends on appears in the compile classpath of our app. This results in [class duplication](https://d.android.com/r/tools/classpath-sync-errors).

## Testing

Test: Our app using Health Connect SDK works well.

## Issues Fixed

Fixes: N/A

### DIFF
--- a/health/connect/connect-client/build.gradle
+++ b/health/connect/connect-client/build.gradle
@@ -31,7 +31,7 @@ BundleInsideHelper.forInsideAar(
 )
 
 dependencies {
-    api(libs.kotlinStdlib)
+    implementation(libs.kotlinStdlib)
     // Add dependencies here
     api("androidx.activity:activity:1.2.0")
     api("androidx.annotation:annotation:1.2.0")


liutikas:
What kind of class duplication are you hitting? Is it https://youtrack.jetbrains.com/issue/KT-55297 ? If so, the fix is described in the issue tracker. The bug is due to jetbrains moving classes between artifacts without proper gradle constraints

yukukotani:
@liutikas Oh yes, my error is the same as [KT-55297](https://youtrack.jetbrains.com/issue/KT-55297). Although it's the root problem, using `api` solves the error, right?

FYI, most of packages facing app developers are using `api` instead of `implementation`.
https://github.com/search?q=repo%3Aandroidx%2Fandroidx+%22libs.kotlinStdlib%22&type=code

liutikas:
> @liutikas Oh yes, my error is the same as [KT-55297](https://youtrack.jetbrains.com/issue/KT-55297). Although it's the root problem, using `api` solves the error, right?

It does not solve really it. The same issue will be seen with this as well. The only difference between api vs implementation is that one is on a compile classpath and the other is on runtime classpath.

> 
> FYI, most of packages facing app developers are using `api` instead of `implementation`. https://github.com/search?q=repo%3Aandroidx%2Fandroidx+%22libs.kotlinStdlib%22&type=code

Sure, this is not necessarily a bad change, but it will not fix your issue in general

PR #499, opened 2023-03-07
Convert java.lang.Math#random to kotlin.random.Random

veyndan:
Bug: 270612487


dlam:
fwiw, Kotlin has its own Random impl, but should be fine.

ianhanniballake (approved):

PR #503, opened 2023-03-15
Fix android 12 issue

ramyfarah:
## Proposed Changes

  -
  -
  -

## Testing

## Issues Fixed


tikurahul:
What is the problem that you are trying to address ?
This alarm does not use `FLAG_IMMUTABLE` because that has a side-effect on the identify of the instance of `PendingIntent` when subsequently retrieving it with `FLAG_NO_CREATE`.
It's okay to do that here, given this `PendingIntent` is not actually shared (with other apps).

tikurahul (changes requested):
This is not something we want to do.

PR #508, opened 2023-03-20
[Navigation doc] differenciate route vs route pattern

martinbonnin:
_This is a docs-only change_

This PR tries to highlight the difference between :

* "route":  a static String linked to a single destination like `"topic/{topicId}"`
* "route instance": a instance of a route that has arguments filled in like `"topic/42"`

Since changing argument names would be a breaking change, this PR tweaks the KDoc. 

Test: N/A docs change

### DIFF
--- a/navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.kt
+++ b/navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.kt
@@ -497,7 +497,7 @@ public open class NavController(
      * Attempts to pop the controller's back stack back to a specific destination.
      *
      * @param route The route of the topmost destination to retain. Both route patterns
-     * and route instances with filled in arguments are supported.
+     * and routes with filled arguments are supported.
      * @param inclusive Whether the given destination should also be popped.
      * @param saveState Whether the back stack and the state of all destinations between the
      * current destination and the [route] should be saved for later
@@ -582,7 +582,7 @@ public open class NavController(
      * **not** handle calling [dispatchOnDestinationChanged]
      *
      * @param route The route of the topmost destination to retain. Both route patterns
-     * and route instances with filled in arguments are supported.
+     * and routes with filled arguments are supported.
      * @param inclusive Whether the given destination should also be popped.
      * @param saveState Whether the back stack and the state of all destinations between the
      * current destination and the destination with [route] should be saved for later to be
@@ -771,8 +771,8 @@ public open class NavController(
      * via [popBackStack] when using a `saveState` value of `true`.
      *
      * @param route The route of the destination previously used with [popBackStack] with a
-     * `saveState` value of `true`. Both route patterns and route instances with filled in
-     * arguments are supported.
+     * `saveState` value of `true`. Both route patterns and routes with filled arguments
+     * are supported.
      *
      * @return true if the saved state of the stack associated with [route] was cleared.
      */
@@ -1901,7 +1901,7 @@ public open class NavController(
         return if (backStackMap.containsKey(id)) {
             restoreStateInternal(id, null, null, null)
         } else {
-            // if it didn't match, it means the route contains filled in arguments and we need
+            // if it didn't match, it means the route contains filled arguments and we need
             // to find the destination that matches this route's general pattern
             val matchingDestination = findDestination(route)
             check(matchingDestination != null) {
@@ -2144,7 +2144,7 @@ public open class NavController(
      * Navigate to a route in the current NavGraph. If an invalid route is given, an
      * [IllegalArgumentException] will be thrown.
      *
-     * @param route route instance with filled arguments for the desired destination
+     * @param route pattern or route with filled arguments for the destination
      * @param builder DSL for constructing a new [NavOptions]
      *
      * @throws IllegalArgumentException if the given route is invalid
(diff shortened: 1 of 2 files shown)


navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.kt
```diff
@@ -496,8 +496,8 @@ public open class NavController(
     /**
      * Attempts to pop the controller's back stack back to a specific destination.
      *
-     * @param route The topmost destination to retain. May contain filled in arguments as long as
-     * it is exact match with route used to navigate.
+     * @param route The route of the topmost destination to retain. Both route patterns
+     * and route instances with filled in arguments are supported.
```
claraf3: The use of "route pattern" and "route" here should be distinctive enough. Can you also clarify that if using route, it must match the original route that was used to navigate?

navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.kt
```diff
@@ -581,7 +581,8 @@ public open class NavController(
      * Attempts to pop the controller's back stack back to a specific destination. This does
      * **not** handle calling [dispatchOnDestinationChanged]
      *
-     * @param route The topmost destination with this route to retain
+     * @param route The route of the topmost destination to retain. Both route patterns
+     * and route instances with filled in arguments are supported.
```
claraf3: Same as above, lets stick to "route pattern" and "route".

navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.kt
```diff
@@ -770,8 +771,8 @@ public open class NavController(
      * via [popBackStack] when using a `saveState` value of `true`.
      *
      * @param route The route of the destination previously used with [popBackStack] with a
-     * `saveState` value of `true`. May contain filled in arguments as long as
-     * it is exact match with route used with [popBackStack].
+     * `saveState` value of `true`. Both route patterns and route instances with filled in
+     * arguments are supported.
```
claraf3: Same comment as with `popBackStack` and likewise for `getBackStackEntry`.

navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.kt
martinbonnin: Agree on "route pattern" and "route" 👍 

I removed the `May contain filled in arguments as long as it is exact match with route used to navigate.` because I felt like this was the general expectation of the method and didn't add much information. Whether using a route pattern or a route, the expectation is that it returns true if found or false if not. Or did I get this wrong?

navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.kt
```diff
@@ -2143,7 +2144,7 @@ public open class NavController(
      * Navigate to a route in the current NavGraph. If an invalid route is given, an
      * [IllegalArgumentException] will be thrown.
      *
-     * @param route route for the destination
+     * @param route route instance with filled arguments for the desired destination
```
claraf3: It accepts both route with filled arguments or the route patterns. 
Given a destination with route "route/{arg}" you can navigate with either "route/{arg}" or "route/myArg".

Suggest: "route with filled arguments or the route pattern for the destination".

navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.kt
```diff
@@ -2157,7 +2158,7 @@ public open class NavController(
...
```
claraf3: same comment as line#2147

claraf3 (commented):
Thanks for the PR. Commented.

navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.kt
claraf3: `returns true if found or false if not` is indeed the case. In the past we matched based on route pattern and didn't differentiate between "route/myArg1" and "route/myArg2". Now that matching takes into account filled in arguments, matching behavior has changed and what was previously expected to match may no longer match. Therefore the clarification could be important. Also, it doesn't hurt to have more docs than less :).

martinbonnin: Kind of disagree. KDoc is a place for reference information and changes belong in the changelog. Also more docs can hurts. In my personal case, this sentence got me puzzled about whether this would throw if passed the wrong route or how come there was an extra sentence in addition to the return value.

I don't want to die on this above hill though! If you think it is important to keep the sentence, please confirm and I'l re-add it (or feel free to edit directly before merging too!)

dlam:
You can safely ignore the lint failure here - it's on our end.

navigation/navigation-runtime/src/main/java/androidx/navigation/NavController.kt
martinbonnin: Oh nice, TIL! I tweaked your suggestion a bit to put "route pattern or route" (always put route pattern first for consistency)

martinbonnin:
Changes pushed. I'm not 100% sure about `filled in arguments` vs just `filled arguments`. I put `filled arguments` for consistency but maybe `filled in` is more idiomatic?

martinbonnin:
2025 cleanup

PR #509, opened 2023-03-21
gradle-enterprise-plugin:3.12.4 and common-custom-user-data-gradle-pl…

mathjeff:
…ugin:1.9

Bug: 274622117

Test: Treehugger runs busytown/*.sh
Test: ./gradlew help
Test: ./gradlew help --scan # and see that a scan is generated and still sets custom values
Test: cd fragment && ./gradlew help --scan # and see that a scan is generated and still sets custom values

Change-Id: I93e3270ec29dddbdca43b1b4f68685c1f8f4db38


liutikas (approved):

mathjeff:
submitted as aosp/2498900

PR #511, opened 2023-03-29
Drop transitive dependency on guava

LouisCAD:
See https://issuetracker.google.com/issues/275580742

## Proposed Changes

  - Don't force the huge and unnecessary Guava dependency
  - Use the dedicated androidx.concurrent-futures-ktx library instead

## Testing

Test: I excluded the kotlinx.coroutines-guava dependency in my project, replaced it with androidx.concurrent-futures-ktx, updated the import where I used the `await()` extension, and it kept working, as expected.

## Issues Fixed

Fixes: https://issuetracker.google.com/issues/275580742

### DIFF
--- a/wear/wear-remote-interactions/build.gradle
+++ b/wear/wear-remote-interactions/build.gradle
@@ -26,8 +26,7 @@ dependencies {
     api("androidx.annotation:annotation:1.1.0")
     api(libs.kotlinStdlib)
     api(libs.guavaListenableFuture)
-    api("androidx.concurrent:concurrent-futures-ktx:1.0.0")
-    implementation("androidx.concurrent:concurrent-futures:1.0.0")
+    api("androidx.concurrent:concurrent-futures-ktx:1.1.0")
 
     androidTestImplementation(libs.testExtJunit)
     androidTestImplementation(libs.testCore)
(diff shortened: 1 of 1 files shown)


yigit:
Thanks for sending this. Unfortunately, wear is not in the list of projects we support on github.
https://github.com/androidx/androidx/blob/androidx-main/CONTRIBUTING.md

I can approve this to copy the PR to AOSP if that helps but if you are familiar w/ gerrit/AOSP'; it would be better to create a CL there.

lmk what you prefer.

yschimke:
@yigit if you approve, let me know if I can help shepherding it through, once created.

LouisCAD:
I'm not familiar with gerrit/AOSP yet, so please copy it for the time being @yigit, thank you 🙏🏼

yigit (approved):
Approving to trigger aosp sync

yigit:
@yschimke , i'll add you as a reviewer on both AOSP CLs. thank you.

yschimke:
Failing on

```
Could not determine the dependencies of task ':wear:wear-remote-interactions:compileDebugAndroidTestJavaWithJavac'.
> Could not resolve all task dependencies for configuration ':wear:wear-remote-interactions:debugAndroidTestCompileClasspath'.
   > Could not find org.jetbrains.kotlinx:kotlinx-coroutines-core:1.3.4.
     Required by:
         project :wear:wear-remote-interactions > androidx.concurrent:concurrent-futures-ktx:1.1.0
```

Not sure why.

yschimke:
It does seem like this artifact is special (a pom with redirects to -jvm), maybe throw in a `implementation(libs.kotlinCoroutinesCore)` and pray?

Or wait for wiser heads to advise.

LouisCAD:
The error is weird. Let's see if the explicit dependency on a newer kotlinx.coroutines version fixes it.

PR #505, opened 2023-03-17
Replace usage of TestDispatcher with StandardTestDispatcher in paging-compose

veyndan:
In Multiplatform Paging, the classes `DirectDispatcher` and `TestDispatcher` in `testutils-ktx` had to be multiplatformized to allow dependent tests in `paging-compose`, `paging-common`, and `paging-runtime` to be moved to the respective `commonTest` folders.

I'm attempting to replace all usages of `DirectDispatcher` and `TestDispatcher` that Multiplatform Paging depends on with those found in `kotlinx.coroutines.test`, so hopefully the `testutils-ktx` dependency no longer has to be multiplatformized in my fork.

I'm starting this effort by doing it with `paging-compose`.

Test: ./gradlew test connectedCheck
Bug: 270612487

PR #518, opened 2023-04-14
Kmp-ify testutils-ktx

veyndan:
Upstreaming [Multiplatform Paging's kmp-ified testutils-ktx](https://github.com/cashapp/multiplatform-paging/tree/androidx-main-3.2.0-alpha04/testutils/testutils-ktx). See https://github.com/androidx/androidx/pull/505#issuecomment-1485536890 for more context.

Bug: 270612487


gradle/libs.versions.toml
```diff
@@ -255,6 +255,7 @@ skikoLinuxArm64 = { module = "org.jetbrains.skiko:skiko-awt-runtime-linux-arm64"
 sqldelightAndroid = { module = "com.squareup.sqldelight:android-driver", version.ref = "sqldelight" }
 sqldelightCoroutinesExt = { module = "com.squareup.sqldelight:coroutines-extensions", version.ref = "sqldelight" }
 sqliteJdbc = { module = "org.xerial:sqlite-jdbc", version = "3.36.0" }
+statelyIsoCollections = { module = "co.touchlab:stately-iso-collections", version = "1.2.0" }
```
ianhanniballake: From the [comments here](https://github.com/touchlab/Stately/tree/main/deprecated#stately-iso-collections), it seems like this library is completely deprecated and seems to say that the `ArrayDeque` that is now part of the Kotlin Standard Library should be enough. Do all the tests still pass if we swap to the stdlib version?

veyndan: >  […] it seems like this library is completely deprecated […]

The library isn't deprecated, but the module is. Unfortunately, the [proposed replacement (`stately-concurrent-collections`)](https://github.com/touchlab/Stately#stately-concurrent-collections) for this module [isn't stable yet](https://mvnrepository.com/artifact/co.touchlab/stately-concurrent-collections).

> […] and seems to say that the `ArrayDeque` that is now part of the Kotlin Standard Library should be enough.

A KMP version of `java.util.ArrayDeque` [was added in Kotlin 1.3.70](https://blog.jetbrains.com/kotlin/2020/03/kotlin-1-3-70-released/#Double-endedqueueimplementation:ArrayDeque). Like `java.util.ArrayDeque`, `kotlin.collections.ArrayDeque` isn't thread-safe.

> Do all the tests still pass if we swap to the stdlib version?

From the previous point, `ArrayQueue` isn't thread-safe. Swapping out `IsoArrayDeque` with `ArrayDeque` _could_ make all the tests pass, but those tests would presumably be a bit flaky.

claraf3:
Hi @veyndan thank you for your patience. After we merged your other PRs removing most usages of TestDispatcher, the remaining tests still using it appears to be single threaded access to the queue. I think its safe to at this point to switch to `ArrayDequeue`. I think this might be a better route for now so we can prevent the negatives of introducing a deprecated 3rd party module.

veyndan:
@claraf3 Sounds good to me! I've just updated the PR to use `ArrayDeque`, but I'm getting a CI error. It seems unrelated to my changes though.

claraf3:
Awesome, thank you Veyndan! Yeah it looks to be our issue. Don't worry about it :)

claraf3 (approved):

ianhanniballake (approved):

testutils/testutils-ktx/src/jvmMain/kotlin/androidx/testutils/VerifyWithPolling.kt
ianhanniballake, line 1: We're getting a lint failure on this file:
```
$SUPPORT/testutils/testutils-ktx/src/jvmMain/kotlin/androidx/testutils/VerifyWithPolling.kt:29: Error: Uses Thread.sleep() [BanThreadSleep from androidx.build]
Thread.sleep(periodMs)
~~~~~

Explanation for issues of type "BanThreadSleep":
Use of Thread.sleep() is not allowed, please use a callback or another way
to make more reliable code. See more details at
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:docs/testability.md#calling-threadsleep-as-a-synchronization-barrier
```

Do you mind seeing if you can suppress the `BanThreadSleep` check here? Ideally you should be able to reproduce this with by running `:internal-testutils-ktx:lint`

veyndan: I tried recreating this issue locally, but it always passes. I'm running the following command within the `paging` directory:

```
./gradlew :internal-testutils-ktx:lint --rerun-tasks --no-configuration-cache   
```

claraf3: I was able to repro the error on my end. I added the suppression and its running through checks. Should be good now, thanks for trying!

PR #519, opened 2023-04-14
Add `ComparableSubject#isAtLeast` to Kruth

veyndan:
To migrate the tests in `paging-common` from Truth to Kruth, `isAtLeast` needs to be added to Kruth:

https://github.com/androidx/androidx/blob/72a6b2eeaa0be3fc4a8ff989f72f771d7f109c2e/paging/paging-common/src/test/kotlin/androidx/paging/FlowExtTest.kt#L289-L290

Bug: 270612487

### DIFF
--- a/testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/ComparableSubject.kt
+++ b/testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/ComparableSubject.kt
@@ -37,4 +37,17 @@ open class ComparableSubject<T : Comparable<T>> constructor(actual: T?) : Subjec
             fail("Expected to be less than $other, but was $actual")
         }
     }
+
+    /**
+     * Checks that the subject is greater than or equal to [other].
+     *
+     * @throws NullPointerException if [actual] or [other] is `null`.
+     */
+    fun isAtLeast(other: T?) {
+        if (actual == null || other == null) {
+            throw NullPointerException("Expected to be at least $other, but was $actual")
+        } else if (actual < other) {
+            fail("Expected to be at least $other, but was $actual")
+        }
+    }
 }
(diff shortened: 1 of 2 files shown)


testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/ComparableSubject.kt
```diff
@@ -37,4 +37,17 @@ open class ComparableSubject<T : Comparable<T>> constructor(actual: T?) : Subjec
...
         }
     }
+
+    /**
+     * Checks that the subject is greater than or equal to [other].
+     *
+     * @throws NullPointerException if [actual] or [other] is `null`.
+     */
+    fun isAtLeast(other: T?) {
+        if (actual == null || other == null) {
```
arkivanov: Let's use the existing [requireNonNull](https://github.com/androidx/androidx/blob/59a572d744be4366849cb0164bc834c2b44e88ea/testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/Utils.kt#L27) function?

veyndan: Can do! The implementation was mostly a copy-paste of the `isLessThan` function above. It could perhaps be simplified there too (in a separate PR).

veyndan: @arkivanov Replacing this line with `requireNotNull` would mean an `IllegalArgumentException` would be thrown instead of a `NullPointerException`.  This diverges from Truth, as their `isAtLeast` function [also throws a `NullPointerException`](https://github.com/google/truth/blob/19dcd43928041722b8f1bd79bcd4da4faea89426/core/src/test/java/com/google/common/truth/ComparableSubjectTest.java#L61-L65). Was the original suggestion to also change the throwable type?

arkivanov: I meant our internal function, see https://github.com/androidx/androidx/blob/59a572d744be4366849cb0164bc834c2b44e88ea/testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/Utils.kt#L27

veyndan: Oops, my bad! You had linked it in the first message — I completely overlooked it.

Changed in https://github.com/androidx/androidx/pull/519/commits/20d36a60ff156846bd7271377b03cab5328940da.

arkivanov (approved):

PR #522, opened 2023-04-19
Use atomicfu in paging-common

veyndan:
Bug: 270612487

### DIFF
--- a/paging/paging-common/build.gradle
+++ b/paging/paging-common/build.gradle
@@ -36,6 +36,7 @@ dependencies {
     api("androidx.arch.core:core-common:2.1.0")
     api(libs.kotlinStdlib)
     api(libs.kotlinCoroutinesCore)
+    implementation(libs.atomicFu)
 
     testImplementation(libs.junit)
     testImplementation(libs.mockitoCore4)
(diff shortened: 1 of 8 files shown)


claraf3 (commented):
The [atomicfu library](https://github.com/Kotlin/kotlinx-atomicfu) seems to be in Beta. @ianhanniballake Does this conflict with our general policy to use stable libraries?

dlam:
We can't use atomicfu in androidx just yet. We've had discussions for this before with compose and datastore. We have a very light dependency on it in nativeMain but it's not stable enough for jvm since we want to bring that stable in a reasonable timeframe.

veyndan:
Makes sense. I decided to use atomicfu here purely because I saw it was already in your `libs.versions.toml` file, so I assumed that you're good with using it.

Multiplatform Paging currently doesn't depend on atomicfu, but instead depends on [Stately Concurrency](https://github.com/touchlab/Stately). If you look at the source code, on the JVM, it just [directly typealiases to `java.util.concurrent.atomic.*`](https://github.com/touchlab/Stately/blob/main/stately-concurrency/src/jvmMain/kotlin/co/touchlab/stately/concurrency/AtomicInt.kt).

Would you be okay with switching over to Stately Concurrency for this PR instead (which _is_ stable)?

veyndan:
FYI I've already proposed the usage of Stately in https://github.com/androidx/androidx/pull/518

claraf3:
Thanks again for your patience on this PR. Im a bit swamped right now but I'll get to this PR as soon as I can. I need to explore other options / write our own locks(?) as we also want to avoid using Stately for this as well.

claraf3:
Hi @veyndan just wanted to update you -- we want to hold off on adding new dependencies until we cut Paging 3.2, so this PR will be part of Paging 3.3. This should happen within the month.

claraf3:
Ended up using stately for Atomics in #577, closing this one as duplicate.

PR #523, opened 2023-04-19
Remove most usages of TestDispatcher in paging-common tests

veyndan:
Removes _all_ usages of `TestDispatcher` in the Paging 3 parts of `paging-common` tests, and _some_ usages of `TestDispatcher` in the Paging 2 parts of `paging-common` tests (`PagedListTest.kt` and `ContiguousPagedListTest.kt` are left).

Bug: 275416222

### DIFF
--- a/paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
+++ b/paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
@@ -22,7 +22,6 @@ import androidx.paging.LoadType.PREPEND
 import androidx.paging.PageEvent.Drop
 import androidx.paging.PagingSource.LoadResult
 import androidx.testutils.MainDispatcherRule
-import androidx.testutils.TestDispatcher
 import com.google.common.truth.Truth.assertThat
 import kotlin.coroutines.ContinuationInterceptor
 import kotlin.test.assertEquals
@@ -47,6 +46,8 @@ import kotlinx.coroutines.flow.flowOf
 import kotlinx.coroutines.flow.map
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.suspendCancellableCoroutine
+import kotlinx.coroutines.test.StandardTestDispatcher
+import kotlinx.coroutines.test.TestDispatcher
 import kotlinx.coroutines.test.TestScope
 import kotlinx.coroutines.test.UnconfinedTestDispatcher
 import kotlinx.coroutines.test.advanceTimeBy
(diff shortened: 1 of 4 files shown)


paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -1572,9 +1573,9 @@ class PagingDataDifferTest(
         differ[11]
         pagingSources[0].nextLoadResult = LoadResult.Invalid()
 
-        // using poll().run() instead of executeAll, otherwise this invalid APPEND + subsequent
+        // using advanceTimeBy instead of advanceUntilIdle, otherwise this invalid APPEND + subsequent
         // REFRESH will auto run consecutively and we won't be able to assert them incrementally
-        loadDispatcher.queue.poll()?.run()
+        loadDispatcher.scheduler.advanceTimeBy(1001)
```
veyndan, line 1578: The default `loadDelay` of `TestPagingSource` is 1000ms. We want to advance the first one only, so any value between `[1001, 1999]` would be valid.

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -2047,7 +2048,6 @@ class PagingDataDifferTest(
                 initialKey = initialKey,
                 pagingSourceFactory = {
                     TestPagingSource(
-                        loadDelay = 0,
```
claraf3, line 2050: If our tests are relying on a specific loadDelay, might be more reliable to declare a specific loadDelay so its not flaky to default impl changes. Perhaps set delay to 1000 here as well?

claraf3:
The presubmit fails on paging-compose is our issue (as with your other PRs). Don't worry about it.

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
veyndan: Makes sense. Updated PR to explicitly set `loadDelay = 1000`.

claraf3 (approved):

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
claraf3: Got it, thanks!

ianhanniballake (approved):

PR #529, opened 2023-05-01
Turn off GH CI for compose-runtime

dlam:
Seems to be failing due to a bug with version constraints that causes it to fail to resolve androidx.lifecycle:lifecycle-viewmodel-savedstate. Disabling for now since the proper fix seems like it will take some time.

Test: GH workflows
Change-Id: I9d6e1c7d86a4d389d52eefa24ec1d2d45b5ec3b6

### DIFF
--- a/.github/ci-control/ci-config.json
+++ b/.github/ci-control/ci-config.json
@@ -7,7 +7,6 @@
     ],
     "compose" : {
         "include" : [
-            "compose-runtime"
         ],
         "default": false
     },


collection/collection/src/jvmMain/kotlin/androidx/collection/Sample.kt
```diff
@@ -0,0 +1,25 @@
+/*
+ * Copyright 2023 The Android Open Source Project
```
yigit: ?

yigit (changes requested):

collection/collection/src/jvmMain/kotlin/androidx/collection/Sample.kt
dlam: Oops :|

dlam:
Should be fixed now

yigit (approved):

PR #534, opened 2023-05-04
Bump build tools in playground to latest public release

dlam:
This fixes AIDL failure in core playground

Test: ./gradlew :core:core:checkReleaseAidlApiRelease
Change-Id: I38614339310c8031995f2e412ce6084f43a0614b

### DIFF
--- a/.github/actions/build-single-project/action.yml
+++ b/.github/actions/build-single-project/action.yml
@@ -38,6 +38,9 @@ runs:
     - name: "Install Cmake"
       shell: bash
       run: echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "cmake;3.22.1"
+    - name: "Install Android SDK Build-Tools"
+      shell: bash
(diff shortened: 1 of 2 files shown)


.github/actions/build-single-project/action.yml
```diff
@@ -38,6 +38,9 @@ runs:
     - name: "Install Cmake"
       shell: bash
       run: echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "cmake;3.22.1"
+    - name: "Install Android SDK Build-Tools"
+      shell: bash
+      run: echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "build-tools;34.0.0-rc3"
```
yigit, line 43: why do we need this if there is already a constant for it?

dlam: What do you mean by a constant? This is to install in the workflow runner.

yigit (approved):

PR #540, opened 2023-05-09
Add MapSubject#isEmpty to Kruth

veyndan:
I want to convert the `paging-common` tests over to Truth. `PageFetcherSnapshotTest.kt` [requires the assertion function `MapSubject#containsExactlyEntriesIn`](https://github.com/androidx/androidx/blob/0d69980384d87e234c8e61d7164143d0a7c56309/paging/paging-common/src/test/kotlin/androidx/paging/PageFetcherSnapshotTest.kt#L1656), which doesn't yet exist in Kruth. The implementation of `MapSubject#containsExactlyEntriesIn` [depends on the existence of `MapSubject#isEmpty`](https://github.com/google/truth/blob/c2e624138fcd7e2979ecd5abcb099d345629ae5e/core/src/main/java/com/google/common/truth/MapSubject.java#L234).

As a first step, I've added support for `MapSubject#isEmpty`, as it is useful by itself.

Bug: 270612487

### DIFF
--- a/testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/MapSubject.kt
+++ b/testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/MapSubject.kt
@@ -25,7 +25,7 @@ class MapSubject<K, V>(actual: Map<K, V>?) : Subject<Map<K, V>>(actual) {
         requireNonNull(actual) { "Expected to be empty, but was null" }
 
         if (actual.isNotEmpty()) {
-            fail("Expected to be empty")
+            fail("Expected to be empty, but was $actual")
         }
     }
 }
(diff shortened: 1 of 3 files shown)


testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/MapSubject.kt
```diff
@@ -0,0 +1,31 @@
...
+import kotlin.test.fail
+
+class MapSubject<K, V>(actual: Map<K, V>?) : Subject<Map<K, V>>(actual) {
+
+    /** Fails if the map is not empty. */
+    fun isEmpty() {
+        requireNonNull(actual) { "Expected to be empty, but was null" }
+
+        if (actual.isNotEmpty()) {
+            fail("Expected to be empty")
```
dlam: Can you include the actual value in here, we don't yet match the error messages from Truth, but one nice this about Truth is that it includes the actual value.

suggest: `Expected to be empty, but was $actual`

veyndan: Makes sense! Just updated the PR

dlam (approved):

PR #546, opened 2023-05-12
Remove double commenting of license header

veyndan:
Major nitpick. Upside with this change is that Android Studio can now collapse the license header automatically.


dlam:
Thanks for fixing this, it was definitely commented out accidentally.

dlam (approved):

PR #547, opened 2023-05-12
Add assertWithMessage(String) to Kruth

veyndan:
I know in the Truth implementation that they wrap the failure message in a `FailureMetadata`. I was unsure whether to add that, as I think it's unclear whether Kruth would currently benefit from such a construct.

I wanted to have a `fail` function in `Subject` with a near identical signature to `kotlin.test.fail`, to make it less likely that one would use the `kotlin.test.fail` function directly.

Bug: 270612487

### DIFF
--- a/testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/BooleanSubject.kt
+++ b/testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/BooleanSubject.kt
@@ -21,7 +21,10 @@ import kotlin.test.assertTrue
 /**
  * Propositions for boolean subjects.
  */
-class BooleanSubject(actual: Boolean?) : Subject<Boolean>(actual) {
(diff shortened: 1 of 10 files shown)


dlam:
Hey - we're working on this change here:

https://android-review.googlesource.com/c/platform/frameworks/support/+/2557535

So you'll probably want to wait for that instead!

veyndan:
Oh sweet! I don't have access to that link unfortunately :(

veyndan:
Btw, `assertWithMessage` and `MapSubject#containsExactlyEntriesIn` are the only methods that need to be implemented to be able to migrate the `paging-common` tests over to Kruth!

dlam:
Feel free to add `MapSubject#containsExactlyEntriesIn` in the meantime. I also de-corped the link :) Of course AOSP is open source but we have weird url mapping magic that makes it easy to give the wrong url.

PR #549, opened 2023-05-17
Add MapSubject#containsExactlyEntriesIn to Kruth

veyndan:
This is the last piece to migrate the `paging-common` tests over to Kruth!

Bug: 270612487

### DIFF
--- a/testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/MapSubject.kt
+++ b/testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/MapSubject.kt
@@ -46,6 +46,29 @@ class MapSubject<K, V> internal constructor(actual: Map<K, V>?) : Subject<Map<K,
             asserter.fail("Expected $expectedMap, but was $actual")
         }
 
-        return NoopOrdered
+        return MapInOrder(expectedMap)
+    }
+
+    private inner class MapInOrder(
+        private val expectedMap: Map<*, *>,
+    ) : Ordered {
+
+        /**
+         * Checks whether the common elements between actual and expected are in the same order.
+         *
+         * This doesn't check whether the keys have the same values or whether all the required keys
+         * are actually present. That was supposed to be done before the "in order" part.
+         */
+        override fun inOrder() {
+            // We're using the fact that Iterable#intersect keeps the order of the first set.
+            checkNotNull(actual)
+            val expectedKeyOrder = (expectedMap.keys intersect actual.keys).toList()
+            val actualKeyOrder = (actual.keys intersect expectedMap.keys).toList()
+            if (actualKeyOrder != expectedKeyOrder) {
+                asserter.fail(
+                    "Entries match, but order was wrong. Expected $expectedMap, but was $actual",
+                )
+            }
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/MapSubject.kt
```diff
@@ -35,4 +35,17 @@ class MapSubject<K, V> internal constructor(actual: Map<K, V>?) : Subject<Map<K,
             asserter.fail("Expected to contain $key, but was ${actual.keys}")
         }
     }
+
+    /** Fails if the map does not contain exactly the given set of entries in the given map. */
```
dlam, line 39:
```suggestion
    /**
     * Fails if the map does not contain exactly the keys in the given map, mapping to values that
     * correspond to the values of the given map.
     */
```

Can we maintain the existing docs?

testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/MapSubject.kt
```diff
@@ -35,4 +35,17 @@ class MapSubject<K, V> internal constructor(actual: Map<K, V>?) : Subject<Map<K,
...
+    fun containsExactlyEntriesIn(expectedMap: Map<K, V>): Ordered {
+        requireNonNull(actual) { "Expected $expectedMap, but was null" }
+
+        if (expectedMap.isEmpty()) {
+            isEmpty()
+        } else if ((expectedMap.entries - actual.entries).isNotEmpty()) {
```
dlam, line 45: This doesn't return `MapInOrder` and instead returns `NoopOrdered`, even if the entries are out of order.

For example this test should fail:
```
    @Test
    fun orderedMap() {
        val map = LinkedHashMap<Int, Int>()
        repeat(5) { map[it] = it }

        val unsortedMap = HashMap<Int, Int>()
        repeat(5) { unsortedMap[4 - it] = 4 - it }

        val reverseMap = LinkedHashMap<Int, Int>()
        repeat(5) { reverseMap[4 - it] = 4 - it }

        assertThat(map).containsExactlyEntriesIn(unsortedMap).inOrder()
        assertThat(map).containsExactlyEntriesIn(reverseMap).inOrder()
    }
```

dlam (changes requested):

testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/MapSubject.kt
veyndan: I think that's the docs for the other `containsExactlyEntriesIn`. I took these docs from [this function definition](https://github.com/google/truth/blob/7d4a851d1b44ce054c0984ae96c97fb6cfcbd1f0/core/src/main/java/com/google/common/truth/MapSubject.java#L227-L244).

dlam: Ah yes you're right - my mistake!

testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/MapSubject.kt
veyndan: Oof, I'd just copied all the `containsExactlyEntriesIn` tests from Truth and assumed it'd cover all cases. Just updated the PR to throw when the elements are out of order.

veyndan:
@dlam Just wondering if you've had time to review the PR since the changes!

testutils/testutils-kmp/src/commonMain/kotlin/androidx/kruth/MapSubject.kt
dlam: I believe they do have some, but it also depends on a different API that we don't have yet, which is probably why it got missed.

In the meantime, can we at least add the test I showed which fails without the `MapInOrder` change? Otherwise, everything else looks good to me !

dlam (approved):
Starting to need this in other places too, so I'll approve for now and add the test separately.

PR #550, opened 2023-05-22
Fix SBOM Generation in Github and add workaround for the Gradle dependency resolution bug

yigit:
Add workarounds for gradle resolution bug:
https://github.com/gradle/gradle/issues/8489

Also fixed a problem with sboms in playground since we don't have prebuilt repos there.

Looks like updating the activity and lifecycle dependencies does work around the issue.
It is not great but the only solution i could find.

Bug: n/a
Test: CI

### DIFF
--- a/buildSrc/private/src/main/kotlin/androidx/build/sbom/Sbom.kt
+++ b/buildSrc/private/src/main/kotlin/androidx/build/sbom/Sbom.kt
@@ -18,11 +18,11 @@ package androidx.build.sbom
 
 import androidx.build.BundleInsideHelper
 import androidx.build.GMavenZipTask
+import androidx.build.ProjectLayoutType
 import androidx.build.addToBuildOnServer
 import androidx.build.getPrebuiltsRoot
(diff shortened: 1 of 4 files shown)


yigit:
FYI i'll have to get reviews in AOSP for this since it does touch multiple projects.

buildSrc/lint.xml
```diff
@@ -43,5 +43,7 @@
     <!-- Disable all lint checks on transformed classes by default. -->
     <issue id="all">
         <ignore path="**/.transforms/**" />
+        <!-- playground builds have dependency files in "transformed" instead of ".transforms" -->
```
yigit: note that this is actually merged here:
https://android-review.git.corp.google.com/c/platform/frameworks/support/+/2599045

but kept it in this CL so i PR here also passes (well, except for navigation which is referring an unreleased artifact)

dlam (approved):

yigit:
rebased

PR #558, opened 2023-06-05
Remove unused `StoreRealActor.kt`

veyndan:


claraf3 (approved):
For posteriority, was used by `paging/common/src/main/kotlin/androidx/paging/multicast/ChannelManager.kt` which has since then been deleted.

PR #560, opened 2023-06-06
Use existing TestScope instead of creating a new one

veyndan:
I'm trying to see if I can elegantly remove the dependence of `PagingDataDifferTest.kt` on `MainDispatcherRule`, without just inlining the class. Keeping track of all the `CoroutineScope`s for each test is difficult though, so I'm trying to slightly reduce it here (i.e., use the `TestScope` that the `SimpleDiffer` is invoked from, instead of creating a new one).


veyndan:
I managed to remove the usage of `MainDispatcherRule` at https://github.com/androidx/androidx/pull/562. This change is no longer required. Personally, I find this change easier to read than what was before, but if you disagree, feel free to just close this PR.

claraf3 (approved):

claraf3:
Thanks for the clean up!

PR #562, opened 2023-06-06
Remove usage of MainDispatcherRule in PagingDataDifferTest

veyndan:
Opted to just pass in an `EmptyCoroutineContext` to `PagingDataDiffer` instead of inlining `MainDispathcherRule`, as it's clearer to follow. The removal of `loadDispatcher` was necessary for this to pass. I didn't trace why this had to be removed, but imo it makes the tests easier to read.

Bug: 270612487

### DIFF
--- a/paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
+++ b/paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
@@ -22,15 +22,13 @@ import androidx.paging.LoadState.NotLoading
 import androidx.paging.LoadType.PREPEND
 import androidx.paging.PageEvent.Drop
 import androidx.paging.PagingSource.LoadResult
-import androidx.testutils.MainDispatcherRule
-import kotlin.coroutines.ContinuationInterceptor
+import kotlin.coroutines.EmptyCoroutineContext
 import kotlin.test.Test
 import kotlin.test.assertEquals
 import kotlin.test.assertFailsWith
 import kotlin.test.assertFalse
 import kotlin.test.assertNull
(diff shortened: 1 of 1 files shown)


claraf3:
`loadDispatcher` wasn't working because without `Dispatchers.setMain(TestDispatcher)`, `loadDispatcher` would be using a different scheduler than `testScope`. It would pass if we initiated it with `StandardTestDispatcher(testScope.testScheduler)`. But I'm fine if we remove loadDispatcher too!

paging/paging-common/src/test/kotlin/androidx/paging/PagingDataDifferTest.kt
```diff
@@ -2185,28 +2175,25 @@ class PagingDataDifferTest(
...
         ) -> Unit
     ) = testScope.runTest {
+        val pagingSources = mutableListOf<TestPagingSource>()
+        val pager = Pager(
+            config = PagingConfig(pageSize = 3, enablePlaceholders = false),
+            initialKey = initialKey,
+            pagingSourceFactory = {
+                TestPagingSource(
+                    loadDelay = 1000,
+                    loadContext = coroutineContext,
```
claraf3: I think we no longer need to set `loadContext` here

veyndan: Ooo nice, just removed it in the latest commit!

claraf3 (approved):

claraf3:
Hi @veyndan there are merge conflicts on this PR, probably because I had just merged in the other PRs. Can you please try rebasing this?

veyndan:
Rebased!

PR #566, opened 2023-06-09
Update prebuilts and metadata versions

yigit:
Also updated the github version update workflow to also update metalava.

Bug: n/a
Test: :datastore:datastore:generateApi

### DIFF
--- a/.github/workflows/update_prebuilts.yml
+++ b/.github/workflows/update_prebuilts.yml
@@ -23,7 +23,9 @@ jobs:
         uses: actions/checkout@v2
       - name: "Run update prebuilt snapshot ids script"
         shell: bash
-        run: development/update_playground.sh a
+        # this does not update metalava because it needs to match the
+        # metalava buildId used by androidx
+        run: development/update_playground.sh
       - name: "Check if anything changed"
         shell: bash
         id: "check-for-changes"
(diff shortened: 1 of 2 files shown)


.github/workflows/update_prebuilts.yml
```diff
@@ -23,7 +23,7 @@ jobs:
         uses: actions/checkout@v2
       - name: "Run update prebuilt snapshot ids script"
         shell: bash
-        run: development/update_playground.sh
+        run: development/update_playground.sh a
```
dlam: typo?

yigit: https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:development/update_playground.sh;l=25

dlam: Oh I forgot I wrote this - we definitely don't want to update metalava to ToT all the time because we need to match whatever version we're using in AOSP.

dlam: The correct way to fix this is actually to add this into importMaven so that it automatically gets bumped when we do it in AOSP like here: [aosp/2619889](http://aosp/2619889)

If we bump to ToT, we will get ahead of AOSP and have API file conflicts

yigit: ok that is reasonable, i'll give it a try and remove the `a` change from here.

dlam (approved):

PR #567, opened 2023-06-12
Gradle Wrapper Validation Action

rnakade:
## Proposed Changes

This action validates the checksums of [Gradle Wrapper](https://docs.gradle.org/current/userguide/gradle_wrapper.html) JAR files present in the source tree and fails if unknown Gradle Wrapper JAR files are found.
See: https://github.com/gradle/wrapper-validation-action

## Testing

Test: N/A

### DIFF
--- a/.github/workflows/validate_gradle_wrapper.yml
+++ b/.github/workflows/validate_gradle_wrapper.yml
@@ -1,0 +1,10 @@
+name: "Validate Gradle Wrapper"
+on: [push, pull_request]
+
+jobs:
+  validation:
+    name: "Validation"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+      - uses: gradle/wrapper-validation-action@v1
--- a/.github/workflows/validate_gradle_wrapper.yml
+++ b/.github/workflows/validate_gradle_wrapper.yml
@@ -1,5 +1,11 @@
 name: "Validate Gradle Wrapper"
-on: [push, pull_request]
+
+on: 
+  pull_request:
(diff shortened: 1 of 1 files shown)


yigit (approved):

JakeWharton:
I would suggest only running this if the affected files change (`gradlew`, `gradlew.bat`, and `gradle/wrapper/` dir). Otherwise, this thing will fail somewhere between 1-4% of all builds.

Like this: https://github.com/cashapp/redwood/blob/88e943159bc3b9c98af128465a04ef2b838a4157/.github/workflows/gradle-wrapper.yaml

liutikas:
> I would suggest only running this if the affected files change (`gradlew`, `gradlew.bat`, and `gradle/wrapper/` dir). Otherwise, this thing will fail somewhere between 1-4% of all builds.
> 
> Like this: https://github.com/cashapp/redwood/blob/88e943159bc3b9c98af128465a04ef2b838a4157/.github/workflows/gradle-wrapper.yaml

Thats a good suggestion. @rnakade can you please limit it on paths changed?

rnakade:
> > I would suggest only running this if the affected files change (`gradlew`, `gradlew.bat`, and `gradle/wrapper/` dir). Otherwise, this thing will fail somewhere between 1-4% of all builds.
> > Like this: https://github.com/cashapp/redwood/blob/88e943159bc3b9c98af128465a04ef2b838a4157/.github/workflows/gradle-wrapper.yaml
> 
> Thats a good suggestion. @rnakade can you please limit it on paths changed?

Done!

PR #569, opened 2023-06-16
Fixes for GitHub playground

dlam:
- Activity needs to depends on benchmark via projectOrArtifact
- Fragment needs to depend on activity via projectOrArtifact
- Collection needs to depend on annotation via projectOrArtifact
 - Bump snapshot id to pull in correct version of compose compiler
   
Test: GH CI actions
Change-Id: Id880f4b55a962fb2b568b0cd3cec9353082a5091

### DIFF
--- a/activity/integration-tests/baselineprofile/build.gradle
+++ b/activity/integration-tests/baselineprofile/build.gradle
@@ -40,8 +40,8 @@ android {
 }
 
 dependencies {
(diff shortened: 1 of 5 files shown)


dlam:
Note appcompat is still failing for prerelease api check that is likely failing due to switching to snapshot

claraf3 (approved):

claraf3 (approved):

PR #574, opened 2023-06-23
Adds retryable compose test rule

dabrosch:
## Proposed Changes

- This solves the "Only a single call to `runTest` can be performed during one test." error when you are trying to retry a test that has uses a ComposeRule.
- Achieved by creating a RetryableComposeRule class.
- This is needed because the environment that gets created from a ComposeRule is not supposed to be reused in any other test runs per discussion here: https://github.com/Kotlin/kotlinx.coroutines/issues/3718#issuecomment-1512832250 so this new rule creates a new environment after each test has ran.
- This should help keep people from following the incorrect workaround such as here: https://issuetracker.google.com/issues/235383900 where they re-use the ComposeRule for subsequent retries by ordering their rules specifically to do so.
- This rule should be ordered after any retry rule.

## Testing

Test: I setup an intentionally failing compose jUnit test with a retry rule outside of my RetryableComposeRule and observed that upon the retry there was NO error of "Only a single call to `runTest` can be performed during one test." 

## Issues Fixed

Fixes: https://issuetracker.google.com/issues/235383900 and https://github.com/Kotlin/kotlinx.coroutines/issues/3718#issuecomment-1512832250

### DIFF
--- a/compose/ui/ui-test-junit4/src/androidMain/kotlin/androidx/compose/ui/test/junit4/AndroidComposeTestRule.android.kt
+++ b/compose/ui/ui-test-junit4/src/androidMain/kotlin/androidx/compose/ui/test/junit4/AndroidComposeTestRule.android.kt
@@ -340,6 +340,68 @@ class AndroidComposeTestRule<R : TestRule, A : ComponentActivity> private constr
     override fun setContent(composable: @Composable () -> Unit) = composeTest.setContent(composable)
 }
 
+/**
+ * An implementation of [ComposeTestRule] that will correctly reset itself in-between test retries.
+ *
+ * NOTE: In order to function properly this rule should be wrapped by your retry rule.
+ *
+ * This is necessary because there is currently no ability to reset the
+ * [AndroidComposeUiTestEnvironment] in the standard [AndroidComposeTestRule].
+ *
+ * @param ruleProvider Function to create the underlying ComposeTestRule rule, defaults to [createEmptyComposeRule]
+ */
+class RetryableComposeTestRule constructor(
+    private val ruleProvider: () -> ComposeTestRule = ::createEmptyComposeRule
+) :
+    ComposeTestRule {
+    private var rule: ComposeTestRule = ruleProvider.invoke()
+
+    override val density: Density
+        get() = rule.density
+    override val mainClock: MainTestClock
+        get() = rule.mainClock
+
+    override fun apply(base: Statement, description: Description): Statement {
+        return object : Statement() {
+            override fun evaluate() {
+                try {
+                    rule.apply(base, description).evaluate()
+                } finally {
+                    rule = ruleProvider.invoke()
+                }
+            }
+        }
+    }
+
+    override suspend fun awaitIdle() = rule.awaitIdle()
+
+    override fun onAllNodes(
+        matcher: SemanticsMatcher,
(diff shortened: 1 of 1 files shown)


AShunevych:
Can you explain how exactly your rule should be implemented ? For example, we have rule chain :

```
//
val basicRule = createAndroidComposeRule<MyActivity>()  //  basic rule 
val retryRule = RetryableComposeTestRule  { basicRule } // retry rule as you propose 
val flakyRule = FlakytestRule().allowFlakyAttemptsByDefault(2) // simple test rule from Barista lib 

// So we have flaky rule as outerRule and all rules goes around
@get:Rule
val rulechain:RuleChain = run {
 RuleChain.outerRule(flakyRule).around(basicRule).around(retryRule) 
}
```
    
Test will restart, but again IllegalStateException("Only a single call to runTest can be performed during one test.") will be thrown exception will be thrown.
Thank you.

dabrosch:
> Can you explain how exactly your rule should be implemented ? For example, we have rule chain :
> 
> ```
> //
> val basicRule = createAndroidComposeRule<MyActivity>()  //  basic rule 
> val retryRule = RetryableComposeTestRule  { basicRule } // retry rule as you propose 
> val flakyRule = FlakytestRule().allowFlakyAttemptsByDefault(2) // simple test rule from Barista lib 
> 
> // So we have flaky rule as outerRule and all rules goes around
> @get:Rule
> val rulechain:RuleChain = run {
>  RuleChain.outerRule(flakyRule).around(basicRule).around(retryRule) 
> }
> ```
> 
> Test will restart, but again IllegalStateException("Only a single call to runTest can be performed during one test.") will be thrown exception will be thrown. Thank you.

The problem is that you have two composeRules in the RuleChain, you have the basicRule AND the retryRule, you should only have the retryRule. 

So I suggest removing the entire `val basicRule` line and instead change the next to this:
```
val retryRule = RetryableComposeTestRule  { createAndroidComposeRule<MyActivity>() } // retry rule as you propose 
```

AShunevych:
> So I suggest removing the entire `val basicRule` line and instead change the next to this:
> 
> ```
> val retryRule = RetryableComposeTestRule  { createAndroidComposeRule<MyActivity>() } // retry rule as you propose 
> ```
Sorry for not responding that long. Thanks!.
I wonder: 
retryRule will inherit from `ComposeTestRule`, not from `AndroidComposeTestRule` is there any way to make it use [AndroidComposeTestRule ](https://developer.android.com/reference/kotlin/androidx/compose/ui/test/junit4/AndroidComposeTestRule)instead ?
I am asking about this because, end-to-end tests may often require interaction with activity.

dabrosch:
Sorry for my late reply as well, hah!

Well, since you see that I am overriding all of the regular ComposeTestRule
public elements, you would have to change the output to be
AndroidComposeTestRule and override all of the additional public elements
to expose what you are looking for.

What I would probably do is inherit from RetryableComposeRule and implement
a new class RetryableAndroidComposeRule.

But your other option is to just be in control of your own launching of
activities (by using the createEmptyComposeRule), and therefore you should
already have a reference to the activity to interact with.

--Daniel

dabrosch:
The ASOP project has finally been updated with a fix for this issue; so closing this workaround. See: https://android-review.googlesource.com/c/platform/frameworks/support/+/4072862

PR #578, opened 2023-06-30
Make `testutils-paging` a KMP

veyndan:
Bug: 270612487


claraf3 (approved):

claraf3:
Are we going to move jvm-free files to common later?

veyndan:
That's the plan! We can move them over to `commonMain` when the dependent classes in `paging-common` have been moved over to `commonMain`.

veyndan:
@claraf3 Moved classes over to `commonMain`!

PR #581, opened 2023-07-04
Only reference AndroidX Annotation's `@MainThread` in the JVM

veyndan:
Bug: 270612487


paging/paging-common/build.gradle
```diff
@@ -75,7 +78,7 @@ androidXMultiplatform {
 
         if (enableNative) {
             nativeMain {
-                dependsOn(commonMain)
+                dependsOn(nonJvmMain)
```
claraf3: As discussed, let's use nativeMain for now until we add other targets.

claraf3 (approved):

PR #582, opened 2023-07-04
Make `EXCEPTION` accessible from `commonMain`

veyndan:
Bug: 270612487

### DIFF
--- a/paging/paging-common/src/commonTest/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
+++ b/paging/paging-common/src/commonTest/kotlin/androidx/paging/PageFetcherSnapshotTest.kt
@@ -17,7 +17,6 @@
 package androidx.paging
 
 import androidx.kruth.assertThat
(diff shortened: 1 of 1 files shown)


claraf3:
Ahh I did this in a merged CL. I cc'd you on it but you probably don't get notification unless I add you as Reviewer? I'll add you as reviewer moving forward.
https://android-review.git.corp.google.com/c/platform/frameworks/support/+/2647529/1/paging/paging-common/src/commonTest/kotlin/androidx/paging/PageFetcherSnapshotTest.kt

Here is the [chain of CLs](https://android-review.googlesource.com/c/platform/frameworks/support/+/2647529/1) with a few minor fixes.

veyndan:
> I'll add you as reviewer moving forward.

Sounds good!

veyndan:
Actually, I just looked at my emails and I do get emails when I'm CC'd. I must've missed that CL. I don't mind being added as a reviewer or just being CC'd

PR #586, opened 2023-07-06
Make `paging-testing` multiplatform with only a JVM implementation

veyndan:
Test: ./gradlew paging:paging-testing:allTests


paging/paging-testing/build.gradle
```diff
@@ -15,24 +15,39 @@
  */
 
 import androidx.build.LibraryType
+import androidx.build.PlatformIdentifier
 import androidx.build.Publish
 
 plugins {
     id("AndroidXPlugin")
-    id("com.android.library")
-    id("org.jetbrains.kotlin.android")
```
claraf3, line 23: I think we need to continue targeting android because the first stable version is to be released with android support. @ianhanniballake What are your thoughts on this?

ianhanniballake: Given that these are testing paging-common APIs, I think it makes to match the same set of targets.

claraf3: @veyndan We are in the process of refactoring Logger in paging-common with added android target so paging-common will become an android library. Working on the CL as we speak, should be up today. I'll cc you on it.

claraf3: I've [updated paging-common](https://android-review.googlesource.com/c/platform/frameworks/support/+/2651248/4) into an android library now.

veyndan: Just updated the PR to add back the android target!

paging/paging-testing/build.gradle
```diff
@@ -15,24 +15,41 @@
...
+    defaultPlatform(PlatformIdentifier.JVM)
+
+    sourceSets {
+        jvmMain {
+            dependencies {
+                api(libs.kotlinStdlib)
+                implementation(project(":paging:paging-common"))
+            }
+        }
+        jvmTest {
```
claraf3, line 42: Does command `./gradlew paging:paging-testing:connectedCheck` still run successfully? I know we don't have android tests but in paging-common I had to [add android-specific test dependencies](https://android-review.git.corp.google.com/c/platform/frameworks/support/+/2653235/2/paging/paging-common/build.gradle#92) or else `connectCheck` task errors out. Not sure if we need it for paging-testing.

veyndan: I just ran it and it completed successfully

claraf3 (approved):

PR #587, opened 2023-07-07
Simplify Android logger

veyndan:
`paging-compose` counterpart to cd7efa354b35c7e2baa2a815d56e1ddda9c288c0.

Test: ./gradlew paging:paging-compose:test


claraf3:
I [refactored Logger](https://android-review.googlesource.com/c/platform/frameworks/support/+/2651250) into an expect/actual and simplified the android logger as you suggested here. Thanks!

PR #589, opened 2023-07-12
Remove duplicated text from compiler-metrics.md

tatocaster:
## Proposed Changes

  - Removes duplicated block of text from the markdown document

## Testing

Bug: N/A
Test: local


tatocaster:
oops mistake, incorrect commit. report and metrics are correct

PR #600, opened 2023-08-11
[Compose-Android] Modify Resources for android internal implementation

hubtwork:
_This is a internal implementation style change_

This PR tries to highlight the difference between :

`resources()` in Resources.android.kt has unusable code block for `LocalConfiguration.current`.
It seems not intuitive implementation which uses just doubling call for `LocalConfiguration.current`.
So catch the value and reuse it for intuitiveness.

Test: N/A expected function result is not different with before one.


miguelalegria (approved):

hubtwork:
@dlam @yigit why no approval?

PR #604, opened 2023-08-18
Use projectOrArtifact in fragment/fragment

dlam:
Test: cd fragment && ./gradlew bOS
Change-Id: I45b2cc1f0383bf4036bf6c7f017ca6ba4454bde3


liutikas (approved):

liutikas:
It looks like you also need to update activity-ktx reference, based on github actions results.

PR #606, opened 2023-08-29
Fix ModalBottomSheet layout issues

ln-12:
## Proposed Changes

The issue is described here: https://issuetracker.google.com/issues/292204649

To fix the layout not being updated on screen rotation, we just have to provide `LocalConfiguration.current` as key to the remember call for the window.
However, there was still a blank space on the right side when in landscape mode (navigation bar on the left). This issue is fixed by using the whole screen width. The insets are already applied to the content view.

## Testing

Test: I tested the behavior before and after my changes on two Pixel emulator, a Pixel 7 and a Samsung Galaxy S10 Plus.

## Issues Fixed

Fixes: The bug on [https://issuetracker.google.com/issues/292204649](https://issuetracker.google.com/issues/292204649) being fixed


ln-12:
Is now already fixed in 
* https://android-review.googlesource.com/c/platform/frameworks/support/+/2747800/2/compose/material3/material3/src/androidMain/kotlin/androidx/compose/material3/ModalBottomSheet.android.kt
* https://android-review.googlesource.com/c/platform/frameworks/support/+/2743558/2/compose/material3/material3/src/androidMain/kotlin/androidx/compose/material3/ModalBottomSheet.android.kt

PR #608, opened 2023-09-06
Do not check @StabilityInferred if class is not public

0xera:
## Proposed Changes

I found out that Compose Compiler add @StabilityInferred only for public classes in [ClassStabilityTransformer.kt](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/compiler/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/lower/ClassStabilityTransformer.kt;drc=4d53400eca9f3ac90c3a3f6cffcbc5bf492ec536;l=89), but that is not checked in stabilityOf and can lead to loosing stability of internal class.

For example:
SimpleClass.kt
```kotlin
internal SimpleClass
```

SimpleComposable.kt
```kotlin
@Composable
internal fun SimpleComposable(clazz: SimpleClass)
``` 

For clean build all things are good, because `canInferStability` returns `false` and stability was checked for class members.
But if we add some change in SimpleComposable.kt. For example:
```kotlin
@Suppress("So experimental")
@Composable
internal fun SimpleComposable(clazz: SimpleClass)
```

We should recompile only SimpleComposable. So SimpleClass will have `IR_EXTERNAL_DECLARATION_STUB` origin and `canInferStability` will returns `true`, but because compiler does not add  @StabilityInferred for internal class SimpleClass will be unstable for second incremental compilation even it does not have any members.

For fix i added checking of class visibility, but may be we should allow to add @StabilityInferred for internal classes?

## Testing

Test: Use simple example above. I do not know how write test for incremental compilation in Compose Compiler :)

### DIFF
--- a/compose/compiler/compiler-hosted/integration-tests/src/test/java/androidx/compose/compiler/plugins/kotlin/ClassStabilityTransformTests.kt
+++ b/compose/compiler/compiler-hosted/integration-tests/src/test/java/androidx/compose/compiler/plugins/kotlin/ClassStabilityTransformTests.kt
@@ -20,6 +20,7 @@ import androidx.compose.compiler.plugins.kotlin.analysis.stabilityOf
 import androidx.compose.compiler.plugins.kotlin.facade.SourceFile
 import org.intellij.lang.annotations.Language
 import org.jetbrains.kotlin.ir.declarations.IrClass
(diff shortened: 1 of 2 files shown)


0xera:
i have moved this little commit to #610 for fix more stability problems

PR #610, opened 2023-09-09
Fix lambda and function reference memorization and internal classes loosing stability on incremental compilation. Fix $stable field generation for classes with combined stability

0xera:
## Proposed Changes

I have started to fix bug(https://issuetracker.google.com/issues/298965625) and have founded other incorrect cases

### Lambda memoization for RuntimeStable types
Problem: On incremental compilation lambda was not remember as Stable class was converted to RuntimeStable, because `@StabilityInferred` was added.

Solution: Replaced `!knownStable()` to `knownUnstable` to avoid memoization when type is definitely unstable. This logic is similar to how Compose Compiler adds `skipToGroupEnd` for restartable functions.

### Function reference memoization for RuntimeStable types of dispatch and extension receivers
Solution: Similar to previous one. knownStable() replaced to !knownUnstable()

### Function reference memoization when extension receiver is not Stable
Problem: If function reference does not have dispatch receiver but has extension it will be remember even it is unstable. This is not correct. 

Solution: So i added a check for stability of extension receiver if there is no dispatch

### Unstable class transforms to RuntimeStable on incremental compilation
Problem: Even if class is unstable `@StabilityInferred` is added and on incremental compilation that class stability will be changed to RuntimeStable. And it breaks some things on incremental compilation. 
For example:
1. lambda is memorized (after my fixes previous problems)
2. unused skipToGroupEnd is generated for restartable functions 

Solution: I removed generation of `@StabilityInferred` if class is unstable. 

Why is it necessary to generate `@StabilityInferred` for unstable classes? Why is it not generated for internal classes? (#608)

### Internal classes loose stability on incremental compilation (#608)
Problem: I found out that Compose Compiler add @StabilityInferred only for public classes in [ClassStabilityTransformer.kt](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:compose/compiler/compiler-hosted/src/main/java/androidx/compose/compiler/plugins/kotlin/lower/ClassStabilityTransformer.kt;drc=4d53400eca9f3ac90c3a3f6cffcbc5bf492ec536;l=89), but that is not checked in stabilityOf and can lead to loosing stability of internal class.

For example:

```kotlin
// SimpleClass.kt
internal SimpleClass

// SimpleComposable.kt
@Composable
internal fun SimpleComposable(clazz: SimpleClass)
``` 

For clean build all things are good, because `canInferStability` returns `false` and stability was checked for class members.
But if we add some change in SimpleComposable.kt. For example adding some annotation:
```kotlin
@Suppress("So experimental")
@Composable
internal fun SimpleComposable(clazz: SimpleClass)
```

We should recompile only SimpleComposable. So SimpleClass will have `IR_EXTERNAL_DECLARATION_STUB` origin and `canInferStability` will returns `true`, but because compiler does not add  @StabilityInferred for internal class SimpleClass will be unstable for second incremental compilation even it does not have any members.

Solution: For fix i added checking of class visibility, but may be we should allow to add @StabilityInferred for internal classes?

### $stable is not generated correctly for CombinedStable classes
Problem: There is little problem for check that symbol in parameters. So parameters will never be added to parts
```kotlin
if (parameters.contains(stability.parameter.symbol)) {
     parameters.add(stability.parameter.symbol)
     parts.add(stability)
}
```

Solution: Just replace `parameters.contains(stability.parameter.symbol)` to `stability.parameter.symbol !in parameters`

## Testing
Test:
  1. Make two file with empty class and `@Composable` function 
  2. Add some needed code in function
  3. Compile and check generated IR
  4. Make small changes in file with function
  5. Compile and compare IR with the previous IR

Fixes: 
- https://issuetracker.google.com/issues/298965625
- and improve other cases for incremental compilation like in previous PR for internal classes https://github.com/androidx/androidx/pull/608

### DIFF
--- a/compose/compiler/compiler-hosted/integration-tests/src/test/java/androidx/compose/compiler/plugins/kotlin/ClassStabilityTransformTests.kt
+++ b/compose/compiler/compiler-hosted/integration-tests/src/test/java/androidx/compose/compiler/plugins/kotlin/ClassStabilityTransformTests.kt
@@ -20,6 +20,7 @@ import androidx.compose.compiler.plugins.kotlin.analysis.stabilityOf
 import androidx.compose.compiler.plugins.kotlin.facade.SourceFile
 import org.intellij.lang.annotations.Language
 import org.jetbrains.kotlin.ir.declarations.IrClass
+import org.jetbrains.kotlin.ir.declarations.IrDeclarationOrigin
 import org.jetbrains.kotlin.ir.declarations.IrModuleFragment
 import org.jetbrains.kotlin.ir.declarations.IrSimpleFunction
 import org.jetbrains.kotlin.ir.expressions.IrExpression
@@ -881,7 +882,6 @@ class ClassStabilityTransformTests(useFir: Boolean) : AbstractIrTransformTest(us
               }
               static val %stable: Int = 0
             }
-            @StabilityInferred(parameters = 0)
             class UnstableDelegate {
               var value: Int = 0
               fun setValue(thisObj: Any?, property: KProperty<*>, value: Int) {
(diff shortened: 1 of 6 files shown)


0xera:
Probably i should mention @lelandrichardson

0xera:
@dlam could you restart pipelines, please? i have fixed pr description. How long does pr stay at the review waiting step? 😄

PR #614, opened 2023-09-22
Fix typo in CheckResult.kt

Gresgo:
## Proposed Changes

Fix typo in the `@CheckResult` annotation example

In the example a method with a parameter `String s` and annotation was declared, but used default `String.trim()` instead

## Testing

Test: no need to test


yigit (approved):
approving to trigger the aosp clone

PR #635, opened 2023-12-16
Properly handle ClassCastException in SavedStateHandle when calling get() with the incorrect type

hoc081098:
When using SavedStateHandle, if you attempt to retrieve a value with the incorrect class type, it is casted to the correct class type and ends up throwing a ClassCastException. But inside `catch` block, we call `remove<T>(key)`, which will throw another ClassCastException, because `remove` method also tries to cast the existing value to the given type. This commit changes `T` to `Any?` to avoid ClassCastException.

## Proposed Changes

  - Change `remove<T>(key)` to `remove<Any?>(key)`.

## Testing

Test: Added `testReturnsNullWhenWrongType` to `lifecycle/lifecycle-viewmodel-savedstate/src/androidTest/java/androidx/lifecycle/viewmodel/savedstate/SavedStateHandleTest.kt`

## Issues Fixed


hoc081098:
Close this 🙏 
I see the `catch(e: ClassCastException)` is never called.
ClassCastException is thrown at the consumer side.
This is the limitation of erased generic type

PR #641, opened 2023-12-25
add missing then call in ContextReceiverTests

AntonButov:
Proposed Changes

  - now compose lambda, does not invoke "assertEquals" calls. You will be able to verify this if you change the expected value to a deliberately incorrect one and run the test.

Testing

Test: ./gradlew connectedDebugAndroidTest -Pandroid.testInstrumentationRunnerArguments.tests_regex='ContextReceiverTests*'

 Issues Fixed

It is addition for tests.


AntonButov:
I added "Test: " into description.

PR #647, opened 2024-01-12
Support building sqlite / room on Github

yigit:
This CL adds sqlite as a project to GitHub and makes it possible compile.

On AOSP, we use sqlite prebuilts to find sources. On GitHub, we'll download sqlite
amalgamation. To ensure they stay consistent, we check the version in the header
file which will also be checked on AOSP build.

Fixes: 306669673
Test: ./gradlew playgroundCIHostTests


buildSrc/private/src/main/kotlin/androidx/build/PrepareSqliteSourcesTask.kt
```diff
@@ -0,0 +1,255 @@
...
+ * [destinationDirectory].
+ * On Github, they are downloaded from SQLite servers.
+ *
+ * To ensure each version is consistent, we use the [sqliteVersion] parameter and check the SQLite
+ * source code for them.
+ */
+@DisableCachingByDefault(
+    because = "This is a simple copy/validation task that is not worth caching"
+)
+abstract class PrepareSqliteSourcesTask @Inject constructor(
```
yigit: i'm not sure if this should live in public or private buildSrc.
it is only used by the sqlite-bundled task so we could consider moving this into that build.gradle file but the logic is not straightforward, hence moved into buildSrc.

buildSrc/private/src/main/kotlin/androidx/build/KonanPrebuiltsSetup.kt
```diff
@@ -34,16 +34,21 @@ object KonanPrebuiltsSetup {
 
     /**
      * Creates a Konan distribution with the given [prebuiltsDirectory] and [konanHome].
+     *
+     * @param prebuiltsDirectory The directory where AndroidX prebuilts are present. Can be `null`
```
liutikas, line 38: make a note that null means fetching konan from network.

buildSrc/private/src/main/kotlin/androidx/build/PrepareSqliteSourcesTask.kt
liutikas: this is super sqlite specific and not even used in aosp, i'd say we move this to sqlite build.gradle

sqlite/sqlite-bundled/build.gradle
```diff
@@ -76,6 +82,14 @@ configurations {
             attribute(Usage.USAGE_ATTRIBUTE, objects.named(Usage, Usage.NATIVE_LINK))
         }
     }
+    // Configuration for producing a shareable directory that includes SQLite amalgamation sources.
+    sqliteSources {
```
liutikas, line 90: side question: what do we want to do with sources of sqlite that be bundle? should they ship in source jars?

buildSrc/private/src/main/kotlin/androidx/build/PrepareSqliteSourcesTask.kt
yigit: done, moved to a groovy file.

sqlite/sqlite-bundled/build.gradle
yigit: @danysantiago , wdyt? 
imo, might be nice for completeness but the only API seen by consumer is the sqlite driver impl created by us (they don't get direct access to sqlite though it might be possible to load it maybe).

buildSrc/private/src/main/kotlin/androidx/build/clang/KonanBuildService.kt
```diff
@@ -233,9 +236,11 @@ abstract class KonanBuildService @Inject constructor(
                 it.parameters.konanHome.set(
                     nativeCompilerDownloader.compilerDirectory
                 )
-                it.parameters.prebuilts.set(
-                    project.getKonanPrebuiltsFolder()
-                )
+                if (!ProjectLayoutType.isPlayground(project)) {
+                    it.parameters.prebuilts.set(
+                        project.getKonanPrebuiltsFolder()
```
liutikas, line 267: Can we assert it is not null here? I would not want to accidentally start downloading these in CI.

buildSrc/private/src/main/kotlin/androidx/build/clang/NativeTargetCompilation.kt
```diff
@@ -137,4 +140,29 @@ class NativeTargetCompilation internal constructor(
             }
         }
     }
+
+    /**
+     * JDK ships with JNI headers only for the current platform. As a result,we don't have access
```
liutikas: missing space after `,`

buildSrc/private/src/main/kotlin/androidx/build/clang/NativeTargetCompilation.kt
```diff
@@ -137,4 +140,29 @@ class NativeTargetCompilation internal constructor(
...
+     * to cross-platform jni headers. They are mostly the same and we don't ship cross compiled code
+     * from GitHub so it is acceptable to use local JNI headers for cross platform compilation on
+     * GitHub.
+     */
+    private fun findJniHeadersInPlayground(
+        javaHome: File
+    ): List<File> {
+        var include = File(javaHome, "include")
+        if (!include.exists()) {
```
liutikas: why is this needed? Please leave a comment

sqlite/sqlite-bundled/build.gradle
```diff
@@ -55,19 +58,23 @@ def SQLITE_COMPILE_FLAGS = [
...
-// nice to have them there so that we can swap them with SQLite amalgamation for Github.
-// TODO(b/306669673): Support GitHub builds
-def copySqliteSourcesTask = tasks.register("copySqliteSources", Copy) {
-    from(new File(AndroidXConfig.getExternalProjectPath(project), "sqlite/dist/orig")) {
-        include "sqlite3.c"
-        include "sqlite3.h"
+// see PrepareSqliteSourcesTask.groovy for implementation details of the register method
+def prepareSqliteSourcesTask = registerPrepareSqliteSourcesTask(
+        project, // project
+        "prepareSqliteSourcesTask" // task name
```
liutikas, line 67: you can use named arguments in groovy
registerPrepareSqliteSourcesTask(project: project, name: "prepareSqliteSourcesTask")

sqlite/sqlite-bundled/build.gradle
```diff
@@ -55,19 +58,23 @@ def SQLITE_COMPILE_FLAGS = [
...
 
-// Copy SQLite sources into a build directory. Even though this is not needed, it will be
-// nice to have them there so that we can swap them with SQLite amalgamation for Github.
-// TODO(b/306669673): Support GitHub builds
-def copySqliteSourcesTask = tasks.register("copySqliteSources", Copy) {
-    from(new File(AndroidXConfig.getExternalProjectPath(project), "sqlite/dist/orig")) {
-        include "sqlite3.c"
-        include "sqlite3.h"
+// see PrepareSqliteSourcesTask.groovy for implementation details of the register method
+def prepareSqliteSourcesTask = registerPrepareSqliteSourcesTask(
```
liutikas, line 65: somewhere within this file we should have a comment explaining the license of sqlite that we are relicensing it here as apache 2.0

liutikas (approved):

sqlite/sqlite-bundled/build.gradle
yigit: not sure about that.
as far as i can see, it requires me to use a Map as an argument in the function declaration, which would make the function declaration less useful.

https://docs.groovy-lang.org/latest/html/documentation/#_named_parameters_2

am i missing something ?

buildSrc/private/src/main/kotlin/androidx/build/clang/KonanBuildService.kt
yigit: not sure if i understand.
this is where we set the parameters so there is nothing to assert.

I could try to assert when we create the distribution (where we read the parameter) but the build service doesn't have access to `Project` so I'm not sure if i can check for ` if(!playground) assertNotNullPrebuilts()`. one option is to pass the project layout type as an argument to the build service.

Is that what you prefer or am i misunderstanding this?

buildSrc/private/src/main/kotlin/androidx/build/clang/NativeTargetCompilation.kt
yigit: this is from my kotlin-sqlite-bindings copy :). i don't remember why so let me try removing it.

sqlite/sqlite-bundled/build.gradle
yigit: added a note on licensing.

buildSrc/private/src/main/kotlin/androidx/build/clang/KonanBuildService.kt
yigit: implemented the option where i pass project layout type.

liutikas: where is this implementation? I cannot seem to find it.

sqlite/sqlite-bundled/build.gradle
liutikas: You are right, I misunderstood how groovy does it. we can leave it as-is

buildSrc/private/src/main/kotlin/androidx/build/clang/KonanBuildService.kt
yigit: mb. the code push didn't go through 🤦  updated now.

yigit: FYI this is the check: https://github.com/androidx/androidx/pull/647/files#diff-4d57a5bf411f5c52d9b975b078f74b4c8034a31922c196ad836485a03f1f130fR56

PR #648, opened 2024-01-17
Modify the offset calculation for adding item after the last item.

goweii:
When the focus is on the last item and an item is added before the focus item. If the previous item of the same row cannot be found, the offset of the adjacent previous item should be used directly. If found, the offset of the item should be added with the size of the item and space as the current offset.

Test: run locally, no crash

## Proposed Changes

  - Modify the offset calculation for adding item after the last item.

## Testing

Test: Verified through local demo, no crashes. And fixed the issues mentioned below.

## Issues Fixed

Fixes: Fixed the issues mentioned below.

When the number of items is less than one line, when the focus is on the last item, if an item is added before the focus, the focus item will be incorrectly placed in the second line, and in reality, there is still space in the first line above it.


goweii:
Please review this PR, I look forward to feedback.

PR #650, opened 2024-01-18
Ignore unnecessary projectOrArtifactUsages

yigit:
This PR updates PlaygroundExtension to include projects that we can support on GitHub even when they were included via projectOrArtifact.

Normally, PlaygroundExtension builds a dependency graph that ignores projectOrArtifact dependencies, which is fine but causes more possible inconsistencies between AOSP and GitHub.
This PR updates it to also generate the AOSP graph (that considers projectOrArtifact == project) and when it finds new projects, includes them unless they have a direct `project` dependency to an unsupported project.

This obviously increases the number of projects but thanks to the Gradle cache and selected test running (playgroundCIHostTests), it is indeed feasible compiling these projects on GitHub.

I've not deleted these unnecessary projectOrArtifact usages to keep this CL focused, will do that in a followup.

Bug: n/a
Test: CI


buildSrc/settingsScripts/project-dependency-graph.groovy
```diff
@@ -35,6 +35,12 @@ class ProjectDependencyGraph {
      * A map of project path to a set of project paths referenced directly by this project.
      */
     private Map<String, Set<String>> projectReferences = new HashMap<String, Set<String>>()
+
+    /**
+     * A map of project path to a set of project paths that directly depend on the key project.
+     */
+    private Map<String, Set<String>> projectReferenceReverseLookup = new HashMap<String, Set<String>>()
```
dlam: Oh geez is there is no way to get a bidirectional map in gradle?

buildSrc/settingsScripts/project-dependency-graph.groovy
```diff
@@ -63,6 +69,31 @@ class ProjectDependencyGraph {
...
+            reverseLookupSet.add(projectPath)
+            projectReferenceReverseLookup[dependency] = reverseLookupSet
+        }
+    }
+
+    /**
+     * Returns a set of project path that includes the given `projectPath` as well as any other project
+     * that directly or indirectly depends on `projectPath`
+     */
+    Set<String> findAllProjectsDependingOn(String projectPath) {
```
dlam, line 83: I'm a bit confused how the order is guaranteed here - we need to make sure addToAllProjects is run first so that the `projectReferenceReverseLookup` is populated, but it is just a property instead of a task input / provider? I'm not sure what the right thing to do here is, so might be better to get someone more familiar with Gradle to review, but I think it would be clearer if we just write a BFS helper with a lambda and inline it below with some comment about the ordering.

```
fun <T> String.forEachBfs(
  projectPathToRefs Map<String, Set<String>>,
  projectRefToPaths Map<String, Set<String>>,
  block: () -> T
): T
```

maybe like this it would be impossible to mess up?

buildSrc/settingsScripts/project-dependency-graph.groovy
yigit: not w/o using guava but i didn't think about finding it. honestly, i'm barely holding up writing these in groovy 😮‍💨 🤷 .
so it works hence enough is my general approach here :).

buildSrc/settingsScripts/project-dependency-graph.groovy
yigit: this is how this class is setup to be able to serve both settings.gradle and PlaygroundExtension.kt.

Projects get added via `addToAllProjects` and these graphs just use that data.

buildSrc/settingsScripts/project-dependency-graph.groovy
liutikas: Can we call this projectConsumers to make it a bit more clear of what this is?

buildSrc/settingsScripts/project-dependency-graph.groovy
```diff
@@ -63,6 +69,31 @@ class ProjectDependencyGraph {
...
+        Set<String> result = new HashSet<String>()
+        ArrayDeque<String> toBeTraversed = new ArrayDeque<String>()
+        toBeTraversed.add(projectPath)
+        while(toBeTraversed.size() > 0) {
```
liutikas: missing space after while

liutikas (commented):
Are we going to have a follow up that updates all the settings.gradle and moving to a disallow list for projects?

yigit:
> Are we going to have a follow up that updates all the settings.gradle and moving to a disallow list for projects?

The plan is to manually replace all projectOrArtifact usages that are not needed, in a followup (i actually did it first but the PR was getting crazy).

We do still need projectOrArtifact in some cases so we cannot delete it yet. 
Furthermore, i'm not sure if we want to turn this into an error (unnecessary usages of projectOrArtifact) because people copy things and it might cause too much noise.

dlam (approved):

liutikas (approved):

PR #669, opened 2024-02-28
Introduce early exit path for  unlikely to be a resource path case.

nodmp:
## Proposed Changes

This CL does some improvements:
- Introduce the early exit path if the given string is not likely the path in resource file. Prevents insertion of a large number of empty messages into the main thread

## Testing

## Issues Fixed

Fixes: The bug on [https://issuetracker.google.com/issues/323403933](https://issuetracker.google.com/issues/323403933) being fixed


dlam:
I am not the main reviewer on appcompat, but my guess is it would be hard to accept this PR because it is not solving the root cause of the problem. Aside from that, we also require a test with every bug fix code change, which in this case would need to clearly show the failing case from the bug being solved by this change.

PR #671, opened 2024-03-04
feat: Allow customizing the Preview's format (e.g. to `ImageFormat.YUV_420_888`)

mrousavy:
## Proposed Changes

  - Add `setFormat(...)` to `Preview.Builder`
  - Use custom format for `SurfaceEdge` in `Preview::createPipeline(..)`
  - Remove format override in `Preview::onMergeConfig`

The reason for this change is that `CameraEffect`s/`SurfaceProcessor`s are always in a fixed format, and in my case I also have an additional CPU processing step in my `SurfaceProcessor`, which I need `ImageFormat.YUV_420_888` images for (instead of `ImageFormat.PRIVATE`).

I cannot use `ImageAnalysis` for this, since I already have Preview, ImageCapture and VideoCapture use-cases attached.

## Testing

Test: I did not figure out how to test this yet, I'll read the contributing guidelines in a sec and start testing this.

## Issues Fixed

Fixes: A comment in https://issuetracker.google.com/issues/326813037

### DIFF
--- a/camera/camera-core/src/main/java/androidx/camera/core/CameraEffect.java
+++ b/camera/camera-core/src/main/java/androidx/camera/core/CameraEffect.java
@@ -108,7 +108,7 @@ public abstract class CameraEffect {
      */
     @Retention(RetentionPolicy.SOURCE)
     @RestrictTo(RestrictTo.Scope.LIBRARY_GROUP)
-    @IntDef(flag = true, value = {INTERNAL_DEFINED_IMAGE_FORMAT_PRIVATE, ImageFormat.JPEG})
+    @IntDef(flag = true, value = {INTERNAL_DEFINED_IMAGE_FORMAT_PRIVATE, ImageFormat.YUV_420_888, ImageFormat.JPEG})
     public @interface Formats {
     }
 
--- a/camera/camera-core/src/main/java/androidx/camera/core/Preview.java
+++ b/camera/camera-core/src/main/java/androidx/camera/core/Preview.java
@@ -229,7 +229,7 @@ public final class Preview extends UseCase {
         CameraInternal camera = requireNonNull(getCamera());
         clearPipeline();
 
-        int format = config.get(OPTION_INPUT_FORMAT);
+        int format = config.getInputFormat();
 
         // Make sure the previously created camera edge is cleared before creating a new one.
         checkState(mCameraEdge == null);
--- a/camera/camera-core/src/main/java/androidx/camera/core/Preview.java
+++ b/camera/camera-core/src/main/java/androidx/camera/core/Preview.java
@@ -229,11 +229,13 @@ public final class Preview extends UseCase {
         CameraInternal camera = requireNonNull(getCamera());
         clearPipeline();
 
+        int format = config.get(OPTION_INPUT_FORMAT);
+
         // Make sure the previously created camera edge is cleared before creating a new one.
         checkState(mCameraEdge == null);
         mCameraEdge = new SurfaceEdge(
                 PREVIEW,
-                INTERNAL_DEFINED_IMAGE_FORMAT_PRIVATE,
+                format,
                 streamSpec,
                 getSensorToBufferTransformMatrix(),
                 camera.getHasTransform(),
@@ -560,9 +562,6 @@ public final class Preview extends UseCase {
     @Override
     protected UseCaseConfig<?> onMergeConfig(@NonNull CameraInfoInternal cameraInfo,
             @NonNull UseCaseConfig.Builder<?, ?, ?> builder) {
-        builder.getMutableConfig().insertOption(OPTION_INPUT_FORMAT,
-                INTERNAL_DEFINED_IMAGE_FORMAT_PRIVATE);
-
         return builder.getUseCaseConfig();
     }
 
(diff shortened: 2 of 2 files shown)


dlam:
We don't normally take contributions for Camera through github (no one from that team is monitoring PRs here), but I can try to mirror it over to AOSP and help get the right people looking at this change.

One thing that we definitely need is a test, as we do not accept any bug fixes without tests for the failing case.

mrousavy:
Ah, gotcha, sorry didn't know.
I'm working on a pretty popular open source Camera library for React Native ([react-native-vision-camera](https://github.com/mrousavy/react-native-vision-camera)) and I currently use Camera2 but want to switch over to CameraX, but I have trouble finding a good person of contact for CameraX to discuss things and best practices with (especially for stuff that isn't in the docs)
Is there any chance you guys have someone on your team with knowledge about CameraX willing to help? We're also funded, so I'd pay for consulting.

Anyways; I could try to add some tests :)

dlam:
I'll reach out - it may take a bit to get back to you due to timezone differences.

Thanks for contributing, but definitely you will need to add tests for this to be considered :)

xizhang:
Thanks for the request Marc. I am an engineer on the CameraX team. Unfortunately, supporting a new format will require more internal change than just the ones in the PR. However you are not alone in wanting YUV as the input format. I will discuss this feature request with the team and  prioritize it accordingly.

mrousavy:
Thanks, appreciate it. Yea, I thought that this might be a bit more complex.

I also think that maybe this should be incorporated into the Video pipeline instead of the Preview pipeline, but the Video pipeline is already streaming in RGB buffers because OpenGL sits between the Camera and the application code where I can work with the Surface, right?

As far as I remember, YUV instead of PRIVATE worked really well for almost all phones, I think some phones just couldn't do YUV 60 FPS but only PRIVATE 60 FPS.

My use case is that I have apps with preview, photo, video and frame processors, and frame processors are able to draw on the resulting video and/or preview (optionally) - so I need to sit in-between the capture pipeline, ImageAnalysis won't work for me.

Thanks for all your hard work!

> I'll reach out - it may take a bit to get back to you due to timezone differences.

If you're talking about me that'd be really cool to chat more about CameraX and using it for VisionCamera - my Twitter DMs are open :)

xizhang:
Thanks for the insight Marc, I was curious about your use case too. As I mentioned in the [SO comment](https://stackoverflow.com/questions/78067957/camerax-custom-opengl-video-pipeline-usecase-videooutput#comment137694830_78075518), while YUV is a viable option, devices are in general better at processing PRIVATE format, so that why we support it in the OpenGL pipeline first. Ideally, developers should be able to choose any input format/output destination for the effects, but we are not there yet.

Preview, VideoCapture, ImageAnalysis and ImageCapture are *outputs* of CameraX. Since you have checked the source code, I want to mention that we are overloading the concept of `UseCase` a bit internally. Internally it also means input streams, which might cause some confusion to the readers.

I understand that some of the CV algorithm runs on YUV and you prefer a YUV input. You can wait for this feature to be implemented, but we haven't planned it out yet. For the time being, you can choose to do one of the following:

1. Use ImageAnalysis's YUV output as the input, then write to the output Surface in the SurfaceProcessor. The input of SurfaceProcessor is ignored.
2. Convert SurfaceProcessor's input to YUV format. e.g. downloading RGB bytes from GPU then convert it to YUV.

Personally I would go with 1) since it's less work.

dlam:
Closing this out since it seems we're going a different direction for this change, but feel free to keep commenting if you want to keep chatting here :)

mrousavy:
Thanks @xizhang!

Yea I gotta say, CameraX is very abstract, but the implementation is quite impressive. Nice work.

Regarding option 1; In that case I'd have to remove one of the use-cases, Preview, Photo or Video right? Because only a maximum of 3 outputs are supported.

So then I'd just treat ImageAnalysis as e.g. a Preview output then, where I first read the Frames, run my processing, and then draw them to the PRIVATE Preview Surface using OpenGL? 

```kt
// pseudo code
val reader = ImageReader.newInstance(...)
reader.addOnImageAvailableListener({ reader ->
  val image = reader.acquireNextImage() // <-- get latest image
  frameProcessor(image) // <-- run my frame processing
  renderImage(image) // <-- draw frame to Preview Surface using OpenGL
}, executor)
```

Are there any limits on ImageAnalysis? Some of my users are doing 60 FPS frame processing, and some are doing stuff like uploading the Frames to a server (WebRTC) and need 4k buffers.

With that said, I'm definitely excited for the custom formats feature in SurfaceProcessors. If ImageAnalysis is too limiting for my use-cases, I'd probably have to wait for that feature. :)

mrousavy:
Probably a dumb question, but there's no native API to convert PRIVATE Images/AHardwareBuffers to YUV_420_888, is there? 
Camera reprocessing also doesn't do conversions, does it? So input PRIVATE, get output YUV_420_888?

xizhang:
Depending on the device, the camera may support 3 or 4 concurrent streams. But CameraX supports a larger number of outputs than the camera supports. CameraX provides this feature called "stream sharing" that copies one PRIVATE stream to both Preview and VideoCapture when the device does not support 4 streams. 

This is what I would do:
1. Bind Preview, VideoCapture, ImageAnalysis and ImageCapture to the camera.
2. Set up CameraEffect/SurfaceProcessor. Normally both the input and output are PRIVATE Surfaces. CameraX copies the output to both Preview and VideoCapture as per the "stream sharing" feature.
3. As a workaround, we ignore the input of SurfaceProcessor. We still provide a Surface to the SurfaceRequest but the updates are ignore. Instead, we process the ImageProxy from ImageAnalysis and draw it to the output Surface of SurfaceProcessor.

This way, you will have YUV input and PRIVATE output. The processing result are sent to Preview and VideoCapture, and you can have all the UseCases. 

As for 4K / 60fps YUV support, it depends on the device. And it's difficult to learn what the device supports without trying it out. If you want to experiment with it, you can configure the resolution using the [ImageAnalysis#setResolutionSelector](https://developer.android.com/reference/androidx/camera/core/ImageAnalysis.Builder#setResolutionSelector(androidx.camera.core.resolutionselector.ResolutionSelector)) API, and configure the FPS using the [Camera2Interop](https://developer.android.com/reference/androidx/camera/camera2/interop/Camera2Interop) API.

xizhang:
And there isn't an API for converting PRIVATE to CPU accessible YUV. The best way is using OpenGL to sample the PRIVATE input and convert it yourself.

Out of curiosity, why does `react-native-vision-camera` want to process YUV and write it to Preview/VideoCapture? It's less efficient than running the pipeline on PRIVATE from end to end. Does your CV algorithm require a YUV input?

PR #683, opened 2024-05-26
Fix: webp files corrupted by ExifInterface.saveAttributes

CoXier:
## Proposed Changes

  - webp file without vp8x header (maybe old format ? )
  - width or height exceeds 8191
  - ExifInterface.saveAttributes

## Testing

## Issues Fixed

Fixes: The bug on [34269705](https://issuetracker.google.com/issues/342697059) being fixed

### DIFF
--- a/exifinterface/exifinterface/src/main/java/androidx/exifinterface/media/ExifInterface.java
+++ b/exifinterface/exifinterface/src/main/java/androidx/exifinterface/media/ExifInterface.java
@@ -3093,8 +3093,6 @@ public class ExifInterface {
     private static final int WEBP_CHUNK_TYPE_VP8X_DEFAULT_LENGTH = 10;
     private static final int WEBP_CHUNK_TYPE_BYTE_LENGTH = 4;
     private static final int WEBP_CHUNK_SIZE_BYTE_LENGTH = 4;
-    // maximum width/height allowed (inclusive), in pixels
-    private static final int WEBP_MAX_DIMENSION = 16383
 
     private static SimpleDateFormat sFormatterPrimary;
     private static SimpleDateFormat sFormatterSecondary;
@@ -6776,11 +6774,8 @@ public class ExifInterface {
 
                         // Retrieve image width/height
                         widthAndHeight = totalInputStream.readInt();
-                        int width = widthAndHeight & 0x0000FFFF;
-                        int height = (widthAndHeight & 0xFFFF0000) >> 16;
-                        if (width > WEBP_MAX_DIMENSION || height > WEBP_MAX_DIMENSION) {
-                            throw new IOException("Maximum side allowed is 16383 pixels.");
-                        }
+                        width = widthAndHeight & 0x3FFF;
+                        height = (widthAndHeight >> 16) & 0x3FFF;
                         bytesToRead -= (vp8Frame.length + vp8Signature.length + 4);
                     } else if (Arrays.equals(firstChunkType, WEBP_CHUNK_TYPE_VP8L)) {
                         // Check signature
--- a/exifinterface/exifinterface/src/main/java/androidx/exifinterface/media/ExifInterface.java
+++ b/exifinterface/exifinterface/src/main/java/androidx/exifinterface/media/ExifInterface.java
@@ -3093,6 +3093,8 @@ public class ExifInterface {
     private static final int WEBP_CHUNK_TYPE_VP8X_DEFAULT_LENGTH = 10;
     private static final int WEBP_CHUNK_TYPE_BYTE_LENGTH = 4;
     private static final int WEBP_CHUNK_SIZE_BYTE_LENGTH = 4;
+    // maximum width/height allowed (inclusive), in pixels
+    private static final int WEBP_MAX_DIMENSION = 16383
 
     private static SimpleDateFormat sFormatterPrimary;
     private static SimpleDateFormat sFormatterSecondary;
@@ -6774,8 +6776,11 @@ public class ExifInterface {
 
                         // Retrieve image width/height
                         widthAndHeight = totalInputStream.readInt();
-                        width = (widthAndHeight << 18) >> 18;
-                        height = (widthAndHeight << 2) >> 18;
+                        int width = widthAndHeight & 0x0000FFFF;
+                        int height = (widthAndHeight & 0xFFFF0000) >> 16;
+                        if (width > WEBP_MAX_DIMENSION || height > WEBP_MAX_DIMENSION) {
+                            throw new IOException("Maximum side allowed is 16383 pixels.");
+                        }
                         bytesToRead -= (vp8Frame.length + vp8Signature.length + 4);
                     } else if (Arrays.equals(firstChunkType, WEBP_CHUNK_TYPE_VP8L)) {
                         // Check signature
--- a/exifinterface/exifinterface/src/androidTest/java/androidx/exifinterface/media/ExifInterfaceTest.java
+++ b/exifinterface/exifinterface/src/androidTest/java/androidx/exifinterface/media/ExifInterfaceTest.java
@@ -50,7 +50,6 @@ import com.google.common.primitives.Ints;
 import com.google.common.truth.Expect;
 
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.TemporaryFolder;
@@ -390,7 +389,6 @@ public class ExifInterfaceTest {
 
     // https://issuetracker.google.com/342697059
     @Test
-    @Ignore("Enable as part of merging the fix in https://github.com/androidx/androidx/pull/683")
     @LargeTest
     public void testWebpWithoutExifHeight8192px() throws Throwable {
         File imageFile =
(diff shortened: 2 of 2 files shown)


CoXier:
Hi @dlam @yigit 

Could you please provide an update on its review status? Thank you!

exifinterface/exifinterface/src/main/java/androidx/exifinterface/media/ExifInterface.java
```diff
@@ -6728,8 +6730,11 @@ private void saveWebpAttributes(InputStream inputStream, OutputStream outputStre
 
                         // Retrieve image width/height
                         widthAndHeight = totalInputStream.readInt();
-                        width = (widthAndHeight << 18) >> 18;
-                        height = (widthAndHeight << 2) >> 18;
+                        int width = widthAndHeight & 0x0000FFFF;
+                        int height = (widthAndHeight & 0xFFFF0000) >> 16;
+                        if (width > WEBP_MAX_DIMENSION || height > WEBP_MAX_DIMENSION) {
+                            throw new IOException("Maximum side allowed is 16383 pixels.");
```
icbaker: Reading https://datatracker.ietf.org/doc/html/rfc6386#section-9.1 it looks like this is really asserting that both the horizontal and vertical scaling factors are zero (meaning no upscaling).

The current code looks like it aims to ignore the scaling factor instead, and this seems to match this wording in the WebP spec:
> Note that the VP8 frame header contains the VP8 frame width and height. That is assumed to be the width and height of the canvas.

Assuming that "frame width" refers to the 14 bits of width in the VP8 frame, ignoring the scaling factor (which seems to fit the wording of rfc6386 too.

I'd suggest changing your code to ignore the scaling factor too - I think you can most easily do that by amending your bit masks to only read 14 bits each (and then no need to check the size, since you can't read more than 16383 if you only read 14 bits).

icbaker:
Please can you add a test image and test case that fails without this fix?

Maybe you could make a copy of [`webp_without_exif.webp`](https://github.com/androidx/androidx/blob/androidx-main/exifinterface/exifinterface/src/androidTest/res/raw/webp_without_exif.webp) and resize it to 8192 or greater in one dimension (just be careful that whatever tool you use to resize doesn't "usefully" add a VP8X or EXIF chunk).

CoXier:
> Please can you add a test image and test case that fails without this fix?
> 
> Maybe you could make a copy of [`webp_without_exif.webp`](https://github.com/androidx/androidx/blob/androidx-main/exifinterface/exifinterface/src/androidTest/res/raw/webp_without_exif.webp) and resize it to 8192 or greater in one dimension (just be careful that whatever tool you use to resize doesn't "usefully" add a VP8X or EXIF chunk).

I think this bug is very clear (I explain why this happens). I'm not familiar how to add test case to this repo and I can provide a test image.

icbaker:
I've added a test that exercises this bug, and marked it `@Ignore`, and sent it for review: http://r.android.com/3204431

I think there are a few options:
1. You copy what I've done there into this PR (I included the command I used to generate the image) and I drop that change
2. You wait for that to be merged, then rebase on top of it and remove the `@Ignore` in this PR
3. You drop this PR and I merge the fix into the same change  as the test (this is probably the quickest option)

CoXier:
Thanks  for  your  work. 2nd option  is  better for  me.

In  the past few months , I tried to  build  the  androidx repo using
Android studio , however I failed. Because of that，I cannot write the test
for this fix.

Thanks a lot  for  your  work，this is my first  pr for  Google  Android.

Ian Baker ***@***.***>于2024年8月1日 周四23:23写道：

CoXier:
> I've added a test that exercises this bug, and marked it `@Ignore`, and sent it for review: http://r.android.com/3204431
> 
> I think there are a few options:
> 
> 1. You copy what I've done there into this PR (I included the command I used to generate the image) and I drop that change
> 2. You wait for that to be merged, then rebase on top of it and remove the `@Ignore` in this PR
> 3. You drop this PR and I merge the fix into the same change  as the test (this is probably the quickest option)

Thank you for your work. The second option works better for me.

Over the past few months, I attempted to build the AndroidX repository using Android Studio, but unfortunately, I was unsuccessful. As a result, I am unable to write tests for this fix.

I really appreciate your efforts; this is my first pull request for Google Android.

CoXier:
@icbaker I already rebase the androidx-main branch and enable the test.

icbaker:
Thanks, please can you take a look at my comment on L6736 of your code as well.

CoXier:
> Thanks, please can you take a look at my comment on L6736 of your code as well.

Where can I find the comment ? Github ?

icbaker (changes requested):

icbaker:
> > Thanks, please can you take a look at my comment on L6736 of your code as well.
> 
> Where can I find the comment ? Github ?

It helps if i actually send it - sorry...

exifinterface/exifinterface/src/main/java/androidx/exifinterface/media/ExifInterface.java
CoXier: I think we should inform the caller about the maximum supported dimensions. If we only read 14 bits, the caller won't be able to perceive any abnormality.

icbaker: Not sure I understand. Each dimension is only permitted to be 14 bits, the other 2 bits of the 16 are not dimensions, they are a scaling factor that isn't relevant here, so not reading these 2 bits seems correct to me.

CoXier: Thank you for your review. After carefully studying the RFC 6386 document and the VP8 specification, I realize that my initial understanding was incorrect. You are right; the width and height should indeed only use 14 bits, not 16.

I agree with your suggestion to modify the code to read only 14 bits for width and height. This not only more accurately follows the specification but also automatically limits the maximum dimensions to 16383x16383 pixels without requiring additional checks.

I will update the code according to your recommendation, using 14-bit masks to read the width and height. Additionally, I think it would be beneficial to explicitly state this maximum size limitation in the documentation to ensure callers are aware of this constraint.

icbaker: > Additionally, I think it would be beneficial to explicitly state this maximum size limitation in the documentation to ensure callers are aware of this constraint.

I don't think it's necessary for `ExifInterface`'s documentation to duplicate information that is available in each image format spec (as is the case here). Otherwise it risks getting very unwieldy - since we end up trying to include all the information from the specs of the many image formats we support.

CoXier: I updated my code just now. Please have a look.

icbaker (approved):

icbaker:
Looks good to me, @dlam or @yigit please can you take a look and merge this?

icbaker:
> Looks good to me, @dlam or @yigit please can you take a look and merge this?

I misunderstood the process - it's being merged in http://r.android.com/3204842 which will then close this PR once it's submitted.

exifinterface/exifinterface/src/main/java/androidx/exifinterface/media/ExifInterface.java
```diff
@@ -6774,8 +6774,8 @@ private void saveWebpAttributes(InputStream inputStream, OutputStream outputStre
 
                         // Retrieve image width/height
                         widthAndHeight = totalInputStream.readInt();
-                        width = (widthAndHeight << 18) >> 18;
-                        height = (widthAndHeight << 2) >> 18;
+                        width = widthAndHeight & 0x3FFF;
+                        height = (widthAndHeight >> 16) & 0x3FFF
```
icbaker: Please can you add the missing semi-colon at the end of this line

CoXier: Sorry for my mistake. I love Kotlin lol

icbaker:
I'll take a look at the test failure on http://r.android.com/3204842 tomorrow - it looks like my test causes an OOM when trying to decode the large WebP file to check it's still valid.

icbaker:
The test is only failing on API 21 on the [FTL](https://firebase.google.com/docs/test-lab) emulators used in the automated testing (I wasn't able to repro the same failre on an API 21 x86 emulator launched from Android Studio locally).

I suggest annotating this test with the following line:

```
@SdkSuppress(minSdkVersion = 22) // Parsing the large image causes OOM on API 21 FTL emulators.
```

CoXier:
> The test is only failing on API 21 on the [FTL](https://firebase.google.com/docs/test-lab) emulators used in the automated testing (I wasn't able to repro the same failre on an API 21 x86 emulator launched from Android Studio locally).
> 
> I suggest annotating this test with the following line:
> 
> ```
> @SdkSuppress(minSdkVersion = 22) // Parsing the large image causes OOM on API 21 FTL emulators.
> ```

I wonder whether the full classname is androidx.test.filters.SdkSuppress?

icbaker (approved):

CoXier:
@icbaker Now that my code changes have been merged, I was wondering if there's an estimated timeline for the next release of androidx.exifinterface:exifinterface that would include these updates.

icbaker:
We are planning a 1.4.0 release, but I'm afraid we don't have a timeline to share at the moment.

PR #689, opened 2024-08-12
Update .github flows to use ENV files over set-output

dlam:
set-output is now deprecated. See: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/

Test: GH Workflows


dlam:
Sample run which shows this works, although still failing for other reasons in other projects: https://github.com/dlam/androidx/actions/runs/10361166927/job/28680991297

liutikas (approved):

PR #690, opened 2024-08-13
Bump NDK for GH to match AOSP

dlam:
Taken from https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:settings.gradle;l=87

Test: GH Workflows
Change-Id: I726b0aadfdcf86014bdb699268ba4f3dd0817688

### DIFF
--- a/.github/actions/build-single-project/action.yml
+++ b/.github/actions/build-single-project/action.yml
@@ -33,7 +33,9 @@ runs:
       run: echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "cmake;3.22.1"
     - name: "Install NDK"
       shell: bash
-      run: echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "ndk;23.1.7779620"
+      run: |
+        NDK_VERSION=$(grep "ndkVersion" settings.gradle | awk -F "=" '{gsub(/"| /, ""); print $2}')
+        echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "ndk;$NDK_VERSION"
     - name: "Install Android SDK Build-Tools"
       shell: bash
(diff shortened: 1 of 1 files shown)


.github/actions/build-single-project/action.yml
```diff
@@ -33,7 +33,7 @@ runs:
       run: echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "cmake;3.22.1"
     - name: "Install NDK"
       shell: bash
-      run: echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "ndk;23.1.7779620"
+      run: echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "ndk;27.0.12077973"
```
liutikas: Can we read the value from settings.gradle and use that, so we dont diverge again?

dlam: Done - but a bit brittle because we need to read it from bash via string match :(

Less brittle than before though.

.github/actions/build-single-project/action.yml
```diff
@@ -33,7 +33,9 @@ runs:
       run: echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "cmake;3.22.1"
     - name: "Install NDK"
       shell: bash
-      run: echo "yes" | $ANDROID_SDK_ROOT/cmdline-tools/latest/bin/sdkmanager --install "ndk;23.1.7779620"
+      run: |
+        NDK_VERSION=$(grep "ndkVersion" settings.gradle | awk -F "=" '{gsub(/"| /, ""); print $2}')
```
liutikas, line 37: Can you fail hard if it fails to find a reasonable value?

liutikas (approved):

.github/actions/build-single-project/action.yml
dlam: Won't it fail when it tries to run `sdkmanager` with the bad value? I added `set -x` for logging as well.

liutikas (approved):

PR #691, opened 2024-08-14
Build and test collections on Kotlin/JS

swankjesse:
## Proposed Changes

This adds a new JavaScript target for annotations/ and collections/ modules.

## Testing

This runs the existing KMP tests on the new JavaScript target.

## Issues Fixed

This unblocks fixing [Issue 359253678](https://issuetracker.google.com/issues/359253678)


collection/collection/src/nonJsTest/kotlin/androidx/collection/ArraySetConcurrencyTest.kt
```diff
@@ -0,0 +1,108 @@
...
+import kotlinx.coroutines.cancel
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.isActive
+import kotlinx.coroutines.launch
+import kotlinx.coroutines.test.runTest
+
+internal class ArraySetConcurrencyTest {
+    private val set = ArraySet<String>()
+
+    /**
```
swankjesse, line 34: Extracted from `ArraySetTest`.

liutikas:
We are not quite ready to enable javascript targets in androidx, we have some infrastructure bits to work through in our AOSP builds (fully offline, compatible with configuration cache, etc) before we can do this.

https://issuetracker.google.com/issues/357844126 is the bug to follow for that support.

Once javascript is enabled (currently targeting to do it for annotation and collection), we can take this PR for the perf optimizations for this library.

swankjesse:
@liutikas good to hear! And I’m looking forward to that.

We’ll ship off a [tracking branch](https://github.com/cashapp/compose-multiplatform-core/) in the interim. Hopefully that’s short-lived!

rotilho:
Now that AndroidX supports JS, is there any chance of resuming this work? It feels like the perfect opportunity to pick it back up!

swankjesse:
Yes! Thanks for the ping!

dlam:
wasm and js have been enabled for collections and released as part of 1.5.0-beta01.

Please note web falls under our experimental / unsupported targets, so it is "beta" but mostly for other platforms:
https://developer.android.com/kotlin/multiplatform#supported-platforms

PR #693, opened 2024-08-21
Move ndkVersion setting into a shared .gradle file

dlam:
Playground builds were overriding the installed ndk version due to agp default and lack of ndkVersion declaration in settings.gradle.

Test: ./gradlew datastore:datastore-core:tasks
Change-Id: Id6d8d9cff46ba3aa3e929e4a7df4912672a12f3a

### DIFF
--- a/playground-projects/activity-playground/settings.gradle
+++ b/playground-projects/activity-playground/settings.gradle
@@ -20,7 +20,10 @@ pluginManagement {
 }
 plugins {
     id "playground"
+    id "com.android.settings" version "8.7.0-alpha02"
 }
+
+apply ../../buildSrc/ndk.gradle
 
 rootProject.name = "activity-playground"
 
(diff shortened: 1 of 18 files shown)


dlam:
I tried for a bit, but couldn't figure out how to declare it in toml. I would also prefer to share the plugin version and apply it from a common place, such as from playground plugin.

playground-projects/activity-playground/settings.gradle
```diff
@@ -20,8 +20,11 @@ pluginManagement {
 }
 plugins {
     id "playground"
```
liutikas, line 22: Why don't we make playground plugin apply the version, so we dont have to add this to every project?

dlam: Yeah I'd prefer that but what's the magic incantation to make that work?

liutikas (approved):

PR #702, opened 2024-09-22
Use sequences for lazy evaluation

GeorgCantor:
## Proposed Changes

Use of Sequences: By converting the collection to a sequence using asSequence(), we enable lazy evaluation, which can improve performance when dealing with large collections by avoiding unnecessary intermediate collections.

Map Not Null: Instead of filtering and then casting, we use mapNotNull to both filter and cast in one step, making the code cleaner.

## Testing

Manually

## Issues Fixed

Use sequences for lazy evaluation


dlam:
Drive-by review - I'll defer to @jbw0033 and @ianhanniballake as the owners if they want to override.

In general manually testing is not good enough. Performance improvements should have a microbenchmark that shows the improvement. How much overhead does this add and at what point does it start paying off?

We are also still filtering twice on the key and value separately, and it seems like we should combine that too.

romainguy:
This begs to become a simple for loop :)

PR #703, opened 2024-09-23
fix: room common js target

sgammon:
## Proposed Changes

  - Add Kotlin `js` target to `annotation/annotations`
  - Add Kotlin `js` target to `room/room-common`

## Testing

Test: This modifies the build only, by adding a new Kotlin target; since room-common is just annotations, the build passing is itself the test.

## Issues Fixed

Fixes: _[No JavaScript variant published for Room Common](https://issuetracker.google.com/issues/359296625)_

### DIFF
--- a/annotation/annotation/build.gradle
+++ b/annotation/annotation/build.gradle
@@ -25,6 +25,7 @@ androidXMultiplatform {
     watchosDeviceArm64()
     watchos()
     tvos()
+    js()
     wasmJs()
 
     defaultPlatform(PlatformIdentifier.JVM)
--- a/room/room-common/build.gradle
+++ b/room/room-common/build.gradle
@@ -36,6 +36,7 @@ androidXMultiplatform {
     mac()
     linux()
     ios()
+    js()
 
     defaultPlatform(PlatformIdentifier.JVM)
 


ianhanniballake:
As per [this comment](https://github.com/androidx/androidx/pull/691#issuecomment-2289896295):

> We are not quite ready to enable javascript targets in androidx, we have some infrastructure bits to work through in our AOSP builds (fully offline, compatible with configuration cache, etc) before we can do this.
>
> https://issuetracker.google.com/issues/357844126 is the bug to follow for that support.

sgammon:
@ianhanniballake No worries, thanks for the link. In the meantime, I haven't been able to build a Room JAR to try this in our own worfklow. Do you have any pointers by any chance?

While I can complete a `./cleanBuild.sh assembleRelease`, I don't see where JARs/AARs are ending up.

danysantiago:
Try running `./gradlew :room:room-common:publishToMavenLocal` and the artifacts should end up in your `~/.gradle/.m2` folder.

dlam:
Closing this as it's infeasible (for now).

sgammon:
@danysantiago I've run commands like that, but with `--info` I get:
```
> Task :room:room-common:publishToMavenLocal
Skipping task ':room:room-common:publishToMavenLocal' as it has no actions.
```

rotilho:
@dlam Is it work that could be resumed now that AndroidX supports JS?

dlam:
Annotations already has a web release under 1.9.1

https://maven.google.com/web/m_index.html?q=Anno#androidx.annotation:annotation-js:1.9.1

For room it is a bit more complicated. We are working on the web story, but I don't think it will be trivial and will require some design work, so I expect it to take some time.

rotilho:
Thanks so much for the quick reply, @dlam!

That makes sense. Do you think Room will be targeting 2.7, or is that unrealistic?

dlam:
Unlikely since web sqlite apis are async and room's aren't under the hood, so we likely need to do some significant changes. It's still quite early and we don't have a set path yet, even the sqlite drivers are not set in stone.

PR #706, opened 2024-10-08
Fix some of playground build issues

liutikas:
## Proposed Changes

- Move to PlaygroundPlugin for applying com.android.settings plugin this allows us to use the same version without it being spread to many files
- Pick a protobuf-java version (3.22.3) that works with AGP in shared-dependnecies.gradle

## Testing

Test: ./gradlew bOS


dlam (approved):

dlam:
Seems failing due to config cache

PR #718, opened 2024-10-10
Avoid double-traversal to find root coordinates in the `boundsInWindow` method

romtsn:
## Proposed Changes

- Replace `boundsInRoot()` with `root.localBoundingBoxOf(this)`, because `boundsInRoot()` does yet another `findRootCoordinates()` pass which is redundant at this point, since we've just found the root coordinates a few lines above

## Testing

Test: I tested it manually in my local fork of this method, plus, existing tests should cover it.

## Issues Fixed

...


yigit (approved):
approving to trigger PR copy to AOSP

PR #720, opened 2024-10-10
Enable running lint-playground

liutikas:
## Proposed Changes

Enable running lint-playground

## Testing

Test: ./gradlew bOS


dlam:
Can you also update CONTRIBUTING.md to include lint in the list of projects we support?

dlam (approved):

PR #725, opened 2024-11-08
Add support for List<Enum> arguments in typesafe navigation

WonderCsabo:
## Proposed Changes

Add support for List<Enum> arguments in typesafe navigation

## Testing

Test: ./gradlew navigation:navigation-common:test
Test: ./gradlew navigation:navigation-common:cC
Test: ./gradlew navigation:navigation-runtime:cC

## Issues Fixed

Fixes: 375559962

### DIFF
--- a/navigation/navigation-common/src/main/java/androidx/navigation/serialization/NavTypeConverter.kt
+++ b/navigation/navigation-common/src/main/java/androidx/navigation/serialization/NavTypeConverter.kt
@@ -101,6 +101,11 @@ internal fun SerialDescriptor.getNavType(): NavType<*> {
                     InternalType.LONG -> NavType.LongListType
                     InternalType.STRING -> NavType.StringListType
                     InternalType.STRING_NULLABLE -> InternalNavType.StringNullableListType
+                    InternalType.ENUM ->
+                        @Suppress("UNCHECKED_CAST")
+                        InternalNavType.EnumListType(
+                            getElementDescriptor(0).getClass() as Class<Enum<*>>
+                        )
                     else -> UNKNOWN
                 }
             }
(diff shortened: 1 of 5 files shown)


navigation/navigation-common/src/main/java/androidx/navigation/serialization/NavTypeConverter.kt
```diff
@@ -471,6 +476,46 @@ internal object InternalNavType {
...
+            previousValue?.plus(parseValue(value)) ?: parseValue(value)
+
+        override fun valueEquals(value: List<D>?, other: List<D>?): Boolean {
+            val valueArrayList = value?.let { ArrayList(value) }
+            val otherArrayList = other?.let { ArrayList(other) }
+            return valueArrayList == otherArrayList
+        }
+
+        override fun serializeAsValues(value: List<D>?): List<String> =
+            value?.map { it.toString() } ?: emptyList()
```
claraf3, line 504: Let's encode the string ouput since the internal parser decodes string inputs.

`value?.map { Uri.encode(it.toString()) } ?: emptyList()`

claraf3:
Hi, thanks for the PR!

Can you please also add tests in [RouteDecoderTest.kt](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:navigation/navigation-common/src/androidTest/java/androidx/navigation/serialization/RouteDecoderTest.kt)? It's basically the opposite equivalent of RouteFilledTest.

WonderCsabo:
Sure, i added the tests there.

navigation/navigation-common/src/main/java/androidx/navigation/serialization/NavTypeConverter.kt
WonderCsabo: Are you sure it is needed? `NavType.EnumType` does not use uri encoding either. It does not seem necessary, since enum entry names can only contain characters which stay the same after uri encoding.

claraf3: The output of serializeAsValues (and serializeAsValue) [should be encoded](https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:navigation/navigation-common/src/main/java/androidx/navigation/NavType.kt;l=132?q=NavType). Though in general that mostly applies to user implementations where we have no control over the final string output. In our case I think it's fine to skip encoding since we know what format the output will be.

WonderCsabo: Okay, then I will leave the code as is.

ianhanniballake (approved):

ianhanniballake:
Thanks so much for fixing this!

WonderCsabo:
@ianhanniballake this fix was [released](https://issuetracker.google.com/issues/375559962#comment7) in [2.8.5](https://developer.android.com/jetpack/androidx/releases/navigation#2.8.5) and [2.9.0-alpha04](https://developer.android.com/jetpack/androidx/releases/navigation#2.9.0-alpha04), but it is not listed in the release notes. Can you add there, please?

ianhanniballake:
We've added a note to the release notes - thanks again!

In the future, note that you can add a `Relnote: ` line in the pull request's description to ensure it gets into the release notes automatically.

PR #736, opened 2024-12-20
Change Dialog 's getWindow() to NonNull

liu-wanshun:
## Proposed Changes

  - `getWindow()` is NonNull ,`mWindow`is initialized during construction and has not been modified since then.

## Testing

Test: override `getWindow()` in subclass of dialog，no problem.


JakeWharton:
The place to fix this, if it indeed is worth changing, would be in the framework, not some arbitrary subclass. But keep in mind that the nullability of these methods models the entire history of behavior–not just the current behavior.

liu-wanshun:
> The place to fix this, if it indeed is worth changing, would be in the framework, not some arbitrary subclass. But keep in mind that the nullability of these methods models the entire history of behavior–not just the current behavior.


@JakeWharton yes,this problem originates from framework, but it is impossible to be null.
image

liu-wanshun:
I also submitted a patch for framework.
https://android-review.googlesource.com/c/platform/frameworks/base/+/3428320

PR #738, opened 2025-01-03
Add enabled param to PullToRefreshBox

jarroyoesp:
## Proposed Changes

`PullToRefreshBox` doesn't have the option to enalbe/disable `PullToRefresh`. Added this param to handle this behaviour.

## Testing

Test: Manually


jarroyoesp:
There is already a feature request to add this flag here: https://issuetracker.google.com/issues/369044003 and seems that they are already solving the issue.

PR #743, opened 2025-01-08
Fix a possible typo `inPredictiveBack` in local variable names, replacing it with `isPredictiveBack`

ShreckYe:
## Proposed Changes

- Fix a possible typo `inPredictiveBack` in local variable names, replacing it with `isPredictiveBack`
  The variable names `isPredictiveBack` and `isPredictiveBackInProgress` are used elsewhere, so I think this should be a typo. `inPredictiveBack` also doesn't make sense by itself.

## Testing

Test: Run `./gradlew :navigation:connectedCheck` in "playground-projects/navigation-playground".

### DIFF
--- a/navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavHost.kt
+++ b/navigation/navigation-compose/src/main/java/androidx/navigation/compose/NavHost.kt
@@ -510,7 +510,7 @@ public fun NavHost(
     val currentBackStack by composeNavigator.backStack.collectAsState()
 
     var progress by remember { mutableFloatStateOf(0f) }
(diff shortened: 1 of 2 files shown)


ianhanniballake:
This is not a typo - the variable is specifically tracking whether we are in the middle of a predictive back gesture - e.g., `inPredictiveBack`.

`isPredictiveBackInProgress` isn't a term used in any Navigation file and, while is close to the same meaning as our `inPredictiveBack`, "progress" as a term is a very specific subset of the Predictive Back API surface related to the dispatch of progress updates, which this variable is not associated with.

ShreckYe:
Got this. Thank you for your kind explanation.

PR #754, opened 2025-01-24
Don't drop retry/refresh events before uiReceiver is set

evant:
## Proposed Changes
Because uiReceiver is set asynchronously it's possible to race with a call to retry or refresh on the PagingDataPresenter.

Changed the implementation to record calling these methods and then forward the event when the uiReceiver is set.

## Testing

## Issues Fixed

Fixes: b/330725007


claraf3 (approved):
Hi Eva, thanks for the PR!

PR #757, opened 2025-02-26
Backport request orientation cache

pyricau:
API 35 / Android 15 added a field cache to `Activity.setRequestedOrientation()` because every call would lead to an IPC call, which can be expensive and hammer the system.

https://cs.android.com/android/_/android/platform/frameworks/base/+/14b59e53b82dbeb93ca1c8ac612c13d62c8672e9

## Proposed Changes

This change backports the API 35 behavior.

Note: I made the change in `androidx.core.app.ComponentActivity` as that's the top level base activity class. Let me know if you'd like me to move it somewhere else instead (or feel free to just edit).

## Testing

Test: we tested this change in our own base activity class in the Square Point of Sale app.

## Issues Fixed

Fixes: 265021810 (not public, I couldn't read it, but it's referenced in the Android 15 change)

### DIFF
--- a/core/core/src/main/java/androidx/core/app/ComponentActivity.kt
+++ b/core/core/src/main/java/androidx/core/app/ComponentActivity.kt
@@ -50,6 +50,8 @@ public open class ComponentActivity : Activity(), LifecycleOwner, KeyEventDispat
      */
     @Suppress("LeakingThis") private val lifecycleRegistry = LifecycleRegistry(this)
 
+    private var lastRequestedOrientation: Int = SCREEN_ORIENTATION_UNSET
+
     /**
      * Store an instance of [ExtraData] for later retrieval by class name via [getExtraData].
      *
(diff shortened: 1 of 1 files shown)


pyricau:
I'm confused why the CLA check is failing, this is what I see when I click the link:

image

dlam:
Our GH build is not in a good state, so I would ignore failures in the checks here, but just enabling it so copybara can mirror the change to Gerrit once we review

ianhanniballake (commented):
The original change also returned the `lastRequestedOrientation` from `getRequestedOrientation`. Was there a reason you didn't override that method also to returned the cached value (as that also seems to do the same binder transaction every time it is called)?

pyricau:
> The original change also returned the `lastRequestedOrientation` from `getRequestedOrientation`. Was there a reason you didn't override that method also to returned the cached value (as that also seems to do the same binder transaction every time it is called)?

... I totally missed it. cs.android.com changed its diff rendering at some point, idk why but some lines aren't fully covered in color

image

ianhanniballake (approved):

core/core/src/main/java/androidx/core/app/ComponentActivity.kt
```diff
@@ -50,6 +50,8 @@ public open class ComponentActivity : Activity(), LifecycleOwner, KeyEventDispat
      */
     @Suppress("LeakingThis") private val lifecycleRegistry = LifecycleRegistry(this)
 
+    private var lastRequestedOrientation: Int = ActivityInfo.SCREEN_ORIENTATION_UNSET
```
pyricau: I need to actually backport that constant as it's hidden.

ianhanniballake (changes requested):
Looks like the formatting needs to be corrected before this can be submitted:

```
Execution failed for task ':core:core:ktCheck'.
> Failed check for the following files:
  $SUPPORT/core/core/src/main/java/androidx/core/app/ComponentActivity.kt

  ********************************************************************************
  You can automatically fix these issues with:
  ./gradlew :core:core:ktFormat
  ********************************************************************************
```

PR #758, opened 2025-02-27
Coerce slider range when dragging to avoid overlap

ln-12:
The material3 [RangeSlider implementation](https://github.com/androidx/androidx/blob/2786b9ec613c6dd14a72914c2213626f24b2cac3/compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/Slider.kt#L4) can crash for large float values when you try to drag one handle beyond the other one as there can be rounding errors for large float values. Details for this crash can be found [here](https://issuetracker.google.com/issues/324934900).  In a first attempt to fix it, a tolerance was introduced in [this PR](https://android-review.googlesource.com/c/platform/frameworks/support/+/2974112/3/compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/Slider.kt#2085) to mitigate this crash. 
However, the crash can still occur when you are using large enough numbers (in my testing 5 digits). Instead of using a tolerance, I propose to cap the values to be non overlapping in the first place. This works for all scenarios (very small to very big float values) in my manual testing.

## Proposed Changes

  - Remove the tolerance for comparison
  - Instead coerce the new range values to be non overlapping: 
    - when dragging the start handle, the value is forced to be <= the end value 
    - when dragging the end handle, the value is force to be >= the start value

## Testing

Test: Manual testing on different devices

## Issues Fixed

Fixes: 324934900
Fixes: 368982579

### DIFF
--- a/compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/Slider.kt
+++ b/compose/material3/material3/src/commonMain/kotlin/androidx/compose/material3/Slider.kt
@@ -2323,7 +2323,6 @@ private val ThumbSize = DpSize(ThumbWidth, ThumbHeight)
 private val VerticalThumbSize = DpSize(ThumbHeight, ThumbWidth)
 private val ThumbTrackGapSize: Dp = SliderTokens.ActiveHandleLeadingSpace
 private val TrackInsideCornerSize: Dp = 2.dp
-private const val SliderRangeTolerance = 0.0001
 
 private enum class SliderComponents {
     THUMB,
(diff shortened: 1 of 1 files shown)


romainguy:
Thank you. This is indeed a better approach that what the original code does since it doesn't take into account the loss of precision across float ranges. Another solution would be to use `endInclusive.ulp` instead of a constant tolerance value.

dlam (approved):
Approving to mirror this change to Gerrit for review by Material folks. (They do not monitor GitHub).

ln-12:
Unfortunately the first version of this PR was not a 100% fix. I added it to our production code by forking the Slider component but now we got one more crash in the next step (also extremely rare as we have ~15k users every day that can use this component):

```
java.lang.IllegalArgumentException: start(2743.6802) must be <= endInclusive(2743.68)
    at com.example.composables.SliderKt.SliderRange(Slider.kt:0)
    at com.example.composables.SliderKt.SliderRange(Slider.kt:2225)
    at com.example.composables.RangeSliderState.onDrag(Slider.kt:0)
    at com.example.composables.SliderKt.scale-P3rRKOc
    at com.example.composables.SliderKt.access$scale-P3rRKOc
    at com.example.composables.RangeSliderState.scaleToUserValue-lF5lt3s(Slider.kt:2130)
    at com.example.composables.RangeSliderState.onDrag(Slider.kt:2094)
    ...
```

I updated the `scale` function to also apply the correction for rounding issues. I will test it in our app and report back if this now solves all cases.

dlam:
Thank you for this contribution - if possible it would be great if you can add a test to `SliderTest.kt`, so we don't regress on this in the future.

ln-12:
Sorry, I didn't find the time to add a test 🫤 But at least I can confirm that we had no more crashes in our app regarding the Slider components after my last changes.

dlam:
No problem - we mirror PRs into Gerrit and then merge them there which closes the PR.

Your PR has been merged. Thank you for your contribution :)

PR #780, opened 2025-06-03
Adapt build scan publication message parsing

alextu:
## Proposed Changes

  - Develocity Gradle Plugin version 4.0.2 outputs a slightly different Build Scan publication log message: `Publishing Build Scan...` instead of `Publishing build scan...`. This PRs adds the new message to the ignore list.

## Testing

Test: Ran `development/build_log_simplifier.py --validate /tmp/gradle.log` with `gradle.log` containing `Publishing Build Scan...` and checked that it's properly stripped.


liutikas:
Thanks for the PR, I'll land it directly at source as part of r.android.com/3675009 that upgrades us to the latest Develocity plugin.

PR #781, opened 2025-06-04
Fix tokens for extraSmallContainerSize() in IconButtonDefaults

rengwuxian:
## Proposed Changes

  - Change the accidentally written `XSmallIconButtonTokens.DefaultLeadingSpace` to `XSmallIconButtonTokens.DefaultTrailingSpace`. For now they're the same value, which follows Material Design 3 Expressive, so no bugs for now, even if this isn't fixed. This is just for logical correctness, and future safety.

## Testing

Test: It's a simple fix, so it's easy to test: just make sure no typo out there.


liutikas:
it looks like we made the same mistake across the other *ContainerSize functions (smallContainerSize, mediumContainerSize, largeContainerSize, extraLargeContainerSize. Could you fix those too?

rengwuxian:
> it looks like we made the same mistake across the other *ContainerSize functions (smallContainerSize, mediumContainerSize, largeContainerSize, extraLargeContainerSize. Could you fix those too?

Thanks for pointing this out. Added those in the latest *amended commit*: 8fdc916 .

PR #789, opened 2025-07-11
Fix formatting issues in KDoc of `Arrangement.kt`

SimonMarquis:
## Proposed Changes

Fix the broken markdown link in KDoc that was incorrectly formatted on 2 lines.

| Info tooltip | In-editor rendering |
|---|---|
| image | image |

## Testing

Test: not applicable


SimonMarquis:
CI checks errors are reporting unrelated issues on `:compose:material3:material3-adaptive-navigation-suite`

liutikas (approved):

SimonMarquis:
@liutikas I've found 315 matches across 88 files with the same issue.
Would you like me to open a new review to fix them all in a batch or should I update this one?

Also, it would be great if kmft would not break these by default. Do you happen to have a specific configuration for it?

SimonMarquis:
It seems like you are already aware of it :)

- https://github.com/facebook/ktfmt/issues/240

Still, let me know which solution you prefer for fixing the existing occurrences.

PR #790, opened 2025-07-11
Fix broken `teardown` job in `integration_tests.yml` workflow

SimonMarquis:
## Proposed Changes

Add missing `$` character when setting the result variable in the `workflow-status` step.
This bug is the main reason for all these failures: https://github.com/androidx/androidx/actions/workflows/integration_tests.yml

## Testing

Test: Run the new workflow and check that the following error is no longer present:

```
Error: Unable to deliver Web Hook SyntaxError: Unexpected token ',', ..."success": , "src" : "... is not valid JSON
```

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -50,7 +50,7 @@ jobs:
           then
             echo "result=true" >> $GITHUB_OUTPUT
           else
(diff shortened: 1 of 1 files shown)


dlam (approved):

dlam:
Thanks for submitting this fix!

PR #792, opened 2025-07-30
 Prevent memory leaks in DialogFragment by cleaning up Handler callbacks in onDestroyView().

WeDoX:
DialogFragment may leak if Handler posts Runnable tasks that are not cleaned up on dismissal.

Override onDestroyView() to remove all  Handler callbacks.

## Proposed Changes
`
public void onDestroyView() {

super.onDestroyView();

.....

if (mHandler != null) {

mHandler.removeCallbacksAndMessages(null);

}

}
`

## Testing

1. Manually tested with device rotation and Android Profiler to confirm no leaks.
2. Verified using LeakCanary in debug builds.
3. Added unit test for Handler cleanup logic.
4. Confirmed null-safety with mock Handler objects.

### DIFF
--- a/fragment/fragment/src/main/java/androidx/fragment/app/DialogFragment.java
+++ b/fragment/fragment/src/main/java/androidx/fragment/app/DialogFragment.java
@@ -1046,5 +1046,9 @@ public class DialogFragment extends Fragment
             mDialog = null;
             mDialogCreated = false;
         }
-    }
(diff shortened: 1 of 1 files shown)


jbw0033:
Can you please add tests to the change?

It would also be nice if you could point us to a sample project to add more context to this change.

PR #795, opened 2025-08-13
[Core] Remove some outdated API wrappers

MGaetan89:
## Proposed Changes

This PR removes some API wrappers that are no longer needed. I've only inlined those that are just a couple of lines long. There are still others that could be inlined.

## Testing

Test: The GitHub checks for the `Core` module is green.


MGaetan89:
Note: the build for `core` is now failing, but it's the same issue on the `androidx-main` branch.

liutikas (approved):

PR #793, opened 2025-08-09
[Core] Remove outdated SDK references

MGaetan89:
## Proposed Changes

The min SDK version has moved to 23. This PR removes some outdated checks and internal/private support APIs.

## Testing

Test: The `AndroidX Presubmits / build-modules (ubuntu-latest, core)` check on GitHub passes.

PR #797, opened 2025-08-15
[Core] Deprecate proxy methods

MGaetan89:
## Proposed Changes

  - Deprecate proxy compat methods that simply call to the framework implementation.
  - Replace usages of the deprecated methods.
  - Add missing `@ReplaceWith` annotation.
  - Run `./gradlew updateApi`.

> [!NOTE]
> This PR focuses only on the files impacted by #793 and #795.

## Testing

Test: GitHub CI

### DIFF
--- a/core/core/src/main/java/androidx/core/view/contentcapture/ContentCaptureSessionCompat.java
+++ b/core/core/src/main/java/androidx/core/view/contentcapture/ContentCaptureSessionCompat.java
@@ -126,7 +126,9 @@ public class ContentCaptureSessionCompat {
      * @param virtualId id of the virtual child, relative to the parent.
      *
      * @return a new {@link ViewStructure} that can be used for Content Capture purposes.
+     * @deprecated
      */
+    @Deprecated
     public @Nullable ViewStructureCompat newVirtualViewStructure(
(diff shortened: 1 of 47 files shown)


core/core/src/main/java/androidx/core/view/contentcapture/ContentCaptureSessionCompat.java
```diff
@@ -126,7 +126,9 @@ private ContentCaptureSessionCompat(@NonNull ContentCaptureSession contentCaptur
      * @param virtualId id of the virtual child, relative to the parent.
      *
      * @return a new {@link ViewStructure} that can be used for Content Capture purposes.
+     * @deprecated
      */
+    @Deprecated
```
MGaetan89, line 131: I wasn't sure how to proceed here.
`ViewStructureCompat` is now deprecated, but changing the return typo would be a breaking change.
Probably a new method should be added, but what name should be used for it?

MGaetan89:
The failures in this PR are the same as on the `androidx-main` branch.

MGaetan89:
I've updated my PR to resolve the conflicts, and integrate the latest changes from the base branch.

PR #802, opened 2025-08-22
compose/ui-graphics: Fix NPE when creating ImageBitmap from invalid ByteArray

y9maly:
## Proposed Changes

Previously, `createImageBitmap` would pass null to the `AndroidImageBitmap` constructor if `BitmapFactory.decodeByteArray` returned null, leading to a NullPointerException later.

This change makes `createImageBitmap` throw an `IllegalArgumentException` immediately at the call site with a clear message:
"Failed to decode ImageBitmap from ByteArray: invalid or unsupported format"

According to the `BitmapFactory.decodeByteArray` documentation:
> Returns: The decoded bitmap, or null if the image could not be decoded.

So when it returns null, the input ByteArray is invalid or unsupported.

## Testing

Test: Tested manually.

```kotlin
fun doSmth(imageBitmap: ImageBitmap) {
    println(imageBitmap.height) // Previously: NullPointerException here
}

// Previously: No exception here
// Now: IllegalArgumentException here
val imageBitmap = ByteArray(0).decodeToImageBitmap()

doSmth(imageBitmap)
```


MohammedKHC:
Why is not this merged??

PR #803, opened 2025-08-27
[Activity] Remove outdated SDK checks

MGaetan89:
## Proposed Changes

This PR removes some outdated SDK version checks in `activity` following the bump of the min SDK version to 23.

## Testing

Test: The `AndroidX Presubmits / build-modules (ubuntu-latest, activity)` check on GitHub provides the same result as on `androidx-main`.


activity/activity/src/main/java/androidx/activity/EdgeToEdge.kt
```diff
@@ -93,9 +93,9 @@ fun ComponentActivity.enableEdgeToEdge(
                 EdgeToEdgeApi28()
             } else if (Build.VERSION.SDK_INT >= 26) {
                 EdgeToEdgeApi26()
-            } else if (Build.VERSION.SDK_INT >= 23) {
+            } else {
                 EdgeToEdgeApi23()
-            } else EdgeToEdgeApi21().also { Impl = it }
+            }.also { Impl = it }
```
MGaetan89: There's a small change here: previously, `Impl` would only be set for API 21-22. Now it is set on every API level.

activity/activity/src/main/java/androidx/activity/EdgeToEdge.kt
```diff
@@ -218,43 +218,13 @@ private interface EdgeToEdgeImpl {
     fun adjustLayoutInDisplayCutoutMode(window: Window)
 }
 
-private open class EdgeToEdgeBase : EdgeToEdgeImpl {
-
-    override fun setUp(
```
MGaetan89, line 223: Since edge-to-edge exists starting from API 21 (as per the comment on line 231), I've removed this default implementation to ensure that subclasses implement it.

liutikas (approved):

liutikas:
Could you please undo all the changes outside of activity/ ? I prefer we keep this change targeted just to this one set of libraries.

MGaetan89:
> Could you please undo all the changes outside of activity/ ? I prefer we keep this change targeted just to this one set of libraries.

Of course, I've updated the PR to keep only the changes that are in `activity/`

liutikas:
It looks like it is failing formatting. Please run `ktFormat` on activity projects.

MGaetan89:
I've reformatted the `EdgeToEdge` file

activity/activity/src/main/java/androidx/activity/EdgeToEdge.kt
```diff
@@ -93,9 +93,10 @@ fun ComponentActivity.enableEdgeToEdge(
                 EdgeToEdgeApi28()
             } else if (Build.VERSION.SDK_INT >= 26) {
                 EdgeToEdgeApi26()
-            } else if (Build.VERSION.SDK_INT >= 23) {
+            } else {
                 EdgeToEdgeApi23()
-            } else EdgeToEdgeApi21().also { Impl = it }
+            }
+    Impl = impl
```
MGaetan89, line 99: I've extracted the `Impl` assignment, otherwise ktlint reformats the `if` chain in a way that `also` only applied to the last element

PR #804, opened 2025-08-27
[AppCompat] Remove outdated SDK checks

MGaetan89:
## Proposed Changes

This PR removes some outdated SDK version checks in `appcompat` following the bump of the min SDK version to 23.

## Testing

Test: The `AndroidX Presubmits / build-modules (ubuntu-latest, appcompat)` check on GitHub provides is successful.


appcompat/appcompat-resources/src/main/java/androidx/appcompat/widget/VectorEnabledTintResources.java
```diff
@@ -38,16 +37,6 @@
 public class VectorEnabledTintResources extends ResourcesWrapper {
     private static boolean sCompatVectorFromResourcesEnabled = false;
```
MGaetan89, line 38: Probably this can be removed too (and the corresponding methods using too), but I wasn't sure if it was ok to go that far.

MGaetan89:
Should I keep the `TintContextWrapper` class even if it is no longer needed?

appcompat/appcompat-resources/api/restricted_current.txt
liutikas, line 1: None of the APIs tracked by this file are allowed to be removed, please add all of them back. We need that for backwards compatibility reasons as other older libraries in AndroidX might still be using them and we don't want them to crash at runtime

MGaetan89: I'll also revert the changes in `recyclerview` and `tracing` to keep this PR focused on `appcompat`.

MGaetan89: Done.
I've restored `TintContextWrapper`, but changed the `wrap` method to return the provided `Context` as-is.
The specific logic in that class is no longer used, since no instance of `TintContextWrapper` is ever created, but I kept it just in case. Let me know if I can remove it too.

MGaetan89: Actually, ErrorProne complains that `TintContextWrapper`'s constructor is unused. So I've removed the custom logic from that class and suppressed the warning.

MGaetan89:
@liutikas it looks like most modules are no longer building on GitHub.

appcompat/appcompat-resources/src/main/java/androidx/appcompat/widget/VectorEnabledTintResources.java
liutikas: Let's do it in a follow up PR.

liutikas (approved):

PR #814, opened 2025-09-20
[SQLite] Remove outdated code

MGaetan89:
## Proposed Changes

This PR removes some outdated SDK version checks in `sqlite` following the bump of the min SDK version to 23.

## Testing

Test: The `AndroidX Presubmits / build-modules (ubuntu-latest, sqlite)` check on GitHub is successful.


sqlite/sqlite/src/androidMain/kotlin/androidx/sqlite/db/SupportSQLiteCompat.android.kt
```diff
@@ -91,6 +91,9 @@ public class SupportSQLiteCompat private constructor() {
     }
 
     /** Helper for accessing functions that require SDK version 21 and higher. */
+    @Deprecated(
```
liutikas: This can be removed as it is library_group scope and this library is an atomic group 

https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:libraryversions.toml;l=307

Same with the one below

MGaetan89: Done 👍🏻
What does it imply that it is an atomic group?

liutikas: It means that all libraries in this maven group are expected to be used at the same version and we enforce that using Gradle constraints.

You can read a bit more about it in `Same-version (atomic) groups` section in 
https://cs.android.com/androidx/platform/frameworks/support/+/androidx-main:docs/api_guidelines/modules.md

liutikas (approved):

PR #828, opened 2025-11-18
Fix ComposableNamingDetector crash on internal functions

SimonMarquis:
`Composable`s with internal modifiers inside a class/interface/object will crash the `ComposableNamingDetector`.

```kotlin
object Foo {
    @Composable
    internal fun Content(): @Composable () -> Unit = {}
}
```

The computed `UMethod#name` for this will be `Content$impl`, which is then used in the `LintFix` to replace a String that does not exist in the source code. Leading to this crash:

```
Couldn't compute fix edits for Change to content$impl
java.lang.IllegalArgumentException: Did not find "Content$impl" in "Content" in [...]
Consider calling ReplaceStringBuilder#range() to set a larger range to search than the default highlight range.
```

## Proposed Changes

Use `UElement#nameFromSource` instead

## Testing

Test: added a `nestedAndInternalModifierIsSupported()` test case

## Issues Fixed

Issue: https://issuetracker.google.com/issues/461736991

### DIFF
--- a/compose/runtime/runtime-lint/src/main/java/androidx/compose/runtime/lint/ComposableNamingDetector.kt
+++ b/compose/runtime/runtime-lint/src/main/java/androidx/compose/runtime/lint/ComposableNamingDetector.kt
@@ -56,7 +56,7 @@ class ComposableNamingDetector : Detector(), SourceCodeScanner {
                 // special case where a generic return type and a Unit type parameter is used.
                 if (node.findSuperMethods().isNotEmpty()) return
 
-                val name = node.nameFromSource
+                val name = node.nameFromSource ?: node.name
 
                 val capitalizedFunctionName = name.first().isUpperCase()
 
--- a/compose/runtime/runtime-lint/src/main/java/androidx/compose/runtime/lint/ComposableNamingDetector.kt
+++ b/compose/runtime/runtime-lint/src/main/java/androidx/compose/runtime/lint/ComposableNamingDetector.kt
@@ -30,7 +30,7 @@ import com.android.tools.lint.detector.api.LintFix
 import com.android.tools.lint.detector.api.Scope
 import com.android.tools.lint.detector.api.Severity
 import com.android.tools.lint.detector.api.SourceCodeScanner
-import com.android.tools.lint.detector.api.nameFromSource
+import com.intellij.psi.PsiNamedElement
 import java.util.EnumSet
 import java.util.Locale
 import org.jetbrains.uast.UMethod
@@ -56,7 +56,7 @@ class ComposableNamingDetector : Detector(), SourceCodeScanner {
                 // special case where a generic return type and a Unit type parameter is used.
                 if (node.findSuperMethods().isNotEmpty()) return
 
-                val name = node.nameFromSource ?: node.name
+                val name = (node.sourcePsi as? PsiNamedElement)?.name ?: node.name
 
                 val capitalizedFunctionName = name.first().isUpperCase()
 
--- a/compose/runtime/runtime-lint/src/main/java/androidx/compose/runtime/lint/ComposableNamingDetector.kt
+++ b/compose/runtime/runtime-lint/src/main/java/androidx/compose/runtime/lint/ComposableNamingDetector.kt
@@ -56,6 +56,7 @@ class ComposableNamingDetector : Detector(), SourceCodeScanner {
                 // special case where a generic return type and a Unit type parameter is used.
                 if (node.findSuperMethods().isNotEmpty()) return
 
+                // NOTE: this is the inlined version of `UElement#nameFromSource` (available starting with Lint `31.10.0`)
                 val name = (node.sourcePsi as? PsiNamedElement)?.name ?: node.name
 
                 val capitalizedFunctionName = name.first().isUpperCase()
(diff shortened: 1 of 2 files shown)


liutikas (approved):

SimonMarquis:
~FYI there are still todo left in the expected output and expected diff. I was not able to set up the project yet, so if you can update them I'd be grateful.~

Copy/pasted the detector in a local project 👍

SimonMarquis:
If you want to see the test fail, you can simply comment

```kotlin
val name = /*node.nameFromSource ?:*/ node.name
```

and you'll get the error:

```
Did not find "button$lint_module" in "button" as suggested in the quickfix for issue ComposableNaming (text in range was "button")
java.lang.AssertionError: Did not find "button$lint_module" in "button" as suggested in the quickfix for issue ComposableNaming (text in range was "button")
	at org.junit.Assert.fail(Assert.java:89)
	at com.android.tools.lint.checks.infrastructure.TestLintClient.checkFix(TestLintClient.java:1374)
```

SimonMarquis:
Let me know if we should specifically document the name mangling that happens on internal identifiers in the detector.

andrewbailey:
Thanks for the contribution! It's mirrored in gerrit as [aosp/3857449](http://r.android.com/3857449).

This change is currently failing in presubmit with the following error. Can you take a look?
```
Buildbot Version: androidbuild_buildbot_20251112.00_RC00
Machine: r-c9274d16ff113e33-3-8350
DIST_DIR=/buildbot/dist_dirs/aosp-androidx-main-linux-androidx_host_tests/P109802297 CHANGE_INFO=/tmp/P109802297-changeInfo frameworks/support/busytown/androidx_host_tests.sh

############################################################################
Attempting to locate the relevant error messages via build_log_simplifier.py
############################################################################

Detected these failing tasks: [':compose:runtime:runtime-lint:compileKotlin']
> Task :compose:runtime:runtime-lint:compileKotlin
e: file://$SUPPORT/compose/runtime/runtime-lint/src/main/java/androidx/compose/runtime/lint/ComposableNamingDetector.kt:33:44 Unresolved reference 'nameFromSource'.
e: file://$SUPPORT/compose/runtime/runtime-lint/src/main/java/androidx/compose/runtime/lint/ComposableNamingDetector.kt:59:33 Unresolved reference 'nameFromSource'.

> Task :collection:collection:zipXmlResultsOfJsBrowserTest SKIPPED

Execution failed for task ':compose:runtime:runtime-lint:compileKotlin'.

* Exception is:
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':compose:runtime:runtime-lint:compileKotlin'.
Caused by: org.gradle.workers.internal.DefaultWorkerExecutor$WorkExecutionException: A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleCompilerRunnerWithWorkers$GradleKotlinCompilerWorkAction
Caused by: org.jetbrains.kotlin.gradle.tasks.CompilationErrorException: Compilation error. See log for more details
	at org.jetbrains.kotlin.gradle.tasks.TasksUtilsKt.throwExceptionIfCompilationFailed(tasksUtils.kt:21)...

BUILD FAILED in 7m 17s

Gradle command failed:
    ./gradlew --ci test allHostTests zipOwnersFiles createModuleInfo -Pandroidx.displayTestOutput=false --continue
```

When you're ready, you can push your changes to this PR. It'll mirror into gerrit automatically and then we can trigger another presubmit run for you.

SimonMarquis:
Oh you're right, `UElement.nameFromSource` was added in `31.10.0` but androidx is currently pinned on `31.8.2`
https://github.com/androidx/androidx/blob/bd9ced3b70ba7c9b555bd9e1ccd0901de2e25173/gradle/libs.versions.toml#L15-L19

I'll inline the call instead.

compose/runtime/runtime-lint/src/main/java/androidx/compose/runtime/lint/ComposableNamingDetector.kt
```diff
@@ -55,7 +56,7 @@ class ComposableNamingDetector : Detector(), SourceCodeScanner {
                 // special case where a generic return type and a Unit type parameter is used.
                 if (node.findSuperMethods().isNotEmpty()) return
 
-                val name = node.name
+                val name = (node.sourcePsi as? PsiNamedElement)?.name ?: node.name
```
andrewbailey, line 61: Can you please add a comment above the inlined implementation here to explain that it's synonymous to `nameFromSource`? We may consider switching back to the original API once we can upgrade Lint.

We also probably want to have that reference in case we need to take any changes from the upstream implementation.

SimonMarquis: Would this be OK? Let me know if you have a better way of linking this.

```suggestion
                // NOTE: this is the inlined version of `UElement#nameFromSource` (available starting with Lint `31.10.0`)
                val name = (node.sourcePsi as? PsiNamedElement)?.name ?: node.name
```

andrewbailey: Yes, that looks good to me — anything that points to where this came from is fine.

SimonMarquis:
I'll try to take care of the ktlint issue.

SimonMarquis:
No luck unfortunately. @andrewbailey I'd love some help here. Would you mind running the command: `./gradlew :compose:runtime:runtime-lint:ktFormat`?

PR #830, opened 2025-11-21
Detect broad configuration inputs from `System.getenv()` and `System.getProperties()`

joshfriend:
## Proposed Changes

Add new issue ID for overly broad capture of configuration cache inputs based on the [Gradle docs](https://docs.gradle.org/current/userguide/configuration_cache_requirements.html#config_cache:requirements:reading_sys_props_and_env_vars)

I had to rework most of the bad method detection because the existing implementation assumes that _any_ occurrence of a given method should be blocked regardless of arity. `System.getenv()` is a new case where the no-arg version is bad but `System.getenv(String)` is fine. Those methods are also static and the existing code was not resolving them.

## Testing

Test: Added new integration tests for the new lint issue. All existing tests pass with no additional changes.

## Issues Fixed

Fixes: https://issuetracker.google.com/issues/462234633

### DIFF
--- a/lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
+++ b/lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
@@ -29,6 +29,7 @@ import com.intellij.psi.PsiClass
 import com.intellij.psi.PsiClassType
 import org.jetbrains.kotlin.psi.KtSimpleNameStringTemplateEntry
 import org.jetbrains.kotlin.psi.KtStringTemplateEntry
+import org.jetbrains.uast.UBinaryExpression
 import org.jetbrains.uast.UCallExpression
 import org.jetbrains.uast.UElement
 import org.jetbrains.uast.UExpression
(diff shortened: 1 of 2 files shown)


lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
```diff
@@ -242,90 +298,99 @@ class DiscouragedGradleMethodDetector : Detector(), Detector.UastScanner {
 
         // A map from eager method name to the containing class of the method and the name of the
         // replacement method, if there is a direct equivalent.
-        private val REPLACEMENTS =
+        private val REPLACEMENTS: Map<String, Replacement> =
```
liutikas, line 301: The fact that we now combine the fully qualified class name, the method name, and optionally argument into a single thing makes the RELEVANT_METHOD_NAMES parameter construction fragile.

I think we should move to using a real class here to reduce changes of mess ups / typos, etc

joshfriend: You mean something like this for the key?

```kotlin
data class SuspectMethod(
  val klass: Class,
  val method: ???,
  val arguments: List<Type>? = null,
)
```

liutikas: I've been spinning on a proper solution for this for an hour now :|

Ideally, when we do checking, we can go from relevant method to required parameter types to isInstanceOf for the relevant class. That would be best perf wise.

So maybe we want to go with Map< Map< Tuple<method name, requiredParams>, classname, Replacement>

WDYT?

moseyj19-design (commented):
Leave as is until I decide to assign people who don't steal my accounts

lint/lint-gradle/src/test/java/androidx/lint/gradle/ConfigurationCacheBroadInputsTest.kt
```diff
@@ -0,0 +1,113 @@
...
+                fun readSetup() {
+                    val ciEnv = System.getenv()
+                }
+            """
+                    .trimIndent()
+            )
+
+        val expected =
+            """
+            src/test.kt:4: Error: Avoid using method getenv [GradleConfigurationCacheInputs]
```
juliamcclellan, line 46: Does this message make sense, given that a different method with the same name is allowed? It seems like it would make sense to include the "()"

lint/lint-gradle/src/test/java/androidx/lint/gradle/ConfigurationCacheBroadInputsTest.kt
```diff
@@ -0,0 +1,113 @@
...
+                """
+                import java.lang.System
+
+                fun readSetup() {
+                    val ciEnv = System.getenv("CI")
+                }
+            """
+                    .trimIndent()
+            )
+        check(input).expect("No warnings.")
```
juliamcclellan, line 69: This could use `expectClean()`.

lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
```diff
@@ -211,6 +252,21 @@ class DiscouragedGradleMethodDetector : Detector(), Detector.UastScanner {
                 Implementation(DiscouragedGradleMethodDetector::class.java, Scope.JAVA_FILE_SCOPE),
             )
 
+        val CONFIGURATION_CACHE_BROAD_INPUTS =
+            Issue.create(
+                "GradleConfigurationCacheInputs",
+                "Avoid using APIs that capture too much of the environment in the" +
```
juliamcclellan, line 258: There should be a space either before the closing " on this line or after the opening " on the next.

lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
juliamcclellan: What about something like `Map<String (method name), List<MethodDetails>>` with
```
data class MethodDetails(
    val containingClass: String,
    val requiredParams: List<String>? = null,
    val replacement: Replacement
)
```
The implementation could look mostly like the previous one, with an additional line in the `potentialReplacements.forEach` which checked the parameters.

lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
```diff
@@ -47,41 +48,78 @@ class DiscouragedGradleMethodDetector : Detector(), Detector.UastScanner {
...
+                if (node.methodName !in RELEVANT_METHOD_NAMES) return
+
+                val method = node.resolve() ?: return
+                val containingClass =
+                    (node.receiver?.getExpressionType() as? PsiClassType)?.resolve()
+                        ?: method.containingClass
+                        ?: return
+                val methodName = method.name
+                val paramSig =
+                    method.parameterList.parameters.joinToString(",") { it.type.canonicalText }
```
juliamcclellan, line 61: One thing I'm concerned about for the parameter type strings is how they would work with generics, e.g. if we had
```
interface Interface<T> {
    fun someMethod(t: T)
}
class StringClass: Interface<String> {
    override fun someMethod(t: String) = Unit
}
```
and wanted to ban `Interface.someMethod(T)`. When this is used as `StringClass.someMethod(String)`, then I don't think matching against the parameter strings would work, and might require looking at the super methods to fully support that case.

liutikas:
Now that i've thought about this for a while, do we even want people to use System.getenv even for the parameter version? It seems to be strictly better to push folks to use `provider.environmentVariable` all the time.

WDYT?

liutikas:
Alternative implementation r.android.com/3991273

joshfriend:
> Alternative implementation r.android.com/3991273

thanks, i have not had time to pursue this PR any further

PR #835, opened 2025-12-11
fix: package typo in benchmark

Sunbreak:
## Proposed Changes

  -
  -
  -

## Testing

## Issues Fixed


liutikas:
This will break our benchmark result tracking as the results will move to a different package. We prefer to keep it as is.

PR #841, opened 2026-01-06
Fix race condition in LimitOffsetPagingSource

rrhett:
## Proposed Changes

Fix two race conditions in LimitOffsetPagingSource:
  - Move the refreshComplete.compareAndSet to before the initialLoad so invalidations that happen while the initial read are occurring are handled.
  - Add a signal from the invalidation callback that the initial read waits on in order to ensure the read doesn't occur until after the invalidation flow is being collected.

## Testing

Test: Run LimitOffsetPagingSourceTest and LimitOffsetPagingSourceTestWithFilteringCoroutineDispatcher

## Issues Fixed

Fixes: The bug on [https://issuetracker.google.com/issues/473587154](https://issuetracker.google.com/issues/473587154) is fixed.

This may also address https://issuetracker.google.com/issues/192269858 but it wasn't clear to me if that bug was still relevant or earlier changes addressed it satisfactorily, so I opened a new bug.

Additionally, due to this being a race condition, it is difficult to robustly test in code. I have a repro app. If I can get guidance on building my repro app against a locally built room-paging library, I'm happy to hit it manually a bit too. The repro is not 100% but not too hard to hit with the repro app.

### DIFF
--- a/room3/room3-paging/src/commonMain/kotlin/androidx/room3/paging/LimitOffsetPagingSource.kt
+++ b/room3/room3-paging/src/commonMain/kotlin/androidx/room3/paging/LimitOffsetPagingSource.kt
@@ -31,6 +31,7 @@ import androidx.room3.paging.util.queryContext
 import androidx.room3.paging.util.queryDatabase
 import androidx.room3.paging.util.queryItemCount
 import kotlinx.coroutines.CancellationException
+import kotlinx.coroutines.CompletableDeferred
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
@@ -74,6 +75,7 @@ internal class CommonLimitOffsetImpl<Value : Any>(
     internal val itemCount = AtomicInt(INITIAL_ITEM_COUNT)
 
     private val refreshComplete = AtomicBoolean(false)
+    private val collectorLatch = CompletableDeferred<Unit>()
 
     private var invalidationFlowJob: Job? = null
 
@@ -81,9 +83,16 @@ internal class CommonLimitOffsetImpl<Value : Any>(
         // register db listeners right away
         invalidationFlowJob =
             db.getCoroutineScope().launch {
-                db.invalidationTracker.createFlow(*tables, emitInitialState = false).collect {
+                db.invalidationTracker.createFlow(*tables, emitInitialState = true).collect {
+                    // Signal even before checking if we're still valid, to ensure the load function
+                    // doesn't block.
+                    val initialState = collectorLatch.complete(Unit)
                     if (pagingSource.invalid) {
                         throw CancellationException("PagingSource is invalid")
+                    }
+                    if (initialState) {
+                        // The first emit essentially just lets us know the flow has started.
+                        return@collect
                     }
                     if (refreshComplete.get()) {
                         pagingSource.invalidate()
@@ -96,10 +105,16 @@ internal class CommonLimitOffsetImpl<Value : Any>(
 
     suspend fun load(params: LoadParams<Int>): LoadResult<Int, Value> {
         val tempCount = itemCount.get()
-        // if itemCount is < 0, then it is initial load
+        // if itemCount is == INITIAL_ITEM_COUNT, then it is initial load
         return try {
+            // Wait until we know the collector has started, which means we're guaranteed to get an
+            // invalidation for any updates that happen after this point.
+            collectorLatch.await()
             if (tempCount == INITIAL_ITEM_COUNT) {
-                initialLoad(params).also { refreshComplete.compareAndSet(false, true) }
+                // Mark the load started for invalidation purposes; once we pass this point, we
+                // want to ensure we process any invalidations by the flow collector.
+                refreshComplete.compareAndSet(false, true)
+                initialLoad(params)
             } else {
                 nonInitialLoad(params, tempCount)
             }
--- a/room3/room3-paging/src/androidDeviceTest/kotlin/androidx/room3/paging/LimitOffsetPagingSourceTest.kt
+++ b/room3/room3-paging/src/androidDeviceTest/kotlin/androidx/room3/paging/LimitOffsetPagingSourceTest.kt
@@ -555,9 +555,22 @@ class LimitOffsetPagingSourceTest {
     }
 
     @Test
+    fun delete_invalidatesPagerAfterRefresh() = runPagingSourceTest { pager, pagingSource ->
+        pager.refresh(initialKey = 10)
+        assertThat(pagingSource.invalid).isFalse()
+
+        dao.deleteTestItems(5, 15)
+        // Wait for the delete to finish.
+        countingTaskExecutorRule.drainTasks(500, TimeUnit.MILLISECONDS)
+        assertThat(pagingSource.invalid).isTrue()
+    }
+
+    @Test
     fun load_refreshKeyGreaterThanItemCount_lastPage() = runPagingSourceTest { pager, _ ->
         pager.refresh(initialKey = 70)
         dao.deleteTestItems(40, 100)
+        // Wait for the delete to finish.
+        countingTaskExecutorRule.drainTasks(500, TimeUnit.MILLISECONDS)
 
         // assume user was viewing last item of the refresh load with anchorPosition = 85,
         // initialLoadSize = 15. This mimics how getRefreshKey() calculates refresh key.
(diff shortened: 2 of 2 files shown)


room3/room3-paging/src/commonMain/kotlin/androidx/room3/paging/LimitOffsetPagingSource.kt
```diff
@@ -96,13 +107,23 @@ internal class CommonLimitOffsetImpl<Value : Any>(
...
+                    // Mark the load started for invalidation purposes; once we pass this point, we
+                    // want to ensure we process any invalidations by the flow collector.
+                    refreshComplete.compareAndSet(false, true)
+                    initialLoad(params)
+                } else {
+                    nonInitialLoad(params, tempCount)
+                }
+
+            @Suppress("UNCHECKED_CAST")
+            if (pagingSource.invalid) INVALID as LoadResult.Invalid<Int, Value> else loadResult
```
rrhett: It's unclear to me that this check is still useful. It is possible that the pagingSource was invalidated by the collector above by the time this line is reached; however, if so, whatever is using this pagingSource should handle that invalidation, and so returning stale data here doesn't seem particularly bad.

If desired, changing this to just return the if would simplify the code a bit, and I don't believe would be wrong at all, though might require a couple tests to change. Happy to do that or not.

room3/room3-paging/src/commonMain/kotlin/androidx/room3/paging/LimitOffsetPagingSource.kt
```diff
@@ -96,13 +107,23 @@ internal class CommonLimitOffsetImpl<Value : Any>(
...
+        // if itemCount is == INITIAL_ITEM_COUNT, then it is initial load
         return try {
-            if (tempCount == INITIAL_ITEM_COUNT) {
-                initialLoad(params).also { refreshComplete.compareAndSet(false, true) }
-            } else {
-                nonInitialLoad(params, tempCount)
-            }
+            // Wait until we know the collector has started, which means we're guaranteed to get an
+            // invalidation for any updates that happen after this point.
+            signal.first { it }
```
danysantiago: The more idiomatic way to do this is with a `CompletableDeferred` which will act as a 'latch'

```
val collectorLatch = CompletableDeferred<Unit>()

// init
invalidationTracker .createFlow(...).collect {
  collectorLatch.complete(Unit) // signal collect has started
}

// load
fun load(...) {
  collectorLatch.await() // waits for invalidation tracker collect coroutine to start
}
```

I think if you replace the `signal: MutableFlow` with this, then we can remove both `refreshComplete` and `collectStarted` since now there is a hard wait until invalidation is installed.

room3/room3-paging/src/commonMain/kotlin/androidx/room3/paging/LimitOffsetPagingSource.kt
danysantiago: This was added to prevent flickering and I think we should keep it to not confuse the pager as it realizes the source is invalid and a new one is needed.

room3/room3-paging/src/commonMain/kotlin/androidx/room3/paging/LimitOffsetPagingSource.kt
```diff
@@ -141,23 +162,12 @@ internal class CommonLimitOffsetImpl<Value : Any>(
...
-        val loadResult =
-            queryDatabase(
-                params = params,
-                sourceQuery = sourceQuery,
-                itemCount = tempCount,
-                convertRows = convertRows,
-            )
-        // TODO(b/192269858): Create a better API to facilitate source invalidation.
-        // Manually check if database has been updated. If so, invalidate the source and the result.
-        withContext(db.queryContext) {
```
danysantiago, line 153: We should keep this logic because once initial load happens and there is UI tied to this data we want to make sure not to return stale that for a paging source that will be invalidated as that causes flickering. So we have this not-so-great call to `invalidationTracker.refresh` that essentially emits invalidations synchronously  guaranteeing that the `loadResult` has good data to return.

danysantiago (changes requested):
Thanks for sending this pull request!

room3/room3-paging/src/commonMain/kotlin/androidx/room3/paging/LimitOffsetPagingSource.kt
rrhett: Thanks for the pointer to `CompletableDeferred`! I figured there was something better :)

We can replace `collectStarted` by using the return value of `collectorLatch.complete` but I think we still want that logic. If we get rid of the check entirely, we'll always invalidate on the first emit which always happens because I set `emitInitialState` to true.

I agree that we do not need `refreshComplete` from a semantic standpoint. The main reason I kept it is to minimize the changes in the CL and it will trigger more extensive changes to the tests, as many of them create the paging source, then update the DB (which will trigger an invalidation) and then use the paging source.

I'll make that change but I'm going to split it into two commits in my fork so you can see the impact of it separately (23 tests failed just removing `refreshComplete`).

For that change, I'll change `runPagingSourceTest` to take a `pagingSourceFactory` instead of a `pagingSource`, and have it create the paging source right before running the block.

room3/room3-paging/src/commonMain/kotlin/androidx/room3/paging/LimitOffsetPagingSource.kt
rrhett: Okay, I'll put that logic back. Based on that I think I can just return the if in `load` so I simplified that method. Let me know if that looks right.

I'm still not entirely convinced this is necessary, as it only reduces the chance of flicker. If the return happens fast enough while there's a write in flight, it can still make it to the UI, and then the invalidate from the tracker happens and causes flicker. Also, if for some reason we're in a case with heavy DB churn, we might essentially starve the UI of any updates until the DB has settled.

That said, it is distinct from the core purpose of this PR which is to fix the underlying race condition around data NOT making it to the UI, so I'm happy to shelve the conversation for now.

rrhett:
Thanks for the detailed review! I've addressed your comments with two commits. PTAL. Also let me know what your preferred way of handling updates is (should I amend, squash, etc) or if I get anything wrong there, I'll do my best to fix it up :)

room3/room3-paging/src/commonMain/kotlin/androidx/room3/paging/LimitOffsetPagingSource.kt
danysantiago: Thanks, SGTM to keep `refreshComplete` for now then and just focus on fixing the race plus less changes on tests, after all it is suppose to help avoiding that case where an invalidation happens when the initial load has not even occurred, causing the paging source to be thrown out immediately if attached to a pager.

room3/room3-paging/src/commonMain/kotlin/androidx/room3/paging/LimitOffsetPagingSource.kt
danysantiago: The `load` looks good. I also agree to keep the focus of the PR on fixing the race so feel free to resolve this conversation.

room3/room3-paging/src/commonMain/kotlin/androidx/room3/paging/LimitOffsetPagingSource.kt
rrhett: Okay, sounds good. So I have 3 commits right now on this PR:

022bd60: initial fix
b983fb4: fix the latch and bring back the synchrononous refresh - it sounds like this is where we want to be?
a5b94dd: remove refreshComplete

So are there any further changes requested / required that aren't captured in one of these? Do you need me to do anything with these commits? Let me know and I'll update my fork as needed. Thanks!

danysantiago: Can you please rebase, removing https://github.com/androidx/androidx/commit/a5b94ddb3251faba38c46654daa388f06834b84b (commit that removes `refreshComplete`) and squashing the other two commits? Once we approve this PR what happens is that a copy is made on the review tool used by Android (in https://android-review.googlesource.com/), you'll be CCed. That is where the big presubmit runs and if all looks good it gets merged through that system which then closes this PR. However, Gerrit works on a 'review per single commit' style, which is why I am asking for the rebase. Thanks!

rrhett: Alright, I believe I've done that properly, PTAL. I also undid a few formatting changes that had made it in through the back and forth, so I believe this is a pretty targeted change now.

danysantiago (approved):
LGTM

rrhett:
It looks like about half the presubmits failed. As near as I can tell, they all have errors that look like:

Project with path ':inspection:inspection' could not be found in project ':compose:ui:ui-inspection'.

This does not appear at first glance to be related to my code, but if you could take a look and let me know if there's anything I can do to fix it or otherwise move it along from my end, I would greatly appreciate it.

dlam:
I would ignore these - our presubmit on GH is quite a bit broken at the moment :(

This has been mirrored into AOSP after dany's review and if it fails presubmit there we'll report back with the error. Sorry about the confusion!

dlam:
And on that note - seems like it passed already :)

https://android-review.googlesource.com/c/platform/frameworks/support/+/3917946

PR #865, opened 2026-03-05
Add filterTouchesWhenObscured attribute to custom dialog

abhishek-900:
fix(android): Add filterTouchesWhenObscured to prevent Tapjacking

Proposed Changes

Added android:filterTouchesWhenObscured="true" to views in custom_dialog.xml.

Prevents touch input when the view is obscured by another window, mitigating overlay/tapjacking attacks (CWE-1021).

Testing

Test: Verified the feedback dialog ignores touch events when an overlay is present and works normally otherwise.

### DIFF
--- a/core/core/src/main/res/layout/custom_dialog.xml
+++ b/core/core/src/main/res/layout/custom_dialog.xml
@@ -1,7 +1,8 @@
 <?xml version="1.0" encoding="utf-8"?>
 <RelativeLayout xmlns:android="http://schemas.android.com/apk/res/android"
     android:layout_width="fill_parent"
-    android:layout_height="fill_parent">
+    android:layout_height="fill_parent"
+    android:filterTouchesWhenObscured="true">
 
     <Button
         android:id="@+id/dialog_button"


liutikas:
Is there an issue on the issue tracker that this is trying to address?

abhishek-900:
Thanks for the review.
This change was introduced to address a tapjacking vulnerability flagged during a security scan. The fix adds android:filterTouchesWhenObscured="true" to the dialog views to prevent touch events when the window is obscured by an overlay (CWE-1021).

There is currently no issue on the tracker for this. If required, I can create one and link it to this PR.

liutikas:
Please create an issue on issue tracker so the owner of the library can evaluate whether this is a real issue and if this is the right fix for it.

abhishek-900:
@liutikas  — As requested, I've filed an issue on the Issue Tracker:
https://issuetracker.google.com/issues/492924089

The fix is ready in this PR. Please let me know if you'd like me to reopen it
or submit a fresh PR once the issue is reviewed.

PR #871, opened 2026-03-30
Fix baseline profile plugin creating dirs at configuration time

joshfriend:
The baseline profile consumer plugin calls `mkdirs()` at configuration time when registering source directories, which creates `src/<variant>/generated/baselineProfiles` for every variant. This causes Gradle's configuration cache to track these directories as file system inputs, leading to unnecessary cache invalidations when the directories are later removed (e.g. by clean tasks or git operations).

The fix removes the `mkdirs()` call since Gradle already creates the directory at execution time via the `@OutputDirectory` annotation on the copy task's `baselineProfileDir` property.

Test: BaselineProfileConsumerPluginTest#testBaselineProfileOutputDirNotCreatedDuringConfiguration


joshfriend:
I'm pretty sure the `:sqlite:sqlite-web:jsBrowserTest'` failure is unrelated, but i cannot retry it

liutikas (approved):

PR #887, opened 2026-06-21
Fix PointerInteropFilter to treat non-touch generic motion events as …

TestPlanB:
# Dispatch ACTION_HOVER_* / ACTION_SCROLL as stateless events in PointerInteropFilter
Context: Non-touch generic motion events ( ACTION_HOVER_ENTER , ACTION_HOVER_MOVE , ACTION_HOVER_EXIT , ACTION_SCROLL , ACTION_BUTTON_PRESS / RELEASE ) are currently piped through the same DispatchToViewState state machine used for the touch stream. As a consequence, once any earlier event causes the filter to flip into NotDispatching (e.g. because Compose consumed a hover change in a later pass), all subsequent generic pointer events are silently dropped until the next ACTION_DOWN resets the state. Hover and scroll events are not part of an ongoing touch stream and should be dispatched as stateless events.

Goal of this change: Dispatch ACTION_HOVER_* and ACTION_SCROLL events to the underlying onTouchEvent handler during PointerEventPass.Initial , regardless of the current DispatchToViewState.NotDispatching state. Also avoid treating such events as part of the touch stream so that the interceptor cannot cause subsequent generic pointer events to be silently dropped.

## Proposed Changes
- PointerInteropFilter.android.kt — Introduce a generic-pointer-event fast path at the top of onPointerEvent :
 
  - New helper PointerEvent.isTouchStreamEvent() classifies events: only ACTION_DOWN / ACTION_POINTER_DOWN / ACTION_MOVE / ACTION_POINTER_UP / ACTION_UP / ACTION_CANCEL / ACTION_OUTSIDE (and null MotionEvents) are considered part of the touch stream. Everything else is dispatched as a stateless generic event.
  - New private method dispatchGenericPointerEvent(pointerEvent, pass) forwards the event to the View only during PointerEventPass.Initial , bails out if the changes are already consumed, and consumes the PointerInputChange s iff the underlying onTouchEvent returns true . The method deliberately never mutates state , never calls stopDispatching() , and never synthesizes an ACTION_CANCEL .
  - The existing DispatchToViewState.Unknown / Dispatching / NotDispatching touch-stream machine is unchanged for real touch events, preserving behavior for click / drag / long-press flows.
- PointerInteropFilter.android.kt — Guard the state mutation and stopDispatching() logic so it only runs for touch-stream events (implicitly handled by the early-return above), so a Compose consumer cannot poison the generic pointer pipeline by consuming a hover change.

## Testing
- PointerInteropFilterTest.kt — Added three regression @Test methods under // Verification of correct passes being used :
  
  - onPointerEvent_hoverDispatchedDuringInitialPassAfterTouchNotDispatching
  - onPointerEvent_scrollDispatchedDuringInitialPassAfterTouchNotDispatching
  - onPointerEvent_genericPointerEventNotDispatchedAfterInitialPass
  These tests first drive the filter into NotDispatching via a touch DOWN whose onTouchEvent returns false , and then verify that hover / scroll / generic events:
  
  1. still reach onTouchEvent during Initial ;
  2. are only dispatched once ( Initial ), never again in Main / Final ;
  3. have their PointerInputChange s consumed iff the View returns true .

## Issues Fixed
Fixes: https://issuetracker.google.com/issues/526074807

### DIFF
--- a/compose/ui/ui/src/androidDeviceTest/kotlin/androidx/compose/ui/input/pointer/PointerInteropFilterTest.kt
+++ b/compose/ui/ui/src/androidDeviceTest/kotlin/androidx/compose/ui/input/pointer/PointerInteropFilterTest.kt
@@ -19,9 +19,11 @@ package androidx.compose.ui.input.pointer
 import android.view.MotionEvent
 import android.view.MotionEvent.ACTION_CANCEL
 import android.view.MotionEvent.ACTION_DOWN
+import android.view.MotionEvent.ACTION_HOVER_MOVE
 import android.view.MotionEvent.ACTION_MOVE
 import android.view.MotionEvent.ACTION_POINTER_DOWN
 import android.view.MotionEvent.ACTION_POINTER_UP
+import android.view.MotionEvent.ACTION_SCROLL
 import android.view.MotionEvent.ACTION_UP
 import android.view.MotionEvent.TOOL_TYPE_UNKNOWN
 import androidx.compose.ui.Modifier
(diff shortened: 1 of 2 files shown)


TestPlanB:
@ShikaSD  Could you please help review this patch? Thanks a lot.

codingjeremy:
I don't understand what _"Fix `PointerInteropFilter` to treat non-touch generic motion events as stateless"_. 

Generally, you should be using`Modifier.pointerInput{}` for input events (not `PointerInteropFilter`). Pointer input already supports general generic `MotionEvent`s like mouse and scroll wheel. 

Can you outline what type of events you are trying to support and/or the specific use case? Also, when you say stateless, are you referring to the platform or Compose?

TestPlanB:
Thanks for your questions. @codingjeremy 

We use PointerInteropFilter inside AndroidView. Currently, the AndroidView cannot receive hover or scroll events.
The root cause is that changedToUpIgnoreConsumed returns false for these events, which prevents the reset() function from being triggered.

```
  if (pass == PointerEventPass.Final) {
                   // changedToUpIgnoreConsumed returns false for scroll/hover events
                    if (changes.fastAll { it.changedToUpIgnoreConsumed() }) {
                        reset()
                    }

                    if (pointerEvent == lastEventDispatchedToInitialPass && isMoveEvent) {
                        // we've reached the final pass, if the motion event that was sent
                        // during the initial pass was consumed, it means Compose claimed it
                        // so we should stop dispatching to the View
                        if (changes.fastAny { it.isConsumed } && !disallowIntercept) {
                            stopDispatching(pointerEvent)
                        } else {
                            changes.fastForEach { it.consume() }
                        }
                    }
                }
```
Demo reproduction code:
```
Box(modifier = Modifier.fillMaxSize().background(Color.Red)){
                    AndroidView({
                        val view =  View(it)
                        view.setBackgroundColor(androidx.compose.ui.R.color.vector_tint_theme_color)
                        view.setOnGenericMotionListener {
                            view,event ->
                            // events to fail to dispatch.  
                           Cannot receive ACTION_SCROLL and similar hover events when using mouse
                           Log.e("hello","event ")
                            false
                        }
                        view
                    }, modifier = Modifier.fillMaxSize())
                }
```

TestPlanB:
Mouse hover and scroll generic motion events fail to trigger reset(), as they never satisfy changedToUpIgnoreConsumed(). This leaves PointerInteropFilter stuck in a blocked dispatch state, which prevents the AndroidView generic motion listener from receiving events.

codingjeremy:
Hi @TestPlanB, thanks for looking into generic motion events for interop.

We want to discourage expanding `PointerInteropFilter`'s scope or adding complex logic to it. (It isn't something we even recommend using.)

### Recommended Alternative: Native Compose Pointer Input
For hover enter/exit/move and scroll events, Jetpack Compose provides first-class native support directly via `Modifier.pointerInput`. We strongly recommend utilizing this instead of legacy interop filters.

Here is a simple example of how to listen to these events natively:

```kotlin
Modifier.pointerInput(Unit) {
    awaitPointerEventScope {
        while (true) {
            val event = awaitPointerEvent()
            when (event.type) {
                PointerEventType.Enter -> {
                    // Hover entered the composable bounds
                }
                PointerEventType.Exit -> {
                    // Hover exited the composable bounds
                }
                PointerEventType.Move -> {
                    // Hover move event inside the bounds
                }
                PointerEventType.Scroll -> {
                    // Scroll / Mouse Wheel event
                }
            }
        }
    }
}
```

This native approach is clean, performant, avoids allocating new legacy MotionEvent objects on every hover/scroll event, and fully integrates with Compose's standard event propagation, bounds-clipping, and layout phases.

In addition, you can access the original `MotionEvent` from the pointer events, so, if you do need those for some reason, you can use that.

PR #893, opened 2026-07-17
Fix scattermap kdoc typo

Hussam-Mak:
## Proposed Changes

- Fix minor KDoc typos in `ScatterMap.kt`.
- Remove a duplicated `to` in the internal `hash()` documentation.
- Correct wording in the `hashCode()`, `trim()`, and `asMutableMap()` documentation.

## Testing

Test: `git diff --check origin/androidx-main..HEAD`

## Issues Fixed

Fixes: N/A

### DIFF
--- a/collection/collection/src/commonMain/kotlin/androidx/collection/ScatterMap.kt
+++ b/collection/collection/src/commonMain/kotlin/androidx/collection/ScatterMap.kt
@@ -368,7 +368,7 @@ public expect sealed class ScatterMap<K, V> {
     ): String
 
     /**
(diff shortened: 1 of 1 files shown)


Hussam-Mak:
﻿Rebased onto current androidx-main.

The original hash() KDoc "to to" fix is no longer needed — that comment was rewritten upstream. This PR now only keeps the three remaining KDoc typos in ScatterMap.kt ("the hash code the sum" -> "is the sum", "Returns be 0" -> "Returns 0", and "vice- versa" -> "vice versa").

@dlam would you be able to take a look when you have a chance? Docs-only, no behavior change.

PR #895, opened 2026-07-18
Add SavedState optimization for collections of nullable CharSequence/Parcelable

marco-schm:
## Proposed Changes

- Add four internal serial descriptors for `Array<CharSequence?>`, `List<CharSequence?>`, `Array<Parcelable?>` and `List<Parcelable?>` in `SavedStateCodecUtils.android.kt`, mirroring the existing `nullablePolymorphicSparseParcelableArrayDescriptor` pattern used for `SparseArray<Parcelable?>`
- Route the new descriptors to the existing built-in serializers (`CharSequenceArraySerializer`, `CharSequenceListSerializer`, `ParcelableArraySerializer`, `ParcelableListSerializer`) in `SavedStateEncoder.android.kt` and `SavedStateDecoder.android.kt`, so these types get the direct single-entry representation instead of falling back to the generic collection encoding
- Add round-trip test cases for all four types to `SavedStateCodecAndroidTest.canUseBuiltInSerializersAutomatically`, verifying the single-entry representation and null element preservation

The underlying `SavedState`/`Bundle` APIs already preserve null elements, so no serializer changes are needed. All added declarations are internal; there are no public API changes.

## Testing

Test: ./gradlew :savedstate:savedstate:testAndroidHostTest

Full savedstate host test suite passes (424 tests, 0 failures). The new test cases fail against the previous implementation (the collections were encoded generically as one entry per element) and pass with this change (single direct entry, nulls preserved).

## Issues Fixed

Fixes: b/400696665

### DIFF
--- a/savedstate/savedstate/src/androidHostTest/kotlin/androidx/savedstate/SavedStateCodecAndroidTest.android.kt
+++ b/savedstate/savedstate/src/androidHostTest/kotlin/androidx/savedstate/SavedStateCodecAndroidTest.android.kt
@@ -663,6 +663,26 @@ internal class SavedStateCodecAndroidTest : RobolectricTest() {
                     )
             }
 
+        arrayOf<CharSequence?>("foo", null, "bar").encodeDecode {
+            assertThat(size()).isEqualTo(1)
+            assertThat(getCharSequenceArray(""))
+                .isEqualTo(arrayOf<CharSequence?>("foo", null, "bar"))
+        }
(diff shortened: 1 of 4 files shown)


marcellogalhardo (approved):
This looks good to me. Thanks for your contribution!

marcellogalhardo:
We had to revert this change due to a performance regression in our benchmarks.

For details, please see: https://issuetracker.google.com/400696665##comment4

PR #897, opened 2026-07-23
Flag Isolated Projects violations on build service registrations

joshfriend:
## Proposed Changes

Add new lint check for isolated projects violations on `BuildServiceRegistry`

Since Gradle 9.7 only `findByName` is permitted on `BuildServiceRegistry.getRegistrations()` when Isolated Projects is enabled (see gradle/gradle#37622). Report all other method calls under `GradleProjectIsolation` and suggest `findByName` for name lookups.

## Testing

Test: Integration tests are included

### DIFF
--- a/lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
+++ b/lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
@@ -27,6 +27,8 @@ import com.android.tools.lint.detector.api.Scope
 import com.android.tools.lint.detector.api.Severity
 import com.intellij.psi.PsiClass
 import com.intellij.psi.PsiClassType
+import com.intellij.psi.PsiWildcardType
+import com.intellij.psi.util.PsiUtil
 import org.jetbrains.kotlin.psi.KtSimpleNameStringTemplateEntry
 import org.jetbrains.kotlin.psi.KtStringTemplateEntry
 import org.jetbrains.uast.UBinaryExpression
@@ -48,6 +50,7 @@ class DiscouragedGradleMethodDetector : Detector(), Detector.UastScanner {
         object : UElementHandler() {
             override fun visitCallExpression(node: UCallExpression) {
                 checkForConfigurationToConfigurableFileCollection(node)
+                checkForBuildServiceRegistrationsAccess(node)
                 val methodName = node.methodName
                 val potentialReplacements = REPLACEMENTS[methodName] ?: return
                 val method = node.resolve() ?: return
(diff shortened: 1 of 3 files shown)


lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
juliamcclellan, line 149: If this used true for `eraseTypeParameter`, would that erase `PsiWildcardType`s into their bounds such that the case below wouldn't be needed? I haven't used `substituteTypeParameter` before so just wanted to check.

juliamcclellan, line 145: Can you add a comment with what `substituteTypeParameter` does?

lint/lint-gradle/src/test/java/androidx/lint/gradle/ProjectIsolationIssueTest.kt
juliamcclellan: Should this one have an autofix?

lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
juliamcclellan: Can you add something like "on BuildServiceRegistration collection" to this message to clarify when the method isn't allowed? (Since the methods could be called on other collections without error)

juliamcclellan, line 158: Can you add a test for this case?

juliamcclellan (changes requested):

lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
joshfriend: Explained it better in the comment above the call 👍

lint/lint-gradle/src/test/java/androidx/lint/gradle/ProjectIsolationIssueTest.kt
joshfriend: It did, I just didn't have a test for it before

lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
joshfriend: `Test allowed usages of build service registrations` has an example `.let { }` call that exercises the extension method exception

juliamcclellan:
Thanks for adding this! The lint presubmit runner is failing with this error:
```
  * What went wrong:
  Execution failed for task ':lint:lint-gradle:ktCheck' (registered by plugin class 'androidx.build.AndroidXImplPlugin').
  > Failed check for the following files:
    /home/runner/work/androidx/androidx/lint/lint-gradle/src/main/java/androidx/lint/gradle/DiscouragedGradleMethodDetector.kt
    
    
    ********************************************************************************
    You can automatically fix these issues with:
    ./gradlew :lint:lint-gradle:ktFormat
    ********************************************************************************
  ```

joshfriend:
remaining CI failures seem to be existing compilation issues from the main branch

juliamcclellan (approved):

PR #898, opened 2026-07-25
fix:AndroidComposeView reads previous MotionEvent after recycling it

TestPlanB:
## Proposed Changes
- Avoid reading `previousMotionEvent` after it has been recycled.
- Use the cached `lastEvent` for hover-exit / down-event detection.
- Add a regression test for the mouse down path after hover exit.

## Testing

AndroidPointerInputTest#mousePress_afterHoverExit_doesNotReadPreviousMotionEventAfterRecycle

It verifies that when AndroidComposeView has a cached ACTION_HOVER_EXIT and then receives a mouse ACTION_DOWN , Compose does not read the old MotionEvent after recycling it.

## Issues Fixed

Fixes: https://issuetracker.google.com/issues/538645026


TestPlanB:
Could you please help review this patch? Thanks a lot.  @codingjeremy

PR #899, opened 2026-07-28
Using sdkmanager NDK installed in playground

danysantiago:
Also moved the ndkVersion used in androidx to a Gradle property so it is accessible from buildSrc.

Bug: 537276977
Test: Github runner


danysantiago:
Closing, only created to validate github runner.

PR #903, opened 2026-08-23
[Core] Delegate UriCompat.toSafeString to platform on API 28+

zanarellidev:
## Proposed Changes

- Delegate `UriCompat.toSafeString(@NonNull Uri uri)` to platform `android.net.Uri.toSafeString()` via `Api28Impl` on API 28 and above.
- Preserves the manual redaction algorithm as a backwards-compatible fallback on earlier API levels (SDK < 28).

## Testing

Test: Existing `UriCompatTest` suite in `:core:core`.


zanarellidev:
@google-cla I signed it!

